PR #2, opened 2026-01-26
feat: add Claude Code plugin and marketplace for Wix CLI skills

mirivoWix:


.claude-plugin/marketplace.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "wix-plugins",
+  "owner": {
+    "name": "Wix"
+  },
+  "metadata": {
+    "description": "Agent skills and MCP servers for building Wix CLI applications"
```
gilisho: Can you update the description please? It should not be specific to Wix CLI.

PR #28, opened 2026-02-20
Revise description in marketplace.json

tuvit:
Updated the description to include tools for Wix development.


gilisho (dismissed):

gilisho (approved):
According to the [marketplace entity schema](https://code.claude.com/docs/en/plugin-marketplaces#optional-metadata), the entry should still be `metadata.description`.

PR #62, opened 2026-03-19
feat(wds-docs): add cheat sheets to cut tool calls by ~75%

mirivoWix:
## Summary

- **Add 3 pre-built reference files** (`COMMON_COMPONENTS.md`, `ICONS.md`, `RECIPES.md`) that cover ~90% of WDS lookups in 1-2 Reads instead of 4+ tool calls per component
- **Restructure SKILL.md** to use cheat-sheet-first flow: read references first, fall back to staged docs only for uncommon components
- **Embed deprecation warnings and type gotchas** directly in cheat sheets to prevent known issues (e.g., `onCloseButtonClick` deprecated, `min` must be number not string, `icon` required on `TableActionCell` secondary actions)

## Problem

Every component lookup required 4+ tool calls at runtime (grep components.md → read Props.md → grep Examples.md headings → read specific example). For a dashboard page using 6 components, that's ~24 tool calls just for WDS docs. In a real codegen run (ShipRateManager app), a WDS docs failure cascade wasted ~62s (18% of total time) and 15+ steps searching for icon names in node_modules.

## What Changed

| File | Purpose |
|------|---------|
| `references/COMMON_COMPONENTS.md` (670 lines) | Top ~20 components with correct prop types, JSX snippets, deprecation warnings |
| `references/ICONS.md` (147 lines) | ~100 most-used icons by category with naming conventions (standard vs Small vs Filled) |
| `references/RECIPES.md` (363 lines) | 6 ready-to-adapt dashboard patterns (CRUD table, form modal, delete confirm, settings card, bulk selection, empty state) |
| `SKILL.md` (146 lines, was 216) | Restructured: FAST PATH (read cheat sheets) → SLOW PATH (staged docs lookup for uncommon components) |

## Expected Impact

| Metric | Before | After |
|--------|--------|-------|
| Tool calls for 6-component page | ~24 | ~2 |
| Icon lookup steps | ~15 (searching node_modules) | 1 Read |
| Deprecated prop errors | Frequent | Prevented inline |
| Pattern assembly | From scratch each time | Copy-adapt from recipes |

## Test plan

- [ ] Copy updated skill to `~/.agents/skills/wds-docs/` and run opencode in a generated-app to verify skill loads and references are found
- [ ] Run `yarn preview` in codegen and generate a dashboard page — verify fewer WDS-related tool calls in codegen.log
- [ ] Verify uncommon components (not in cheat sheet) still work via the fallback staged lookup

### DIFF
--- a/skills/wds-docs/references/COMMON_COMPONENTS.md
+++ b/skills/wds-docs/references/COMMON_COMPONENTS.md
@@ -1,0 +1,669 @@
+# WDS Common Components Cheat Sheet
+
+Quick-reference for the ~20 most-used components. Use this **first** before falling back to the full docs.
+
+All imports: `import { ComponentName } from '@wix/design-system';`
+All icons: `import { IconName } from '@wix/wix-ui-icons-common';`
+
+---
+
+## Page
+
+Compound: `<Page.Header>`, `<Page.Content>`, `<Page.Tail>`, `<Page.Section>`, `<Page.Footer>`
+
+| Prop | Type | Notes |
+|------|------|-------|
+| `height` | string | Use `"100vh"` or `"calc(100vh - 48px)"` for BM pages |
+| `maxWidth` | number | Default 1248. Excludes padding. |
+| `minWidth` | number | Default 864. |
+| `sidePadding` | number | Default 48px. |
+
+`<Page.Header>` key props: `title` (string), `subtitle` (string), `actionsBar` (ReactNode — put buttons here), `breadcrumbs` (ReactNode).
+
+```tsx
+<Page height="calc(100vh - 48px)">
+  <Page.Header title="Products" actionsBar={<Button prefixIcon={<Add />}>Add Product</Button>} />
+  <Page.Content>
+    {/* content here */}
+  </Page.Content>
+</Page>
+```
+
+---
+
+## Button
+
+| Prop | Type | Values |
+|------|------|--------|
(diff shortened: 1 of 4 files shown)


skills/wds-docs/references/COMMON_COMPONENTS.md
gilisho, line 1: This file is more than 600 lines, hence probably too long for the agent to read at once and we should make sure this doesn't affect cost+duration.
In some cases, the agent will read stuff from here the it doesn't need (e.g. read about modal components when it just needs table).
I suggest to split this file according to the components and different intents.

skills/wds-docs/references/RECIPES.md
gilisho, line 1: 1. We have recipes for tables (mostly), for cards... these are mixed intents. Not sure we want the agent to read 360 lines just like that. Worth considering splitting.
2. Maybe rename this file/folder to dashboard-recipes because it's only for dashboard AFAICT.

gilisho (commented):
Tool call reduction is very nice!
Please pay attention to my comments. My assumption is that the long files may slow down the agent a bit... But this is an assumption that should be put to the test (or more correctly, evals 😄 ).
I think we should perform evals to make sure there's an improvement in cost+time - run the 3 basic prompts (and preferably more if you have ideas) in Claude Code with the baseline skills vs the skills on this branch and see the differences.

mirivoWix:
BAD PR, DONT REVIEW IT 😢

PR #66, opened 2026-03-24
fix(wix-cli-service-plugin): replace static type definitions with MCP…

yanivefraim:
… doc lookups

Static request/response type definitions in SPI reference docs were prone to becoming outdated. Replace them with links to ReadFullDocsMethodSchema and ReadFullDocsArticle MCP tools so implementations always use the latest types.


mirivoWix (approved):
🚀 
We will have to improve it more, but it's good for now. It will fix all our bugs with the types

PR #71, opened 2026-03-30
docs(wix-cli): use CODE_IDENTIFIER for site component type prefix and extract app identifiers to reference

mirivoWix:
## Summary
- Replace hardcoded `platform.` type prefix with `{CODE_IDENTIFIER}.{PascalCaseFolderName}` in site component extensions, aligning with how `prompt-utils.ts` already uses `CODE_IDENTIFIER`
- Extract app namespace and code identifier instructions into a shared `references/APP_IDENTIFIERS.md` file to reduce token usage in the orchestrator skill
- Add CODE_IDENTIFIER provided/not-provided handling to site-component skill (matching data-collection's namespace pattern)
- Reorder app namespace instructions to check existing first, then create if missing

## Test plan
- Deploy preview with App Builder
- Tested locally with skills on new cli app
image

### DIFF
--- a/skills/wix-cli-orchestrator/SKILL.md
+++ b/skills/wix-cli-orchestrator/SKILL.md
@@ -158,25 +158,10 @@ Follow the checklist; steps below add detail.
 Only ask for configuration values when **absolutely necessary** for the implementation to proceed — i.e., the sub-agent literally cannot generate working code without it. If a value can be configured later or added as a manual step, don't block on it.
 
 **App Namespace Requirement:**
-When creating a Data Collection, you MUST ask the user for their app namespace from Wix Dev Center. This is a required parameter that must be obtained from the user's Dev Center dashboard and cannot be recommended or guessed.
+When creating a Data Collection, you MUST ask the user for their app namespace. If not provided, read [APP_IDENTIFIERS.md](references/APP_IDENTIFIERS.md) and give the user the instructions to obtain it.
 
-**Instructions to give the user:**
-
-**If you don't have an app namespace yet:**
-1. Go to [Wix Dev Center](https://manage.wix.com/studio/custom-apps/)
-2. Select your app
-3. In the left menu, select **Develop > Extensions**
-4. Click **+ Create Extension** and find the **Data Collections** extension
-5. Click **+ Create**
-6. You will be prompted to create an app namespace - follow the prompts to set it up
-
-**If you already have an app namespace:**
-1. Go to [Wix Dev Center](https://manage.wix.com/studio/custom-apps/)
-2. Open your app dashboard
-3. Click the three dots (...) menu button in the top-right corner (next to "Test App" button)
-4. Select "View ID & keys" from the dropdown menu
-5. In the modal that opens, scroll to the bottom to find the "Namespace" field
-6. Copy the Namespace value
+**Code Identifier Requirement:**
+When creating a Site Component, you need the user's Code Identifier. If not provided, read [APP_IDENTIFIERS.md](references/APP_IDENTIFIERS.md) and give the user the instructions to obtain it.
 
 If unclear on approach (placement, visibility, configuration, integration), ask clarifying questions. If the answer could change the extension type, wait for the response before proceeding. Otherwise, proceed with the best-fit extension type.
 
--- a/skills/wix-cli-site-component/SKILL.md
+++ b/skills/wix-cli-site-component/SKILL.md
@@ -418,7 +418,7 @@ export const sitecomponentMyComponent = extensions.siteComponent({
   ...manifest,
   id: "{{GENERATE_UUID}}",
   description: "My Component",
-  type: "platform.MyComponent",
+  type: "<CODE_IDENTIFIER>.MyComponent",
   resources: {
     client: {
       component: "./extensions/site/components/my-component/component.tsx",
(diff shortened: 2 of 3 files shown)


skills/wix-cli-orchestrator/SKILL.md
```diff
@@ -162,21 +162,33 @@ When creating a Data Collection, you MUST ask the user for their app namespace f
...
-**If you don't have an app namespace yet:**
+**First, check if you already have an app namespace:**
+1. Go to [Wix Dev Center](https://manage.wix.com/studio/custom-apps/)
+2. Open your app dashboard
+3. Click the three dots (...) menu button in the top-right corner (next to "Test App" button)
+4. Select "View ID & keys" from the dropdown menu
+5. In the modal that opens, scroll to the bottom to find the "Namespace" field
+6. If you see a Namespace value, copy it
+
+**If there is no app namespace, create one:**
```
mirivoWix: Every new cli app has app namespace by default  

Image

skills/wix-cli-orchestrator/SKILL.md
```diff
@@ -162,21 +162,33 @@ When creating a Data Collection, you MUST ask the user for their app namespace f
...
 4. Click **+ Create Extension** and find the **Data Collections** extension
 5. Click **+ Create**
 6. You will be prompted to create an app namespace - follow the prompts to set it up
 
-**If you already have an app namespace:**
+**Code Identifier Requirement:**
+When creating a Site Component, you need the user's Code Identifier from Wix Dev Center. This is a required value used as the type prefix for site component extensions (e.g. `type: '<CODE_IDENTIFIER>.ComponentName'`). Every Wix app has one automatically — there is no need to create it.
+
+If the user already provided their Code Identifier, pass it to the implementation sub-agent. If not, ask the user for it before proceeding with these instructions:
+
```
mirivoWix: Every new cli app has Code Identifier by default  

Image

skills/wix-cli-site-component/SKILL.md
```diff
@@ -430,13 +430,23 @@ export const sitecomponentMyComponent = extensions.siteComponent({
...
+The Code Identifier is a required value that cannot be guessed. Every Wix app has one automatically.
+
+**If Code Identifier is provided in the prompt:**
+- Use it as the type prefix: `<actual-code-identifier>.ComponentName`
+
+**If Code Identifier is NOT provided:**
+- Use the placeholder `<CODE_IDENTIFIER>` in the type field: `<CODE_IDENTIFIER>.ComponentName`
+- Add to Manual Action Items: "Replace `<CODE_IDENTIFIER>` with your actual Code Identifier from Wix Dev Center"
+
 **CRITICAL: UUID Generation**
```
mirivoWix: Copied from wix-cli-data-collection skill

Image

skills/wix-cli-orchestrator/SKILL.md
```diff
@@ -162,21 +162,33 @@ When creating a Data Collection, you MUST ask the user for their app namespace f
...
+**If there is no app namespace, create one:**
 1. Go to [Wix Dev Center](https://manage.wix.com/studio/custom-apps/)
 2. Select your app
 3. In the left menu, select **Develop > Extensions**
 4. Click **+ Create Extension** and find the **Data Collections** extension
 5. Click **+ Create**
 6. You will be prompted to create an app namespace - follow the prompts to set it up
 
-**If you already have an app namespace:**
+**Code Identifier Requirement:**
```
gilisho, line 163: I think we can remove this from the orchestrator since these instructions already present in the site component skill

skills/wix-cli-orchestrator/SKILL.md
gilisho: And existing apps won’t have a namespace OOTB?

gilisho (commented):
Suggestions:
1. Move the app namespace and code identifier to reference files inside orchestrator skill instead of having them inline in main skill file. This way it won’t be loaded when agents don’t need them.
2. The site component and orchestrator have both examples on code identifier. They are a bit different from each other too which could lead to confusion (if there no code identifier sent: in site component- leave placeholder as is; in orchestrator - ask the user). I think we should have a unified instruction. Perhaps leave placeholder and leave it as manual step at the end.

gilisho:
https://github.com/wix/skills/pull/71#issuecomment-4155898930
sorry missed this comment since I saw it on Slack. Next time mark the PR as a draft (:

skills/wix-cli-orchestrator/SKILL.md
mirivoWix: Yes, but we still have the step what to do if there is no app namespace

mirivoWix: I just changed the order

skills/wix-cli-orchestrator/SKILL.md
mirivoWix: But we span subagent here for each skill, this should be done before in the parent agent

mirivoWix:
@gilisho i think you missed what i did
1. In the orchestrator skill (based on your design), it determines when to invoke sub-agents.
2. For the app namespace, I only changed the order of operations: it now first checks whether the app namespace exists. If it doesn’t, it is created using the new extension flow.
3. Regarding the code identifier, the logic remains consistent with the app namespace. The orchestrator is responsible for retrieving the code identifier, and the site component skill should obtain it from the parent agent, just like in the data collection flow.
4. In the app builder scenario, we always pass the code identifier, so this approach will work as expected.

skills/wix-cli-orchestrator/SKILL.md
gilisho: I was just asking to understand the current situation. 👍

gilisho:
Yeha could be. Diff is pretty confusing 😄 

But anyway, we still add here 10 lines to the orchestrator skill that in most cases won’t be relevant to the agent. I believe it’s probably a good time to create a reference file for namespace+code identifier stuff to reduce token usage.  Wdyt?
It was probably the optimal choice from the beginning.

gilisho (approved):

PR #81, opened 2026-04-13
Add wix-app-management skill with Stores recipes

yotamsu:
Add already exist skill to the public repo.
The skills md were already exist so they just copied.
I chose to locate the yaml under yaml folder to not effect the context of the skill downloading. 
And we need to check if the index skill are seems to be good enough: those the skill that under wix-app-management and under each folder. (not the the md that under reference folder)

3 discussions :
- are we want to add yaml folder and there locate all the yaml ? like the pr did
- under every business solution there another SKILL.md along the SKILL.md that under the wix-app-management. does we need to name it differently ?
- Are the references structure is good enough ? or we need just the main SKILL.md and locate all the other folder under reference folder ? then we will have reference folder and under it another reference folder


mirivoWix:
@yotamsu 
> under every business solution there another SKILL.md along the SKILL.md that under the wix-app-management. does we need to name it differently ?

Yes, you can't call it SKILL.md, name it differently

> Are the references structure is good enough ? or we need just the main SKILL.md and locate all the other folder under reference folder ? then we will have reference folder and under it another reference folder

The folders should be under reference, like this:
wix-app-management/
├── SKILL.md (overview and navigation)
└── references/
.... 


Check this out https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices :)

.claude-plugin/plugin.json
```diff
@@ -12,6 +12,22 @@
...
+    "./skills/wix-app-management/events",
+    "./skills/wix-app-management/forms",
+    "./skills/wix-app-management/media",
+    "./skills/wix-app-management/payments",
+    "./skills/wix-app-management/platform",
+    "./skills/wix-app-management/pricing-plans",
+    "./skills/wix-app-management/restaurants",
+    "./skills/wix-app-management/rich-content",
+    "./skills/wix-app-management/sites",
+    "./skills/wix-app-management/stores",
```
mirivoWix:
```suggestion
    "./skills/wix-app-management",
```

skills/wix-manage/SKILL.md
```diff
@@ -0,0 +1,216 @@
+---
+name: wix-app-management
+description: "Wix business solution management recipes — REST API operations for configuring and managing Wix business solutions. Routes to: stores, bookings, get-paid, CMS, contacts, events, forms, media, app-installation, pricing-plans, restaurants, rich-content, sites, blog, calendar, domains, site-properties, ecommerce."
+compatibility: Requires Wix REST API access (API key or OAuth).
+---
```
mirivoWix, line 5: Let's consult about the name https://wix.slack.com/archives/C097SLAR37A/p1776253558113159

mirivoWix (dismissed):
Great job! let's wait for the name of skill approval 🙏

PR #82, opened 2026-04-14
Auto Skill Eval 1 #test-eval-action

AmitFeldman:
Minimal action that verifies EvalForge secrets/variables are correctly configured by calling GET /projects/:projectId with app credentials and posting the result as a PR comment.

Only runs when PR title contains #test-eval-action.


OrLeib (approved):
LGTM, let's verify

PR #109, opened 2026-04-27
fix(wix-manage): improve domain-search-and-purchase recipe

danielbrodi:
Improves the existing domain-search-and-purchase recipe so the MCP agent reliably guides users through domain search and generates a working purchase link.

Jira epic: DOM2-7294

## What changed

**Better purchase URL**
Switched from `/dashboard/{siteId}/premium-express-checkout-app/storefront-bundle-selection` to `/premium-domains/split-page?domainName={domain}`. The old URL required a site ID (calling ListWixSites, asking the user to pick a site) which added complexity and caused the agent to fail before reaching the link generation step. The new URL needs only the domain name and opens a page where the user can choose between getting the domain free with a site plan or buying domain only.

**Enriched domain search and suggestions**
The availability check and suggestions steps now include response JSON examples, free-text query support (e.g. "pancakes business"), TLD filtering, presentation tips, and handling for all edge cases (taken, unsupported TLD, premium domains). Previously the agent had the API endpoints but lacked guidance on how to handle results.

**Explicit agent behavior rules**
Added clear rules: never tell users to "go to the Wix dashboard and search", never say you can't generate a purchase link, never ask which site to connect to (the checkout page handles that).

**New example flows**
Added brainstorming flow (user has no specific domain, just a business idea) and fleshed out all other flows to show the complete path from user request to purchase link.

## Why

Manual testing on ChatGPT with the production Wix MCP showed the agent fails to generate the checkout link with the current recipe, even though the recipe is loaded and discoverable. The agent gets stuck on site ID selection or gives up and tells users to go search manually in the Wix dashboard.


skills/wix-manage/references/domains/domain-search-and-purchase.md
```diff
@@ -1,104 +1,98 @@
...
-**Example Request**:
-```bash
-curl -X GET \
-  'https://www.wixapis.com/domain-search/v2/check-domain-availability?domain=mybusiness.com' \
-  -H 'Authorization: <AUTH>'
-```
+The `domain` parameter **must** include the TLD (e.g., `mybusiness.com`, not just `mybusiness`). If the user gives a name without a TLD, default to `.com` first.
 
-**Example Response**:
+**Response when available**:
```
hadarzaguri, line 41: when available?

skills/wix-manage/references/domains/domain-search-and-purchase.md
```diff
@@ -1,104 +1,98 @@
...
     "available": true,
-    "premium": false,
-    "premiumType": "UNKNOWN_PREMIUM_TYPE"
+    "premium": false
   }
 }
 ```
 
-### Response Fields
+**Response when taken**:
```
hadarzaguri, line 52: when taken?

skills/wix-manage/references/domains/domain-search-and-purchase.md
danielbrodi: These are just two example API responses showing what the CheckDomainAvailability endpoint returns in each case - one when the domain is available (available: true) and one when it's taken (available: false). Helps the agent know what to expect and how to handle each result.

skills/wix-manage/references/domains/domain-search-and-purchase.md
danielbrodi: Same as above - just an example response for the taken case so the agent knows both possible outcomes and how to react (when taken, immediately suggest alternatives instead of just saying its taken and stopping).

gilisho (approved):

PR #110, opened 2026-04-27
fix(wix-app): default-export event handlers + correct CRM import path

orgoldreich1:
## Summary

Two related fixes for the `wix-app` skill's backend-event references:

1. **`export default` on every event handler example.** Wix CLI builds call `module.default()` on each event source file to register the webhook (see `@wix/astro`'s `getWebhookSlug` / `isValidEventFileExport`). Handler files without `export default` fail with:
   ```
   Invalid export for event source at <path>.
   Expected event listener call to be the default export.
   ```
   The previous examples showed the listener call as a bare statement, so agents reproducing the pattern produced unbuildable code.

2. **Fix the broken `@wix/crm/events` import path** in `BACKEND_EVENT.md`. That subpath isn't declared in `@wix/crm`'s `package.json` exports, so:
   - `tsc` fails with `TS2307: Cannot find module '@wix/crm/events'`
   - Node fails with `Package subpath './events' is not defined by "exports" in @wix/crm`

   Switched to the namespaced form that `COMMON-EVENTS.md` already uses:
   ```ts
   import { contacts } from "@wix/crm";
   export default contacts.onContactCreated(...)
   ```

## How this was caught

Investigating a recent codegen run that failed `wix build` with `Invalid export for event source` (job `d46835dc-91b0-44b9-8c41-b8fbb81fe390`). The agent's batch-write content matched the documented pattern verbatim — both the missing `export default` and the broken `@wix/crm/events` path.

## Verification

Compiled all 6 examples against the actual installed SDK type packages (`@wix/crm`, `@wix/ecom`, `@wix/bookings`, `@wix/blog`, `@wix/data`, `@wix/essentials`, `@wix/astro`), then ran the exact runtime check the Wix CLI build performs (`isValidEventFileExport`) on each compiled module:

| Example | tsc | `module.default` is function |
|---|---|---|
| BACKEND_EVENT canonical (`contacts.onContactCreated`) | ✅ | ✓ |
| BACKEND_EVENT elevated permissions | ✅ | ✓ |
| COMMON-EVENTS CRM | ✅ | ✓ |
| COMMON-EVENTS eCommerce | ⚠️ pre-existing strict-mode warning on `event.data.order` | ✓ |
| COMMON-EVENTS Bookings | ⚠️ pre-existing strict-mode warning on `event.data.booking` | ✓ |
| COMMON-EVENTS Blog | ✅ | ✓ |

## Out of scope (separate follow-up)

Two TS strict-mode warnings on `event.data.order` / `event.data.booking` (typed as optional in their respective SDKs) — pre-existing, unrelated to the build-time error this PR fixes. Worth a separate cleanup PR with `?` guards or destructure-with-fallback.

### DIFF
--- a/skills/wix-app/references/BACKEND_EVENT.md
+++ b/skills/wix-app/references/BACKEND_EVENT.md
@@ -56,9 +56,9 @@ The `id` must be a unique, static UUID v4 string. Generate a fresh UUID for each
 Import the event from the correct SDK module and pass a handler. Wix invokes the handler with the event payload and metadata when the event occurs. Handler signatures are documented in the [JavaScript SDK reference](https://dev.wix.com/docs/sdk).
 
 ```typescript
-import { onContactCreated } from "@wix/crm/events";
+import { contacts } from "@wix/crm";
 
-onContactCreated((event) => {
+export default contacts.onContactCreated((event) => {
   console.log("Contact created:", event.entity);
   // Custom logic: sync to CRM, send welcome email, etc.
 });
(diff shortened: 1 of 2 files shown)


mirivoWix (approved):
Amazing 🚀

PR #127, opened 2026-04-29
docs(wix-app): expand Stores V1/V3 reference into full cheatsheet

mirivoWix:
## Summary

- Replaces the minimal `STORES_VERSIONING.md` with a copy-paste-ready cheatsheet so agents stop searching MCP docs on every Stores task — the goal is faster, cheaper Stores app generation.
- Updates the `SKILL.md` callout to point agents at the new reference *before* searching SDK docs and to request both V1 and V3 scopes.

## Measured impact

Benchmark prompt: **"create a dashboard page that shows all my products"**

| | Cost | Time |
|---|---|---|
| **Before** (avg) | $0.96 | 3.74 min |
| **After** run 1 | $0.34 | 1.74 min |
| **After** run 2 | $0.53 | 2.11 min |
| **After** run 3 | $0.44 | 1.95 min |
| **After** (avg) | **$0.44** | **1.93 min** |
| **Improvement** | **−54%** | **−48%** |

The "before" run wasted cycles searching SDK docs for V1 and V3, hitting runtime errors (`_createdDate is not declared as sortable`), then retrying. The "after" runs go straight to the cookbook recipe — cursor pagination on V3, offset on V1, no retries, no runtime errors.

## What changed

`skills/wix-app/references/STORES_VERSIONING.md` now contains:

- **Module map** (V1 → V3, including `@wix/categories` for V3 categories)
- **Permissions cheatsheet** — V1 and V3 scopes side-by-side using raw `SCOPE.X.Y` IDs
- **Dual-catalog recipes** for list / get / create / update / delete products, inventory increment/decrement, collections↔categories
- **V1 → V3 field mapping** condensed from the [migration guide](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-v3/catalog-v1-to-v3-migration-guide) (product fields, pricing model, inventory, media)
- **Webhook mapping** (V1 product/collection events → V3 product/category events)
- **Top V3 gotchas**: every product has ≥1 variant, `revision` required on update, array fields overwrite (don't merge), UPPER_CASE enums, prices as strings, hidden products require `SCOPE.STORES.PRODUCT_READ_ADMIN`, V3 cursor paging has no total/skip, `_createdDate` sort fails at runtime, etc.
- **Recommended code layout** — centralized version-branching helper so the rest of the app stays version-agnostic

`skills/wix-app/SKILL.md` Stores Versioning Requirement block is reworded to advertise the new reference as an 80%-coverage cheatsheet and to require both V1+V3 scopes.

## Why

Today, agents handling any Stores task have to: search SDK docs for V1, search again for V3, read both full articles to understand differences, fetch permissions for both. That's slow and token-heavy. Pre-loading the common module map, permissions, recipes, and field map into the skill removes most of those round-trips — the measured impact above is the result.

## Reviewer notes

- Permission scope IDs follow the team convention (raw `SCOPE.DC-XXX.YYY` IDs, no human-readable label in parens) and are cross-checked against `@permissionScopeId` annotations in the live SDK type files.
- Code recipes use the dual-catalog pattern from the official [Catalog Versioning sample flow](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-versioning/sample-flow).
- All TypeScript snippets compile under the strictest flag set the wix-app skill demands: `strict + noImplicitAny + strictNullChecks + exactOptionalPropertyTypes + noUncheckedIndexedAccess`. Verified against `@wix/stores@1.0.758` and `@wix/categories@1.0.189`.
- File trimmed from 533 → 329 lines (kept all the recipes, cut duplicate examples and design-philosophy prose).


skills/wix-app/references/STORES_VERSIONING.md
```diff
@@ -1,73 +1,329 @@
...
+
+### List products with pagination
+
+V3 and V1 paginate completely differently:
 
-  if (catalogVersion === 'STORES_NOT_INSTALLED') {
-    return [];
+| Aspect | V1 (`products.queryProducts()`) | V3 (`productsV3.queryProducts(query)`) |
+|--------|--------------------------------|----------------------------------------|
+| API style | Fluent builder: `.skip().limit().find()` | Direct call: `queryProducts({ cursorPaging })` |
```
yanivefraim: **V3 `queryProducts` is also a fluent builder — not a direct call.**

Verified against the actual types in `@wix/auto_sdk_stores_products-v-3`:

```typescript
// Actual signature:
declare function queryProducts(options?: QueryProductsOptions): ProductsQueryBuilder;

// QueryProductsOptions only has `fields` — no `cursorPaging`:
interface QueryProductsOptions {
    fields?: RequestedFieldsWithLiterals[] | undefined;
}

// The V3 builder uses .skipTo(cursor) instead of .skip(offset):
interface ProductsQueryBuilder {
    eq: (...) => ProductsQueryBuilder;
    limit: (limit: number) => ProductsQueryBuilder;
    skipTo: (cursor: string) => ProductsQueryBuilder;  // cursor-based
    ascending: (...) => ProductsQueryBuilder;
    find: () => Promise<ProductsQueryResult>;
    paging: 'cursor';
}

// V3 result uses .items (not .products), .cursors.next, .hasNext():
interface ProductsQueryResult extends QueryCursorResult {
    items: V3Product[];
    next: () => Promise<ProductsQueryResult>;
}
interface QueryCursorResult {
    cursors: Cursors;  // { next?: string, prev?: string }
    hasNext: () => boolean;
}
```

The table, code recipe, and `ProductsPage` interface all need updating:
- **Table**: Both V1 and V3 use a fluent builder. The difference is `.skip(offset)` (V1) vs `.skipTo(cursor)` (V3). `totalCount` is available in V1 (`QueryOffsetResult`) but not V3 (`QueryCursorResult`).
- **Code**: The V3 path should use the builder:
```typescript
const builder = productsV3.queryProducts({ fields: ['CURRENCY'] })
  .limit(limit);
if (cursor) builder.skipTo(cursor);
const res = await builder.find();
return {
  products: res.items,
  nextCursor: res.cursors.next ?? null,
  hasNext: res.hasNext(),
  totalCount: null,
};
```

(There is also a `typedQueryProducts(query, options)` that takes a raw query object and returns `{ products, pagingMetadata }` — but `queryProducts` is the standard entry point.)

yanivefraim (commented):
Verified all types, permissions, and signatures against the installed `@wix/stores` and `@wix/categories` packages. The vast majority is correct — permissions, CRUD signatures, field mapping, gotchas, and inventory recipes all check out. Found 3 type issues below.

skills/wix-app/references/STORES_VERSIONING.md
```diff
@@ -1,73 +1,329 @@
...
+```typescript
+// V1 write ops (createCollection, updateCollection, addProductsToCollection, …)
+// live on the `products` namespace, NOT `collections`. The `collections` namespace is read-only.
+import { products } from '@wix/stores';
+import { categories } from '@wix/categories';
+
+if (v === 'V3_CATALOG') {
+  const category = await categories.createCategory({          // returns Category directly
+    name: 'Sale',
+    treeReference: { appNamespace: 'STORES_NAMESPACE', treeKey: 'allProducts' },
```
yanivefraim: **`treeReference` belongs in the second `options` argument, not inside the category object.**

Verified against `@wix/auto_sdk_categories_categories` types:

```typescript
// Actual signature:
declare function createCategory(
  category: { name: string; parentCategory?: { _id: string }; ... },
  options?: { treeReference: { appNamespace: string; ... }; ... }
): Promise<Category>;
```

Correct usage:
```typescript
const category = await categories.createCategory(
  { name: 'Sale' },
  { treeReference: { appNamespace: '@wix/stores' } }
);
```

Also note: `appNamespace` should be `'@wix/stores'` (the package name), not `'STORES_NAMESPACE'`.

skills/wix-app/references/STORES_VERSIONING.md
```diff
@@ -1,73 +1,329 @@
...
-**Search queries** (results show both V1 and V3 namespaces):
-- `"getCatalogVersion"` → permissions and usage
-- `"createProduct"` → `productsV3` (V3) and `products` (V1)
-- `"queryProducts"` → `productsV3` (V3) and `products` (V1)
-- `"getInventoryItem"` → `inventoryItemsV3` (V3) and `inventory` (V1)
+| V1 event | V3 event |
+|----------|----------|
+| `products.onProductCreated` / `onProductChanged` / `onProductDeleted` | `productsV3.onProductCreated` / `onProductUpdated` / `onProductDeleted` |
+| `products.onVariantsChanged` | `productsV3.onProductUpdated` (with `variantsInfo` in `modifiedFields`) |
+| `collections.onCollectionCreated` / `onCollectionChanged` / `onCollectionDeleted` | `categories.onCategoryCreated` / `onCategoryUpdated` / `onCategoryDeleted` |
```
yanivefraim: **V1 collection webhook names are wrong — they live on `products`, not `collections`.**

Verified against `@wix/auto_sdk_stores_products` exports:

```typescript
// Actual exports:
products.onProductCollectionCreated
products.onProductCollectionChanged
products.onProductCollectionDeleted
```

The namespace is `products` (not `collections`) and the method names have a `Product` prefix. This is consistent with write ops also living on `products`.

Should be:
```
| `products.onProductCollectionCreated` / `onProductCollectionChanged` / `onProductCollectionDeleted` | `categories.onCategoryCreated` / ... |
```

mirivoWix:
Thanks for the careful review @yanivefraim! Verdict on each point:

**✅ Comment 2 — `createCategory` `treeReference`** — fully correct, already fixed in [`8c9f0d7`](https://github.com/wix/skills/commit/8c9f0d7). `treeReference` now passes via the `options` (2nd) arg, `appNamespace: '@wix/stores'`, `treeKey: null`. The earlier strict typecheck silently accepted the wrong shape because `Category` declares a *read-only* `treeReference` field — would have failed at runtime.

**✅ Comment 3 — V1 collection webhooks** — fully correct, fixed in [`26c01f8`](https://github.com/wix/skills/commit/26c01f8). Confirmed via the `@wix/stores` type defs:
- `products.onProductCollectionCreated` / `onProductCollectionChanged` / `onProductCollectionDeleted`
- The `collections` namespace is read-only (`getCollection`, `queryCollections`, …) — both reads and writes for collections live on `products`.

I also caught one of my own while verifying: V1 variant webhook is `products.onProductVariantsChanged`, not `onVariantsChanged`. Fixed in the same commit.

**🟡 Comment 1 — V3 `queryProducts`** — partially correct. The bare `customQueryProducts` declared as `(options) => ProductsQueryBuilder` is what you cited. But the *exported* `queryProducts` is wrapped:

```typescript
declare const queryProducts: MaybeContext<BuildRESTFunction<typeof customQueryProducts> & typeof customQueryProducts>;
```

`BuildRESTFunction` adds a second call signature `(query: V3ProductQuery, options?) => Promise<V3QueryProductsResponse>` — so the direct-call form (`productsV3.queryProducts({ cursorPaging: { limit, cursor } })`) **also compiles and works at runtime**, returning `{ products, pagingMetadata }`. Both forms are valid.

That said, your broader point stands and I took it: the builder is more idiomatic, matches V1's shape (`.skip()` ↔ `.skipTo()`, both return `res.items`), and the old comparison table was misleading because it implied the V3 builder couldn't paginate. Rewrote the recipe and the table:

```typescript
let builder = productsV3.queryProducts().limit(limit);
if (typeof cursor === 'string') builder = builder.skipTo(cursor);
const res = await builder.find();
return {
  products: res.items,
  nextCursor: res.cursors.next ?? null,
  hasNext: res.hasNext(),
  totalCount: null,
};
```

Added a one-line note that the direct-call form also exists, since some agents might encounter it in older code or examples.

All three snippets re-verified with `tsc --strict --noImplicitAny --strictNullChecks --exactOptionalPropertyTypes --noUncheckedIndexedAccess` against `@wix/stores@1.0.758` and `@wix/categories@1.0.189`.

PR #146, opened 2026-05-03
Clarify Bookings free-to-paid price updates

ayal:
## Summary
- Clarifies that changing a free Bookings service to a paid one must switch `payment.rateType` from `NO_FEE` to `FIXED` while setting `payment.fixed.price`.
- Adds the validation error recovery path for `Payment of type FREE cannot be used with payment.rate`.

## Test plan
- Markdown-only recipe update; no local test runner found in the repo root.


ayal:
#skipreview

PR #164, opened 2026-05-04
docs(wix-app): lazy-load WDS skill and align backend type names with blueprint IDs

mirivoWix:
## Why

Two small skill cleanups motivated by analyzing real codegen runs (3 jobs, ~$2.75 / 12 min total). Each fix targets a specific kind of cost or confusion observed in the logs.

### 1. Lazy-load the `wix-design-system` skill

**Today:** the `wix-app` skill says "Invoke `wix-design-system` skill FIRST when using `@wix/design-system`". In practice, agents read this as "always invoke first, just in case", and load the WDS skill content (~5KB) into context at the start of every run.

**Problem:** for backend-only or data-only jobs (Service Plugin + Backend Event + Data Collection, no UI), WDS is never used. The skill content sits in cache and inflates every subsequent step's cache-read tokens — at $0.30/MTok of cache read, this adds up over a 30-step run.

**Fix:** invoke WDS only **before writing the first `.tsx`/`.jsx` that imports `@wix/design-system`**. For UI-bearing jobs nothing changes (you still load it before the first JSX file); for backend-only jobs the load is skipped entirely.

### 2. Rename two extension types to match blueprint IDs and Wix docs

**Today:**

| Skill row name        | Blueprint type ID | Wix docs name              |
| --------------------- | ----------------- | -------------------------- |
| `Event Extension`     | `BACKEND_EVENT`   | "Event Extension"          |
| `Backend Endpoints`   | `BACKEND_API`     | "Backend APIs"             |

**Problem:** when an agent has a blueprint type like `BACKEND_EVENT` and is scanning the Extension Types Reference Table to find the matching row, the words don't line up. It has to do a fuzzy translation — which is exactly the kind of step that misfires on ambiguous user prompts. We saw a real run where a `SERVICE_PLUGIN` job got misrouted to `BACKEND_EVENT.md` based on user-prompt phrasing rather than the blueprint.

**Fix:** rename the rows so both the blueprint ID and the Wix docs vocabulary line up:
- `Event Extension` → `Backend Event Extension` (carries "Backend" from the blueprint ID and "Event Extension" from the docs)
- `Backend Endpoints` → `Backend API` (matches blueprint ID `BACKEND_API` and the "Backend APIs" docs term)

Pure renames, no structural changes. The reference files keep their existing paths (`BACKEND_EVENT.md`, `BACKEND_API.md`).

## Scope

- One file changed: `skills/wix-app/SKILL.md`
- 7 insertions, 7 deletions, all wording-only
- No reference files renamed, no parser/tooling changes, no breaking changes for downstream consumers

## Test plan

- [ ] Read the diff in `skills/wix-app/SKILL.md` and confirm all renames are consistent (Quick Decision Helper, Decision Flow, Extension Types Reference Table)
- [ ] Confirm WDS skill mentions in steps 4 and the implementation rules are aligned (lazy-load, not preemptive)
- [ ] Spot-check that no reference file or external doc inside `references/` needs updating to match


skills/wix-app/SKILL.md
```diff
@@ -330,7 +330,7 @@ Stop and report errors if any step fails. Check `.wix/debug.log` on failures.
 - **Skip discovery** when all required APIs are in reference files
 - **maxResults: 5** for all MCP SDK searches
 - **ReadFullDocsArticle** only when search results need more context
-- **Invoke wix-design-system** first when using WDS (prevents import errors)
+- **Invoke wix-design-system** only before writing the first `.tsx`/`.jsx` that imports it — never preemptively (skips ~5KB of irrelevant context for backend-only jobs)
```
gilisho: I think this line can be deleted. We already mention it twice before this line.

gilisho (dismissed):

skills/wix-app/SKILL.md
mirivoWix: Ok

PR #166, opened 2026-05-04
fix(wix-app): document widget styles function and createdBy anti-patterns

OrLeib:
## Summary

- Promote the "no functions in styles objects" warning in `CUSTOM_ELEMENT_WIDGET.md` from a buried one-liner into a top-level **Critical TypeScript Constraints** section with full failing/working examples (extraction + inline-spread variants).
- Add an explicit `createdBy` anti-pattern under `DATA_COLLECTION.md` **Anti-Patterns** — agents kept emitting `createdBy: { author: "CODE_GEN" }` on `extensions.genericExtension`, which fails `tsc --noEmit` with `TS2322: Type '{ author: string; }' is not assignable to type 'CreatedBy'`.

## Motivation

Both anti-patterns were observed repeatedly in real codegen jobs, driving multi-iteration TypeScript fix loops and wasting tokens. Surfacing them at the right reference-loading point (just-in-time, when the relevant skill is invoked) should eliminate them on the first pass instead of in fixer loops.

## Scope

Documentation only. No behavior change for skills that don't import these references.

## Files changed

- `skills/wix-app/references/CUSTOM_ELEMENT_WIDGET.md` (+44/-1)
- `skills/wix-app/references/DATA_COLLECTION.md` (+16)

## Related

Complementary to #164 (`docs(wix-app): lazy-load WDS skill and align backend type names with blueprint IDs`) — different angle on the same cost-reduction theme. No file overlap.

### DIFF
--- a/skills/wix-app/references/DATA_COLLECTION.md
+++ b/skills/wix-app/references/DATA_COLLECTION.md
@@ -362,6 +362,22 @@ All embedded script configuration must go through embedded script parameters (`e
 
 **Don't create single-value collections.** A collection with one field like `{ theme: "dark" }` should be an embedded script parameter or widget setting instead.
 
+**Don't add a top-level `createdBy` field to the data extension.** Some scaffolding examples include a `createdBy: { author: "CODE_GEN" }` block at the end of the `extensions.genericExtension({...})` call. This field is **not part of the public schema** and will fail `tsc --noEmit` with `TS2322: Type '{ author: string; }' is not assignable to type 'CreatedBy'`.
+
+❌ **Wrong:**
+
+```ts
+export const dataExtension = extensions.genericExtension({
+  compId: "...",
+  compName: "data-extension",
+  compType: "DATA_COMPONENT",
+  compData: { dataComponent: { collections: [...] } },
+  createdBy: { author: "CODE_GEN" }, // ← never include this
+});
+```
+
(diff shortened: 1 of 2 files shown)


skills/wix-app/references/DATA_COLLECTION.md
```diff
@@ -362,6 +362,22 @@ All embedded script configuration must go through embedded script parameters (`e
...
+  compId: "...",
+  compName: "data-extension",
+  compType: "DATA_COMPONENT",
+  compData: { dataComponent: { collections: [...] } },
+  createdBy: { author: "CODE_GEN" }, // ← never include this
+});
+```
+
+✅ **Right:** stop after `compData`. See [extension-template.ts](data-collection/extension-template.ts) for the complete, correct shape.
+
```
mirivoWix, line 380: It's added in Codegen automaticlly, we had an issue with the type but it was fixed

OrLeib:
Closing this as stale after re-checking against current `wix/skills@main` and `ditto@master`.

The `createdBy` part is no longer relevant: `wix-private/ditto#759` has merged, and codegen now injects `createdBy: "CODE_GEN"` for `extensions.genericExtension(...)`, which fixes the original type mismatch that motivated this guidance.

The custom-element widget styles guidance is still a reasonable idea, but `main` already has the short warning and this PR bundles it with the now-stale `createdBy` change. If we still want stronger examples for the styles-object issue, it should be reopened as a smaller PR focused only on `CUSTOM_ELEMENT_WIDGET.md`.

PR #167, opened 2026-05-04
feat(wix-app): introduce fast-validate / full-validate modes

OrLeib:
## Summary

Split `references/APP_VALIDATION.md` into two named modes and update `SKILL.md`'s checklist + Step 5 prose to match:

- **`fast-validate`** — targeted `npx tsc --noEmit <changed-paths>` only (~10–15s). Use during iterative fix loops.
- **`full-validate`** — all 4 steps (install + full `tsc --noEmit` + `npx wix build` + `npx wix preview`, ~60–90s). **Always** the final gate before reporting completion.

Decision rule (now explicit in `APP_VALIDATION.md`):

- After the **first** code generation pass → `full-validate` (catches install/build issues `tsc` alone misses).
- After **each subsequent edit** in a fix loop → `fast-validate` only.
- When `fast-validate` passes and you believe the work is complete → `full-validate` once as the final gate.

## Motivation

Real codegen jobs were running the full 4-step validation between every single edit during a fix loop, wasting ~25–50s per iteration. The workflow had no language to express "I just edited one file, only re-typecheck", so the agent defaulted to the heavy path.

## Scope

Naming + workflow guidance only. The underlying validation steps are unchanged.

## Files changed

- `skills/wix-app/SKILL.md` (+8/-11) — checklist Step 5 sub-items + Step 5 prose
- `skills/wix-app/references/APP_VALIDATION.md` (+17) — new "Validation Modes" table near the top

## Related

- Companion to #166 — both are part of the same cost-reduction effort traced from a real codegen job. No file overlap.
- Complementary to #164 (`docs(wix-app): lazy-load WDS skill...`) — different angle on the same theme.

### DIFF
--- a/skills/wix-app/SKILL.md
+++ b/skills/wix-app/SKILL.md
@@ -25,10 +25,8 @@ Helps build extensions for Wix CLI applications. Covers all extension types: das
   - [ ] Extension(s) registered in extensions.ts
   - [ ] Invoked `wix-design-system` skill FIRST when using @wix/design-system (for correct imports, especially icons)
 - [ ] **Step 5:** Ran validation (see [Validation](#validation))
-  - [ ] Dependencies installed
-  - [ ] TypeScript compiled
-  - [ ] Build succeeded
-  - [ ] Preview deployed
+  - [ ] Used **`fast-validate`** during fix loops (targeted `tsc --noEmit` only)
+  - [ ] Final **`full-validate`** passed (dependencies installed, TypeScript compiled, build succeeded, preview deployed)
 - [ ] **Step 6:** Collected and presented ALL manual action items to user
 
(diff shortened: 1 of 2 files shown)


OrLeib:
Closing this as superseded after re-checking against current `ditto@master`.

The behavior this PR tried to document is now encoded directly in codegen:

- the `validate` tool runs `tsc --noEmit` by default,
- `validate({ runBuild: true })` is reserved for the final build pass,
- `installDeps` defaults to false,
- the codegen rules explicitly say to typecheck during the fix loop and run the build once at the end.

Also, this PR's proposed `full-validate` wording includes `wix preview`, while current codegen rules say not to run preview/dev/release commands. So the PR is no longer accurate as-is.

PR #169, opened 2026-05-05
feat(wix-headless): add skill for one-prompt headless site builds

ozsay:
Single orchestrator skill that runs discovery, design, feature wiring, and preview from one prompt. Includes per-vertical instructions (stores, ecom, cms, blog, forms, gift-cards), shared contracts, scaffolding scripts, and component templates.

## Plugin loading state

This PR adds the skill files but the skill is **not exposed by the Claude Code or Cursor plugins** — both manifests use an explicit `skills` allowlist that omits `wix-headless`. Codex and Gemini do not document an allowlist mechanism, so they will auto-discover it; this is accepted for now.

| Agent | Loads `wix-headless`? | Mechanism |
|---|---|---|
| Claude Code | No | `.claude-plugin/plugin.json` `skills` array — explicit value suppresses the default `skills/` scan ([docs](https://code.claude.com/docs/en/plugins-reference)) |
| Cursor | No | `.cursor-plugin/plugin.json` `skills` array added in this PR — explicit value replaces folder discovery ([docs](https://cursor.com/docs/reference/plugins)) |
| Codex | **Yes** | `.codex-plugin/plugin.json` `skills` field is documented as string-only (`"./skills/"`); no allowlist or exclude syntax exists |
| Gemini CLI | **Yes** | `gemini-extension.json` has no `skills` field — extensions auto-discover everything in `skills/` with no documented opt-out |

### Shipping `wix-headless` later

When the skill is ready:
1. Add `"./skills/wix-headless"` to both `.claude-plugin/plugin.json` and `.cursor-plugin/plugin.json` `skills` arrays.
2. Add a row to the "Available Skills" table in `README.md`.

If full exclusion across Codex/Gemini becomes a hard requirement before then, the only universal option is relocating the directory out of `skills/` (e.g. `incubator/wix-headless/`).


.cursor-plugin/plugin.json
```diff
@@ -11,5 +11,10 @@
   "license": "MIT",
   "keywords": ["wix", "wix-cli", "mcp", "ecommerce", "cms", "backend-api"],
   "mcpServers": "./.mcp.json",
+  "skills": [
```
mirivoWix, line 14: What about codex, claude, gemini and vscode plugins?

ozsay: codex and gemini can't filter skills (from claude findings) so headless will be loaded there

adimara (approved):

PR #204, opened 2026-05-13
Add domain-dns-assistant skill using public Wix APIs

mhalevy-wix:
Translates the internal Domain Connection & DNS Assistant (skill 745615d6) into a public-API recipe skill. Replaces internal gRPC FQDNs with:
- GET  /domains/v1/connected-domains (list domains + get connectionType)
- GET  /domains/v1/dns-propagations/{domain} (check DNS status)
- PATCH /domains/v1/dns-zones/{domain} (apply DNS record fixes)

Preserves full logic: chooseDomain → checkDnsPropagation → checkManagementType → resolveDnsIssue, including NS-only skip, POINTING manual steps, auto-fix with error handling, and guardrails.


ayal (commented):
Thanks Maya, this is a useful flow, but it should not be added as a new top-level skill.

Per the contribution guidance we are adding in [`CONTRIBUTING.md`](https://github.com/wix/skills/blob/ayal/docs-contribution-guidance-main/CONTRIBUTING.md), new Wix management/API skills should fit into the existing `wix-manage` skill unless a repository admin is intentionally creating a new top-level skill.

Please update this PR to:

- Move the markdown into `skills/wix-manage/references/domains/`, for example `skills/wix-manage/references/domains/check-and-fix-dns.md`.
- Add an entry for it under the Domains section in `skills/wix-manage/SKILL.md`.
- Add it to `yaml/wix-manage/domains/documentation.yaml` with the relevant `domains` tag so the skill eval workflow can pick it up.
- Verify the API details with the Wix MCP docs tools / official docs before keeping them in the skill.

A few API details also look inconsistent with the live docs and should be fixed while moving it:

- Connected Domains `connectionType` values are documented as `POINTING`, `NAMESERVERS`, and `HIDDEN`, but the skill branches on `DNS` and `REGISTRATION`: https://dev.wix.com/docs/api-reference/account-level/domains/connected-domains/connected-domain-v1/list-connected-domains
- DNS Propagation is documented as `GET https://www.wixapis.com/premium/domains/v1/dns-propagations/{dnsPropagationId}`, while the skill uses `/domains/v1/dns-propagations/{domainName}`: https://dev.wix.com/docs/api-reference/account-level/domains/dns-propagation/get-dns-propagation
- Please verify the `failureInfo` response shape before relying on `failureInfo.invalidRecords`; if it is needed but not clear from the method schema, call that out or confirm it with the API owner.
- DNS zone updates should follow the documented `PATCH https://www.wixapis.com/domains/v1/dns-zones/{domainName}` schema for `additions` and `deletions`: https://dev.wix.com/docs/api-reference/account-level/domains/domain-dns/update-dns-zone

After the move, let the PR skill eval run against the PR MCP override, then use the results as a loop: review failures, tighten the skill, and rerun until performance is good for the target scenarios.

ayal:
One clarification on the docs/API verification point: please do not treat this as a manual docs checklist only.

The intended workflow is to connect an agent to the Wix MCP and have it actually attempt the user tasks this skill is meant to support, using the PR skill content. The agent should use the MCP docs/API tools to discover the real endpoints, examples, schemas, enum values, and response shapes while completing those tasks.

A good way to build the skill is:

1. Pick the concrete tasks you expect this skill to handle, for example diagnosing a connected domain, handling a nameserver-connected domain, handling a pointing domain, and applying/failing an auto-fix.
2. Run those tasks with an agent connected to the Wix MCP.
3. When the agent succeeds, ask it to distill the happy path, the API details it had to discover, and the missing context that would have let it succeed immediately.
4. Put only that minimal connective guidance into the skill.
5. Let the PR skill eval run through the MCP PR override, then iterate on failures until performance is good.

So the goal is not “read docs and rewrite them into a skill.” The goal is to prove the task flow with an MCP-connected agent, then capture only the small amount of guidance needed to make future agents succeed faster and more reliably.

PR #216, opened 2026-05-17
feat: add @wix/skills npm package + release workflow [CODEAI-505]

yanivefraim:
## Summary

Adds the `package.json`, GitHub Actions release workflow, and README updates needed to publish `wix/skills` content as an npm package — **without actually publishing yet**. Publishing is unblocked once `@wix/skills` is registered as a Trusted Publisher on npmjs.org.

## Why

Today the skill bodies in this repo are pulled directly from `main` by various consumers with no version pin — any change merged here lands in their environments immediately, with no rollback path. Adding `@wix/skills` as a versioned npm package lets consumers (especially `@wix/cli`) declare it as a regular dependency and pin a specific version, so CLI and skills can be kept in sync by construction.

## What this PR changes

1. **`package.json`** (new, at repo root)
   - Name: `@wix/skills`, initial version `0.0.1`.
   - `publishConfig: { access: "public", registry: "https://registry.npmjs.org/" }`.
   - `files: ["skills/**", "README.md", "LICENSE"]` — plugin manifests, evals, assets stay in the repo but are excluded from the published tarball.

2. **`.github/workflows/release.yml`** (new)
   - Manual `workflow_dispatch` trigger with inputs: `version_strategy`, `dry_run`.
   - Uses **npm Trusted Publishing via GitHub OIDC** (`permissions: id-token: write`, `npm publish --provenance --access=public`). No stored `NPM_TOKEN` secret needed.
   - Modeled on [`wix/interact`'s `release-interact.yml`](https://github.com/wix/interact/blob/master/.github/workflows/release-interact.yml), simplified for a single-package repo.
   - Supports dry-run preview.

3. **`README.md`** (updated)
   - New "npm Package (versioned distribution)" section under Installation.
   - New "Versioning" section with patch/minor/major semantics (target: AI-generated-code stability).
   - New "Releasing" section pointing at the workflow.

## Verified locally

- `npm pack` produces `wix-skills-0.0.1.tgz` — 536 KB compressed, 1.8 MB unpacked, 220 files.
- Tarball contains only `package/skills/**`, `package/README.md`, `package/LICENSE`, `package/package.json`. No leakage.
- From a consumer project, `require.resolve('@wix/skills/package.json')` resolves correctly; consumer's `node_modules/@wix/skills/skills/` contains `wix-app`, `wix-design-system`, `wix-headless`, `wix-manage`.

## Test plan

- [ ] Reviewers: confirm the `package.json` `files` list is what we want published vs kept out of the tarball (plugin manifests etc.).
- [ ] Reviewers: confirm `.github/workflows/release.yml` matches preferred conventions; flag any deviations from the `wix/interact` precedent.
- [ ] Once Trusted Publisher is registered on npmjs.org: run the workflow with `dry_run: true` to validate the full flow without publishing.
- [ ] Then run with `dry_run: false` for the first real publish.

## NOT in this PR (deliberately)

- The actual `npm publish` — blocked on Trusted Publisher registration on npmjs.org.
- `release/<major>.x` maintenance branches — not needed until the first major bump.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@wix/skills",
+  "version": "0.0.1",
+  "description": "Wix Skills for AI coding agents — versioned npm distribution of the wix/skills content.",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wix/skills.git"
+  },
+  "homepage": "https://github.com/wix/skills",
+  "bugs": {
+    "url": "https://github.com/wix/skills/issues"
+  },
+  "files": [
+    "skills/**",
+    "README.md",
+    "LICENSE"
+  ],
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://registry.npmjs.org/"
+  }
+}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@wix/skills",
-  "version": "0.0.1",
+  "version": "1.0.0",
   "description": "Wix Skills for AI coding agents — versioned npm distribution of the wix/skills content.",
   "license": "MIT",
   "repository": {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,78 @@
+name: Release - Bump Version and Publish @wix/skills
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_strategy:
+        type: choice
+        description: Version strategy
+        options:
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@wix/skills",
+  "version": "0.0.1",
```
gilisho: Let's make it major 1 so it's aligned with the Wix CLI.

gilisho (approved):

PR #217, opened 2026-05-17
fix url aware

ImriKochWix:
## Summary

Three skill scripts — `seed-utilities.sh`, `check-manifest.mjs`, `copy-utility-templates.mjs` — silently broke when invoked via `bash <(curl ...)` / `node <(curl ...)`. They locate their dependencies (`shared-utilities/`, `references/verticals/`, `templates/`) via `BASH_SOURCE` / `import.meta.url`, which resolve to `/dev/fd/N` under process substitution. The sibling lookups landed in `/dev`, scripts exited with `cannot find ...`, and the skill aborted mid-flow.

The breakage directly contradicted SKILL.md's preamble, which tells agents to stream scripts from `https://dev.wix.com/skills/wix-headless/scripts/...`. Agents that trusted the preamble hit the bug.

## What changed

Each script now auto-detects whether it can resolve its dependencies on disk (tgz install) or needs to fetch them over HTTP (process-substitution stream). Same script, both modes, no caller change.

### `seed-utilities.sh`
- Replaced `BASH_SOURCE`-derived path lookup + glob with an explicit utility list (`analytics.ts`, `ricos.ts`, `wix-image.ts`).
- On disk (sibling `shared-utilities/` exists): `cp` from `<SKILL_ROOT>/shared-utilities/`.
- Otherwise: `curl -fsSL` from `https://dev.wix.com/skills/wix-headless/shared-utilities/`. Same sibling-exists probe as the Node scripts — robust to both `/dev/fd/*` and `/tmp/sh-*` FIFO-fallback environments.

### `check-manifest.mjs`
- Added `readSkillText(relPath)` and `copySkillFile(relPath, destPath)` helpers that branch on a `SKILL_ROOT_DISK` probe (does `<scriptDir>/../references/verticals` exist?).
- All skill-local reads (vertical pack markdowns, recovery templates) go through these helpers; project-side writes are unchanged.
- Main flow is now `async` to accommodate `fetch()`.

### `copy-utility-templates.mjs`
- Same dual-mode pattern via a single `copyTemplate(sourceRel, destPath)` helper that returns `{ ok, reason }` so errors can name the pack and source path.

### `SKILL.md` — Path resolution section
- Was: "Compute `<SKILL_ROOT>` from this file's path" — only works on disk.
- Now: explicitly branches on disk-install vs URL-stream mode, defines `<SKILL_ROOT>` for both, and documents the working invocation for each. Note: `node <(curl -s ...)` does **not** work for `.mjs` — process substitution gives Node a path like `/dev/fd/N` with no extension, so Node falls back to CJS and rejects ESM syntax. The streamed Node form is `curl -s URL | node --input-type=module - <args>`. Agents in URL mode had to invent the .mjs invocation form before; now it's documented (and tested end-to-end).

## Why dual-mode and not URL-only

Keeping the on-disk fast path means tgz-installed users get zero network round-trips during script execution. The detection block is ~5 lines per script. Worth it.

## Test plan

- [ ] On-disk: `tar xzf wix-headless.tgz && cd wix-headless && bash scripts/seed-utilities.sh` from a fresh Astro project — copies the three utility files into `src/utils/`, no network calls.
- [ ] URL stream: `bash <(curl -s https://dev.wix.com/skills/wix-headless/scripts/seed-utilities.sh)` from a fresh Astro project — same result, files fetched over HTTP.
- [ ] `curl -s https://dev.wix.com/skills/wix-headless/scripts/check-manifest.mjs | node --input-type=module - /tmp/proj components stores,ecom` — parses pack manifests over HTTP, recovers any missing templates over HTTP, exits 0 / 1 appropriately.
- [ ] `curl -s https://dev.wix.com/skills/wix-headless/scripts/copy-utility-templates.mjs | node --input-type=module - /tmp/proj components` against a project with `.wix/site.json.verticals = ["stores", "ecom"]` — copies `back-in-stock.ts` and `discounts.ts` over HTTP.
- [ ] Syntax checks: `bash -n scripts/seed-utilities.sh && node --check scripts/check-manifest.mjs && node --check scripts/copy-utility-templates.mjs` all pass.

### DIFF
--- a/skills/wix-headless/SKILL.md
+++ b/skills/wix-headless/SKILL.md
@@ -28,8 +28,6 @@ Your CWD at runtime is the **project directory**, not the skill root. How you re
   - `bash <SKILL_ROOT>/scripts/X.sh <args>` → `bash <(curl -s <SKILL_ROOT>/scripts/X.sh) <args>`
   - `node "<SKILL_ROOT>/scripts/X.mjs" <args>` → `node <(curl -s <SKILL_ROOT>/scripts/X.mjs) <args>`
 
-  The three scripts that depend on sibling skill files (`seed-utilities.sh`, `check-manifest.mjs`, `copy-utility-templates.mjs`) auto-detect this mode and fetch their dependencies over HTTP — no tgz install needed.
-
 | What | Absolute path |
 |---|---|
 | Shared return contract | `<SKILL_ROOT>/references/shared/RETURN_CONTRACT.md` |
--- a/skills/wix-headless/SKILL.md
+++ b/skills/wix-headless/SKILL.md
@@ -21,7 +21,14 @@ If any prerequisite is missing, surface the specific gap to the user and stop **
 
 ## Path resolution — read this first
 
-Your CWD at runtime is the **project directory**, not the skill root. Compute `<SKILL_ROOT>` from this file's path: this file is at `<SKILL_ROOT>/SKILL.md` — strip the trailing `/SKILL.md`. Hold the absolute path in session scratch for the whole run.
+Your CWD at runtime is the **project directory**, not the skill root. How you resolve `<SKILL_ROOT>/...` references depends on how you loaded the skill:
+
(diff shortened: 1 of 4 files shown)


ImriKochWix:
#skipreview

yotamsu (approved):

PR #222, opened 2026-05-17
refactor(wix-manage): replace GetSiteData with ProfileClient, inline catalog APIs

itayhewix:
## Summary

- **Replace `GetSiteData`** with `ProfileClient.GetMyMetaSiteProfile` (`POST https://www.wix.com/wix-profile-client/v4/profile/metasite`) — verified working with curl, correct URL confirmed from ambassador `entity.json`
- **Inline profile API** (fields table, response shape, currency rule) directly into `recommend-ecommerce-strategy.md` Step 3 — LLM always sees it without a separate load
- **Inline catalog APIs** (GetCatalogAnalytics, GetProductCatalogData, GetCategoryIds) into Step 5 — now generic for all ecom domains, not discount-only; added SHIPPING goal aggregates
- **Delete `api-discount-recommendations.md`** — content fully absorbed into strategy file
- **Remove `siteId`** from all `CallWixSiteAPI` calls — confirmed not a real parameter from production logs
- **Document response shape**: nested `aSingleValue.aString|aLong`, `aLong` is a JSON string requiring parse before arithmetic
- **Add currency rule**: all money values are in `payment_currency` units, never assume USD

## Field mapping (GetSiteData → ProfileClient)

| Old | New |
|---|---|
| `country` | `merchant_business_country` |
| `industry` | `success_last_30_days_industry` |
| `businessType` | `success_last_30_days_sub_industry` |
| `visitors` | `success_last_30_days_distinct_visitors` |
| `revenue` | `online_gpv_last_30_days` |
| `ordersCount` | `last_30_days_orders_count` |
| `currency` | `payment_currency` |
| `language` | `language` |
| `aov` | derived: `parseInt(online_gpv_last_30_days) / parseInt(last_30_days_orders_count)` |
| `currentDiscounts` | not available in profile API |
| `discountMargin` | not available in profile API |


yotamsu (commented):
When you delete a reference please delete the corresponding yaml as well.
In addition you need to change the skill.md file that under wix-manage:
https://github.com/wix/skills/blob/main/skills/wix-manage/SKILL.md

and add there all the skills you have added in the previous pr, cause now someone who download wix-manage skill probably will not find the reference you have added.

yotamsu (approved):

PR #253, opened 2026-05-24
feat(manage): added evalforge gate on PRs

danshw:


.github/actions/evalforge-yaml-gate/src/utils/github.ts
```diff
@@ -0,0 +1,70 @@
...
+  mdFiles: ChangedFile[];
+  evalsAdded: ChangedFile[];
+  evalsModified: ChangedFile[];
+  evalsRemoved: ChangedFile[];
+};
+
+export function classifyChanges(files: ChangedFile[]): Classification {
+  const c: Classification = { mdFiles: [], evalsAdded: [], evalsModified: [], evalsRemoved: [] };
+  for (const f of files) {
+    if (MD_RE.test(f.filename) && f.status !== 'removed') {
```
lizasofer, line 19: **Determinism — potential false positive on nested `.md` files**

`MD_RE` only matches `.md` files at exactly one level deep under an area folder: `skills/wix-manage/references/<area>/<name>.md`. If an author adds a doc at a deeper path (e.g. `skills/wix-manage/references/blog/guides/setup.md`), it silently bypasses the gate entirely — no coverage check, no eval run, no error. This is fine if the convention is strictly enforced elsewhere, but if it is not, an author could unknowingly land uncovered doc changes by placing files in a subfolder.

Consider either:
1. Documenting this as an intentional constraint (only top-level area `.md` files are gated), or
2. Widening `MD_RE` and adding a clear error for docs at unsupported depths so authors get feedback instead of silent pass-through.

.github/actions/evalforge-yaml-gate/src/utils/coverage.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function computeCoverage(
+  changedFiles: ChangedFile[],
+  scenarios: Map<string, LoadedScenario>,
+  canonicalUrlOf: (file: string) => string | null,
+): CoverageResult {
+  const coveredBy = new Map<string, string[]>();
+  const uncovered: Uncovered[] = [];
+
+  const scenariosByArea = new Map<string, { name: string; urls: Set<string> }[]>();
```
lizasofer, line 47: **Determinism — area-scoped coverage can cause false positives**

Coverage matching is area-isolated: a scenario in `references/blog/evals/` can only cover docs under `references/blog/`. If a team creates a cross-cutting scenario (e.g. a "getting started" guide that exercises both blog and bookings APIs), it must be duplicated into each area folder to provide coverage — otherwise the gate will report "missing coverage" for the areas it does not live in.

This is a legitimate design choice (area isolation prevents unrelated scenarios from accidentally satisfying coverage), but the `formatUncovered` error message does not mention the area constraint. An author whose scenario *does* reference the right URL but lives in a different area folder will see a confusing "missing coverage" error with no hint about why.

Suggestion: include the area name in the uncovered error message, e.g. _"No scenario under `references/blog/evals/` covers this file. Scenarios in other areas do not count."_

.github/actions/evalforge-yaml-gate/src/utils/gate.ts
```diff
@@ -0,0 +1,204 @@
...
+      if (id) selected.add(id);
+    }
+  }
+  const scenarioByPath = new Map<string, LoadedScenario>();
+  for (const ls of headScenarios.values()) scenarioByPath.set(ls.path, ls);
+  for (const f of [...classifiedChanges.evalsAdded, ...classifiedChanges.evalsModified]) {
+    const ls = scenarioByPath.get(f.filename);
+    if (!ls) continue;
+    const id = nameToId.get(ls.scenario.name);
+    if (id) selected.add(id);
```
lizasofer, line 130: **Wastefulness — no content-level diffing between pushes; every `synchronize` event triggers a full eval run**

The gate runs on every push (`synchronize`) and unconditionally syncs scenarios + creates a new MCP version + triggers an eval run. The workflow-level `cancel-in-progress: true` cancels the *GitHub Action*, but if the previous EvalForge eval run has already been triggered (line 166), it keeps executing on the EvalForge side — burning eval compute with results that will never be read.

This means rapid iteration (e.g. 5 pushes while fixing a typo in a scenario) can produce 5 EvalForge eval runs, of which only the last one matters. At scale this is a significant cost multiplier.

Consider adding a content-level check before triggering:
1. **Hash the effective inputs** (scenario YAML content + covered doc content + MCP version params) and compare against the previous run. Skip if unchanged.
2. **Cancel the previous EvalForge run** when a new push arrives — add a `cancelEvalRun` API call at the start of `runGate` (or in the concurrency cancellation path) to stop the in-flight run on the EvalForge side, not just the GitHub Action.
3. **Reuse MCP versions** — `ensureMcpVersion` already handles 409 (version exists), but each SHA gets a unique label (`pr-42-abc1234`). If the doc content has not changed between pushes, the same MCP version could be reused instead of creating a new one per commit.

lizasofer (commented):
Review focused on determinism (no false positives) and wastefulness (redundant eval runs).

.github/actions/evalforge-yaml-gate/src/utils/evalforge.ts
```diff
@@ -0,0 +1,188 @@
...
+      const versions = await this.listMcpVersions(mcpId, projectId);
+      const existing = versions.find(v => v.version === versionLabel);
+      if (!existing) throw new Error(`Version ${versionLabel} not found after 409`);
+      return existing;
+    }
+  }
+
+  async listTestScenarios(projectId: string): Promise<RemoteScenario[]> {
+    // EvalForge returns `tags: undefined` for untagged scenarios — normalize so callers can assume `string[]`.
+    const raw = await this.request<RemoteScenario[]>('GET', `/projects/${enc(projectId)}/test-scenarios`);
```
lizasofer, line 155: **Efficiency — fetches all scenarios in the project on every run**

`listTestScenarios` GETs the entire scenario list for the project with no filtering. The gate only needs scenarios that are either:
- Tagged with this PR's draft tag, or
- Named in the set of scenarios changed/referenced by the PR

As the project grows (hundreds or thousands of scenarios), this becomes an increasingly expensive and slow call — both in network transfer and in the downstream iterations that build `remoteByName`, `nameToId`, check for foreign draft tags, etc.

If the EvalForge API supports a query/filter parameter (e.g. filter by tag or by name), use it here to fetch only the relevant subset. For example, querying for scenarios tagged `draft:<repo>#<pr>` plus scenarios whose names match `changedHeadScenarios.keys()` would drastically reduce the payload and make the gate faster as the eval suite scales.

.github/actions/evalforge-yaml-gate/src/utils/promote.ts
```diff
@@ -0,0 +1,68 @@
...
+import { getSimpleConfig } from './config';
+import { EvalForgeClient, DRAFT_PREFIX, draftTagFor } from './evalforge';
+import { loadEvals, type LoadedScenario } from './evals';
+import { toScenarioBody } from './sync';
+import { deletePrMcpVersions } from './pr-cleanup';
+import { BASE_WORKSPACE_SUBDIR } from './paths';
+import { workspaceRoot } from './workspace';
+
+export async function runPromote(): Promise<void> {
+  const config = getSimpleConfig();
```
lizasofer, line 12: **Race condition — promote and cleanup both trigger on `pull_request: closed`**

Both `evalforge-yaml-promote.yml` and `evalforge-yaml-cleanup.yml` trigger on `pull_request: closed`. They're distinguished by their `if` conditions (`merged == true` vs `merged != true`), which is correct — but `promote` lacks a `concurrency` group.

The cleanup workflow uses `concurrency: evalforge-yaml-gate-pr-${{ ... }}` with `cancel-in-progress: true`, which means if both workflows happen to be queued (e.g. due to event delivery timing), the cleanup's concurrency group could interfere with a re-triggered gate run but won't protect against promote. More importantly, if a PR is merged and GitHub delivers multiple `closed` events (rare but possible with webhook retries), promote could run concurrently with itself.

Consider adding a matching `concurrency` group to the promote workflow to prevent duplicate runs.

lizasofer:
**Readability — `dist/index.js` should not be committed**

This single file accounts for 49,139 of the 54,197 added lines (91% of the PR). It makes the diff essentially unreviewable in GitHub's UI and obscures the actual source changes.

Build the bundle in CI instead — e.g. add a build step before the action runs, or use a pre-publish workflow with `@vercel/ncc`. This is the standard pattern for JS GitHub Actions (even `actions/checkout` does this). The `.gitignore` already excludes `dist/*.map`, so extending it to `dist/` entirely is a one-line change.

.github/actions/evalforge-yaml-gate/src/utils/gate.ts
```diff
@@ -0,0 +1,204 @@
...
+import { pollUntilDone, EvalRunTimeoutError } from './eval-run';
+import {
+  formatEvalFailed, formatEvalPassed, formatEvalTimeout, formatForeignDraftConflicts,
+  formatLoadErrors, formatNoChanges, formatOrphanedMds, formatServiceError, formatUncovered,
+} from './comment';
+
+type Commenter = ReturnType<typeof makeCommenter>;
+
+export async function runGate(): Promise<void> {
+  const config = getEvalConfig();
```
lizasofer, line 22: **Readability — `runGate` is a 170-line sequential monolith**

`runGate` has ~10 sequential phases (load → classify → orphan check → coverage → base load → diff sync → sync execute → select scenarios → MCP version → trigger + poll → report), all inlined in one function. Each phase is only understandable in context of what came before, which makes reviewing any single concern difficult.

Suggestion: extract named phases that each return an early-exit signal or the data needed by the next phase. For example:

```ts
export async function runGate(): Promise<void> {
  const ctx = buildContext();  // config, octokit, comment, workspace, draftTag

  const validated = await validateYaml(ctx);
  if (!validated) return;

  const classified = await classifyPrChanges(ctx);
  if (!classified.hasGatedChanges) return;

  const covered = await checkCoverage(ctx, classified);
  if (!covered) return;

  const synced = await syncDraftScenarios(ctx, classified, covered);
  if (!synced) return;

  await runAndReportEvals(ctx, synced);
}
```

Each function would be self-contained and independently testable — and the top-level `runGate` would read as a plain-English checklist of what the gate does.

.github/actions/evalforge-yaml-gate/src/utils/evalforge.ts
```diff
@@ -0,0 +1,188 @@
...
+
+const MCP_URL = 'https://mcp.wix.com/mcp';
+const MCP_CONFIG_KEY = 'wix-mcp-remote';
+
+export const TERMINAL_RUN_STATUSES = ['completed', 'failed', 'cancelled'] as const;
+export type RunStatus = 'pending' | 'running' | typeof TERMINAL_RUN_STATUSES[number];
+
+export type CapabilityVersion = { id: string; capabilityId: string; version: string };
+
+import type { RemoteScenario, ScenarioBody } from './sync';
```
lizasofer: **Minor — circular-ish type re-exports make ownership unclear**

Not critical, but worth noting: `RemoteScenario` and `ScenarioBody` are defined in `sync.ts`, then re-exported here via `import type` + `export type`. Tests end up importing `RemoteScenario` from `evalforge.ts` even though `sync.ts` owns it. This creates a dependency cycle in the reader's mental model — you have to check two files to figure out where a type actually lives.

No immediate harm, but as the codebase grows this pattern makes refactoring harder (you can't tell from imports alone which module is the source of truth for a type).

lizasofer:
**General suggestion — consider splitting into phased PRs for future changes**

Not critical for this PR since we've already reviewed it, but for future iteration: this change could have landed as 2–3 smaller PRs that each move the project forward independently:

1. **Schema + parser + mapper + tests** — the YAML contract. Reviewable and mergeable on its own with no runtime effect.
2. **Gate/promote/cleanup handlers + sync logic + EvalForge client** — the runtime behavior, building on the merged schema.
3. **Workflow files** — the wiring that activates everything.

This would let each PR get focused review, land faster, and reduce the risk of a single large rollback if one phase has issues. It also makes it easier to gate the rollout (e.g. merge 1+2 but hold 3 until the team is ready to enforce the gate).

.github/actions/evalforge-yaml-gate/src/utils/promote.ts
danshw: fixed - promote has concurrency: cancel-in-progress: false. Cleanup gated by if: merged != true

.github/actions/evalforge-yaml-gate/src/utils/github.ts
danshw: we'll support only top level for now, i don't currently see any usage for nested skills

.github/actions/evalforge-yaml-gate/src/utils/gate.ts
danshw: let's keep as is and evaluate if it's needed, it's not super straight forward

.github/actions/evalforge-yaml-gate/src/utils/evalforge.ts
danshw: fixed - RemoteScenario/ScenarioBody owned by evalforge.ts; sync.ts imports from there

.github/actions/evalforge-yaml-gate/src/utils/evalforge.ts
danshw: this is pending a fix by codegen

lizasofer (dismissed):

lizasofer (approved):

PR #274, opened 2026-05-25
ml-setup

asafreuveni:


ayal (dismissed):

yaml/wix-manage/multilingual/documentation.yaml
```diff
@@ -0,0 +1,8 @@
+apiDoc:
+  title: Wix Multilingual Recipes
+
+  docs:
+    - file: ../../../skills/wix-manage/references/multilingual/multilingual-setup.md‎
+      title: Multilingual Setup and translation
+      docsEntry: https://dev.wix.com/docs/api-reference/business-management/multilingual/introduction
```
yotamsu: it need to be under a section, not an article :
https://dev.wix.com/docs/api-reference/business-management/multilingual

yotamsu (commented):
you need to add your skill to here as well : (add a new category like the others)
https://github.com/wix/skills/blob/main/skills/wix-manage/SKILL.md

to make it available for users who install there skill to the file system.
(they will download this wix-manage skill and it will be used as an index skill to access your new skill)

yaml/wix-manage/multilingual/documentation.yaml
asafreuveni: both comments fixed. can you review again?

yotamsu:
#skipreview

yotamsu (approved):

PR #281, opened 2026-05-25
fix(domains): flatten contact address fields in domain-purchase recipe

ayal:
## Summary

The `domain-search-and-purchase` recipe shows the upsert/preview contact body with address fields nested under an `address: { ... }` sub-object, but the API actually expects those fields flat on the contact. Agents that follow the recipe literally hit a 400 with `city/country/streetAddress/postalCode must not be empty` and only succeed on a retry-with-guess.

Real example from a recent Aria conversation (`808578e0-fba9-48ba-8365-8d1368221de7`): agent sent `registrantContact.address.{...}` → 400 → retried with flat fields → 200.

## Changes

- Step 2c preview response example: flatten the address fields and add an explicit note that there is no nested `address` sub-object
- Step 3 upsert request body example: flatten the address fields the same way

## Source of truth

Verified against the canonical curl in `wix-private/premium:domains/domain-registration-intent/proto/appendices/docs/examples/DomainRegistrationIntentService.UpsertDomainRegistrationIntent.curl`, which has `street_address`, `city`, `postal_code`, `country` directly on `registrant_contact` / `admin_contact` / `tech_contact`.

## Test plan

- [ ] Re-run the domain purchase flow end-to-end and confirm the upsert succeeds on the first attempt
- [ ] Sanity-check the preview response shape against an account that already has contacts on file


orrozen (approved):

ayal:
#skipreview

ayal:
#skipreview

ayal:
#forcemerge

mirivoWix (approved):

PR #309, opened 2026-05-31
feat(wix-headless): expand allowed tools for skill execution

mirivoWix:
## Summary

Adds an `allowed-tools` allowlist to the `wix-headless` skill's `SKILL.md` frontmatter so the commands the skill relies on run without triggering permission prompts.

Allowlisted tools:
- `Bash(npx @wix/cli@latest *)`, `Bash(npx @wix/cli *)`
- `Bash(npm create @wix/new@latest *)`, `Bash(npm install *)`, `Bash(npm run *)`
- `Bash(node *)`, `Bash(bash *)`, `Bash(curl *)`, `Bash(cd *)` 
- `Read`, `Write`, `Edit`, `Skill`

### DIFF
--- a/skills/wix-headless/SKILL.md
+++ b/skills/wix-headless/SKILL.md
@@ -1,6 +1,20 @@
 ---
 name: wix-headless
 description: "Build a complete Wix Managed Headless site from a single prompt, OR connect an existing project (HTML/JSX/Vite app, Claude Design output, etc.) to Wix Headless for hosting + Business Solutions. Entry point for both: (1) new-site requests — runs discovery, design, feature wiring, and preview; and (2) existing-project requests — runs `npm create @wix/new@latest init`, analyzes the project for needed Business Solutions, installs apps, **wires the Wix SDK into the existing source files so each installed app actually powers its corresponding feature**, and releases. Triggers: build me a site, create a website, make me a website, new website, online store, I want to sell X, start a business online, launch a site, ecommerce, portfolio, business website, sell online, online shop, connect this to Wix Headless, add Wix Headless to this project, host this on Wix, deploy this to Wix, implement the features of this project using Wix Headless. Use this skill instead of the WixSiteBuilder MCP tool for new-site requests."
+allowed-tools:
+  - Bash(cd *)
+  - Bash(npx @wix/cli@latest *)
+  - Bash(npx @wix/cli *)
+  - Bash(npm create @wix/new@latest *)
+  - Bash(npm install *)
+  - Bash(npm run *)
+  - Bash(node *)
+  - Bash(bash *)
+  - Bash(curl *)
+  - Read
+  - Write
+  - Edit
+  - Skill
 ---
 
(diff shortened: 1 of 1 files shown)


ozsay:
**Reviewed against a real end-to-end run** (session `91bc3084`, GadgetsForAnyone — 18m, 13 subagents, clean). I enumerated every tool used by the orchestrator **and all 13 subagents** and checked it against this allowlist. Two real gaps:

### 1. The var-assignment prefix defeats `Bash(curl *)` (highest impact)

Prefix-glob patterns match the *start* of the command string. In the run, essentially **every** `wixapis.com` curl is written as:

```bash
TOKEN=$(cat /tmp/wix_token_gadgets.txt)
curl -sS -D /tmp/_h.$$ ...
```

That command starts with `TOKEN=$(cat …)`, **not** `curl`, so `Bash(curl *)` never fires and the call still prompts. Same shape breaks `node`/`npx` too:
- `ACCOUNT_TOKEN=$(npx @wix/cli@latest token)` then `curl …`
- `SKILL_ROOT=…` / `SITE_ID=…` / `*_STDERR=$(mktemp)` multi-line blocks

Since curl is the path this PR most targets, please confirm how the matcher handles (a) leading env-var assignments and (b) newline-separated multi-statement commands. If it matches the literal string, this PR buys almost no reduction on the curl path. Fix options: split token-mint and curl into separate Bash calls in the recipes, or accept curl prompts and document it.

### 2. Read-only helpers are 75% of Bash traffic and aren't covered

Bash calls by leading command in the run (103 total):

| head | calls | covered? |
|---|---|---|
| `ls` | 34 | ❌ |
| `grep` | 27 | ❌ |
| `find` | 16 | ❌ |
| `mkdir` | 4 | ❌ |
| `cat`/`head`/`wc` | ~3 | ❌ |
| `cp` | 1 | ❌ |
| `npx @wix/cli@latest *` | ~5 | ✅ |
| `npm install` / `node` / `bash` / bare `curl` | ~5 | ✅ |

`ls`/`grep`/`find` alone are **77 of 103** Bash calls — all read-only and safe to allow. Suggest adding:

```yaml
  - Bash(ls *)
  - Bash(grep *)
  - Bash(find *)
  - Bash(cat *)
  - Bash(head *)
  - Bash(mkdir *)
  - Bash(cp *)
```

### Minor

- **`Agent` is missing.** The orchestrator fans out to 13 subagents via the `Agent` tool (13 calls); the heavy work (107 Read / 87 Bash / 27 Write / 11 Edit) all happens inside them. Worth allowlisting.
- `Bash(npm create @wix/new@latest *)`, `Bash(npm run *)`, `Bash(npx @wix/cli *)` (no `@latest`) — **never used** in this run; scaffolding goes through `bash …/scaffold.sh`. Harmless but unverified.

Otherwise the `Read`/`Write`/`Edit`/`Skill` + toolchain entries all match real usage. 👍

ozsay (approved):

PR #412, opened 2026-06-17
feat(wix-headless): full-page LLM generation — header + footer + home (CODEAI-696, experiment)

yanivefraim:
## Experiment: LLM generates the entire visible page

Fresh off **main**. Pushes generation as far as it goes: the LLM authors the **site header, footer, and home page** — not just the home body.

### What's LLM-generated vs composed
| File | Source |
|---|---|
| `Navigation.astro` (header) | **LLM** (`navPath`) |
| `Footer.astro` | **LLM** (`footerPath`) |
| `index.astro` (home) | **LLM** (`homePath`) |
| `Layout.astro` (html/head shell, ClientRouter, imports Nav+Footer) | composed (plumbing) |
| `global.css`, `astro.config.mjs` | composed (plumbing) |

### Change
- `compose.mjs` — new required inputs `navPath` / `footerPath` / `homePath`, each written **verbatim**; errors (`NO_NAV`/`NO_FOOTER`/`NO_HOME`) if missing. Manifest records `nav`/`footer`/`home` = `"llm-generated"`.
- Removed the `Navigation.astro`, `Footer.astro`, `index.astro` templates.
- `BUILD-astro.md` — three-file generator dispatch + contracts (keep `site-nav`/`site-footer` classes + `transition:persist` + nav/home markers so chrome styling, View-Transitions, and vertical patches still work).

### Deliberately **no width guardrail**
Started fresh per request — this branch does **not** carry the `max-w-*` collision guard from #411. So the raw full-page generation is exposed to that footgun; the test run will show whether/how often it trips, with header & footer now in scope too.

### Spectrum
#408 all-pinned → #410 LLM hero → #411 LLM home → **this: LLM header+footer+home**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #416, opened 2026-06-17
fix(wix-headless): numeric spacing scale — kill the max-w-<size> collision (CODEAI-696)

yanivefraim:
## The root fix for the `max-w-3xl` collapse

**Root cause:** `compose.mjs` emitted a *named* `--spacing-*` scale using t-shirt sizes (`md`, `3xl`, …). In Tailwind v4 the width family (`max-w-`/`w-`/`min-w-`) is generated from the `--spacing-*` namespace, so `max-w-3xl` resolved to `--spacing-3xl` (~96px) instead of `--container-3xl` — collapsing reading columns to one word per line. Hand-written templates dodged it by only using `max-w-6xl`/`max-w-prose`; LLM-generated layout (#411/#412) tripped it.

**Fix (chosen: Tailwind-native numeric):** the skill's spacing values already match Tailwind's default scale 1:1 (`md`=1rem=`4`, `lg`=1.5=`6`, `4xl`=6rem=`24`), so we drop the named scale and use Tailwind's built-in numeric spacing. With no named `--spacing-<size>` tokens, nothing collides with width keys, and `max-w-md/3xl/6xl` resolve to `--container-*` correctly.

### Changes
- **`scripts/compose.mjs`** — stop emitting named `--spacing-*`; remove it from `REQUIRED`, the `DESIGN.md` read, and the container≠spacing check. Containers kept.
- **`scripts/emit-design-tokens.mjs`** — stop projecting named `--spacing-*`.
- **Migration** — 112 spacing utility classes (`gap-md`→`gap-4`, `py-4xl`→`py-24`, …) + 122 `var(--spacing-*)`→literal-rem, across 27 template + instruction-doc files.
- **Spec docs** (`DESIGN_MD.md`, `DESIGN_SYSTEM.md`, `STYLING.md`) — drop the `spacing` token group (Designer no longer authors spacing) and correct STYLING's wrong claim that `max-w-3xl`→`--container-3xl`.

### Mapping
`2xs`→`1` · `xs`→`2` · `sm`→`3` · `md`→`4` · `lg`→`6` · `xl`→`8` · `2xl`→`12` · `3xl`→`16` · `4xl`→`24`

### Verification
- ✅ compose emits **0** `--spacing-*` and the 4 `--container-*` tokens.
- ✅ **0** leftover t-shirt spacing classes / `var(--spacing-*)` refs across templates + docs (only intentional explanatory mentions remain).
- ✅ both scripts parse.
- ⏳ Live compiled-CSS check (`max-w-3xl`→`--container-3xl`) + Playwright render — pending a network-capable build env (sandbox here had no network for `wix build`).

### Scope note
Branched off `main`. The feature branches (#408 fragments / #410 hybrid / #411 full-home / #412 full-page) carry extra files (e.g. `fragments/**`) that need the same mechanical migration when they rebase onto this — the `max-w-*` guardrail added to #411/#412 stays valid in the meantime.

---

## Before / After — live sites

Both are full-page LLM-generated sites (the LLM writes `max-w-3xl` freely). Same class, opposite outcome:

| | Site | `max-w-3xl` resolves to | Rendered width @1200px |
|---|---|---|---|
| **Before** (pre-fix, named `--spacing-*` scale) | https://solara-yog-e96ca135-yanivef.wix-site-host.com/faq | `var(--spacing-3xl)` | **64px** — FAQ answer text collapses to a few characters per line |
| **After** (this PR, numeric spacing) | https://meridian-c-ed1b2263-yanivef.wix-site-host.com/ | `var(--container-3xl)` | **896px** — proper reading column |

Measured live in-browser (computed `max-width` on the `.max-w-3xl` wrappers). Compiled CSS on the "after" site: `.max-w-3xl{max-width:var(--container-3xl)}`, with `0` `--spacing-*` tokens in `@theme`.

Tracked in Jira: **CODEAI-704**.


skills/wix-headless/references/astro/templates/global.css
adimara: The --spacing-* no longer exist. An LLM reading this comment would think they're still emitted.

skills/wix-headless/references/astro/templates/gift-cards/components-gift-cards.css
adimara: These spacing tokens no longer exist.

skills/wix-headless/references/astro/templates/bookings/services/[slug].astro
Comment (adimara) [inline skills/wix-headless/references/astro/templates/bookings/services/[slug].astro]: After this fix, there is no spacing scale to fall back to, so max-w-4xl would fall back to Tailwind's default or undefined — not to a 7rem spacing value. The old warning made sense pre-fix; post-fix it's misleading.

skills/wix-headless/references/astro/stores/COMPONENTS_CSS.md
adimara: --spacing-md no longer exists.

skills/wix-headless/references/shared/IMPLEMENTER.md
adimara: spacing was removed from the DESIGN.md vocabulary.

skills/wix-headless/references/astro/stores/COMPONENTS_CSS.md
adimara: spacing was removed

skills/wix-headless/references/astro/ecom/COMPONENTS_CSS.md
adimara: spacing was removed

skills/wix-headless/references/astro/designer/INSTRUCTIONS.md
```diff
@@ -103,7 +103,7 @@ All page scopes share common inputs and rules. Scope-specific details follow.
     ```
 8. **Comments in frontmatter.** Use `//` or `/* */` — never HTML `<!-- -->` comments in `.astro` frontmatter (it's TypeScript, not HTML). HTML comments in the template section are fine.
 9. **Responsive.** All pages must work at mobile (320px), tablet (768px), and desktop (1024px+). Use Tailwind responsive prefixes (`md:`, `lg:`) and the spacing scale from `@theme`.
```
adimara: There is no spacing scale in @theme anymore.

adimara (commented):
this fix only works because the custom spacing values happened to be identical to tailwind's defaults. if someone ever wants a truly custom spacing scale (different values), putting them under
   --spacing-* will break max-w-* again. the namespace would need to change (e.g. --rhythm-*) to avoid the collision. 


the CSS template files (components-bookings.css, components-ecom.css, components-gift-cards.css) now have hardcoded rem values like gap: 1.5rem 
  instead of var(--spacing-lg). That's intentional and correct given the fix — but it means these component styles are now not customizable through the spacing system at all. Before, a brand  could technically override --spacing-lg and all components would follow. Now they can't. That's probably fine since the whole point was to remove the scale, but it's a subtle loss of flexibility worth being aware of.

skills/wix-headless/references/astro/templates/global.css
yanivefraim: Fixed in d78da2f — dropped `--spacing-*` from the @theme substitution comment and reframed it: spacing is Tailwind numeric, and a `--spacing-*` scale must never be added since that namespace also powers the `max-w-*`/`w-*` width utilities.

skills/wix-headless/references/astro/templates/gift-cards/components-gift-cards.css
yanivefraim: Fixed in d78da2f — removed the stale `--spacing-{sm,md,lg,…}` entry from the reused-tokens header.

skills/wix-headless/references/astro/templates/bookings/services/[slug].astro
Comment (yanivefraim) [reply skills/wix-headless/references/astro/templates/bookings/services/[slug].astro]: Fixed in d78da2f — the spacing-fallback rationale is obsolete post-fix. Reworded: bare `max-w-*` keys now resolve to Tailwind `--container-*`; `container-reading` remains the recommended prose width.

skills/wix-headless/references/astro/stores/COMPONENTS_CSS.md
yanivefraim: Fixed in d78da2f — swapped the `--spacing-md` example for `--radius-md` and noted spacing is Tailwind-numeric. Applied the same fix to the identical line in `ecom/COMPONENTS_CSS.md`.

skills/wix-headless/references/shared/IMPLEMENTER.md
yanivefraim: Fixed in d78da2f — dropped `spacing` from the DESIGN.md vocabulary list.

skills/wix-headless/references/astro/stores/COMPONENTS_CSS.md
yanivefraim: Fixed in d78da2f — dropped `spacing` from the DESIGN.md vocabulary list.

skills/wix-headless/references/astro/ecom/COMPONENTS_CSS.md
yanivefraim: Fixed in d78da2f — dropped `spacing` from the DESIGN.md vocabulary list.

skills/wix-headless/references/astro/designer/INSTRUCTIONS.md
yanivefraim: Fixed in d78da2f — changed "the spacing scale from @theme" to Tailwind numeric spacing (`gap-4`, `py-24`, …).

yanivefraim:
Thanks @adimara — all addressed in d78da2f.

**Stale `--spacing-*` references (the 8 inline comments):** all cleaned — the `@theme` substitution comment, the gift-cards reused-tokens header, the bookings `max-w` warning rationale, the DESIGN.md vocabulary lists (stores / ecom / IMPLEMENTER), the `--spacing-md` token examples, and the designer "spacing scale from @theme" line.

**One real bug your review surfaced indirectly:** the recent `Merge main` pulled in the rewritten bookings vertical (610d0ea), whose `components-bookings.css` shipped 6 `var(--spacing-sm/md/xl/xs)` rules — those resolve to nothing post-fix (broken gaps/padding). Migrated them to literal rem.

On the two design points:

1. **Namespace fragility** (a truly custom spacing scale under `--spacing-*` would re-collide with `max-w-*`; would need e.g. `--rhythm-*`). Agreed, and captured as a forward guard in `STYLING.md`: never reintroduce a named `--spacing-*` scale (that namespace generates the width utilities); if a brand ever needs custom spacing **values**, use a different namespace (`--rhythm-*`) or arbitrary values. You're right that the current scale only works because its steps equal Tailwind's numeric defaults — so this is exactly where to warn future authors.

2. **Loss of per-token customizability** (component CSS now hardcodes rem instead of `var(--spacing-lg)`). Intentional, as you noted — removing the scale is the whole point. Brands still customize via the color/type/radius/container `@theme` tokens; spacing is deliberately pinned to Tailwind's rhythm. If we ever want it brand-overridable again, that's the `--rhythm-*` path from point 1 — not `--spacing-*`.

yanivefraim:
#skipreview

PR #436, opened 2026-06-22
feat(eval-gate): pass commitSha and skillsRepo to eval-pipeline run-c…

hadarzaguri:
…omparison

Aligns with ax-tools evalforge-pipeline PR #217 — the /run-comparison endpoint now creates mcp-with (?skillsCommitHash+skillsRepo) and mcp-without (plain prod) internally, so we forward config.headSha and config.mcpSkillsRepo from the action. Also logs both values at startup.


hadarzaguri:
#forcemerge

hadarzaguri:
#skipreview

danshw (approved):

PR #459, opened 2026-06-29
feat(wix-manage): add analytics semantic models skill

nqstr:
## What

Adds a new **Analytics** area to `wix-manage` covering the [Semantic Model API](https://dev.wix.com/docs/api-reference/business-management/analytics/semantic-models/introduction) — retrieving a Wix site's analytics through semantic models.

The skill documents the **List → Get → Query** flow and the sharp edges an agent needs to get correct numbers:

- `interval` is required on every query.
- Half-open range semantics: `[start, end)` — `start` included, `end` excluded (set `end` to the day after the last day you want).
- Time zone: pass the site's IANA `timeZone` (from Get Site Properties) in `interval.timezone` so results match the Wix dashboard; **defaults to UTC** when omitted.
- Field names must come from `Get Semantic Model`; field `dependencies` are silently omitted if their deps aren't in the query.
- 1,000-row result cap (paginate with `paging.offset`); `formattingEnabled` and `totalsIncluded` options.

## Files

- `skills/wix-manage/references/analytics/query-site-analytics.md` — the skill
- `skills/wix-manage/SKILL.md` — Analytics index section + routing entry
- `yaml/wix-manage/analytics/documentation.yaml` — doc registration
- `yaml/wix-manage-evals/analytics/query-site-analytics.yml` — eval scenario (`tool` + `llm_judge` assertions)

## Notes

- Placed inside `wix-manage` per CONTRIBUTING.md (REST API → wix-manage; new top-level skills are admin-only).
- API details verified against the Wix MCP docs and full method schemas (not memory).
- Read-only skill (Site Analytics – read scope), so no mutating-confirmation flow.
- New `analytics` eval tag — schema-valid (only `draft:`/`pending:`/`rejected:` are reserved); confirm it's registered in EvalForge if the execution stage relies on a tag registry.


hadarzaguri (dismissed):
Fix conflict

hadarzaguri (approved):

PR #461, opened 2026-06-29
Fix incorrect paging structure in Query Sites skill

ayal:
## Why

It was reported that the **Query Sites** skill (`skills/wix-manage/references/sites/query-sites.md`) had a bad paging structure. I confirmed it by calling the live API (`POST /site-list/v2/sites/query`) — the documented structure does not match what the API returns, so anyone following it would break their pagination.

## What was wrong

| Skill said | Reality (verified live) |
|---|---|
| Response has `pagingMetadata` (`.cursors.next`, `.hasNext`) | Response has **`metadata`** (`.cursors.next`, `.hasNext`) + a separate top-level **`cursorPaging`**. `pagingMetadata` does not exist. |
| Site URL is `siteUrl` | Field is **`viewUrl`** |
| Filter results client-side | API supports server-side **`filter`** and **`sort`** in the query |
| Cursor source unspecified | Cursor is at **`metadata.cursors.next`** |
| — | Max **100** sites/page; cursor encodes filter/sort and does **not** carry `limit` |

A loop following the old docs reads `pagingMetadata.hasNext` → `undefined` → stops after the first page (or throws).

## What changed

- Corrected response structure (`metadata` + `cursorPaging`) and the no-`pagingMetadata` warning
- Replaced the fictional site object with the full authoritative `Site` field table (incl. `premium` = "Whether the site has a Wix Premium (paid) plan", `editUrl`, `editorType`, `viewUrl`, …)
- Documented `filter`/`sort` and the 100-item page limit
- Added a working pagination loop using `metadata.cursors.next` / `metadata.hasNext`

## Verification

Called the live API end-to-end with the documented loop: paged through **all 232 sites across 3 pages**, terminated correctly on `hasNext: false`, and field shapes (`premium`, `viewUrl`, etc.) confirmed against real responses and the `Site` schema.

### DIFF
--- a/skills/wix-manage/references/sites/query-sites.md
+++ b/skills/wix-manage/references/sites/query-sites.md
@@ -1,36 +1,42 @@
 ---
 name: "Query Sites"
-description: Lists and queries all sites associated with a Wix account using Sites API. Covers pagination with cursor-based navigation.
+description: Lists and queries all sites associated with a Wix account using the Sites API. Covers filtering, sorting, and cursor-based pagination.
 ---
 # Query Sites
 
-This recipe demonstrates how to list and query all sites associated with a Wix account.
+This recipe demonstrates how to list and query the sites associated with a Wix account.
 
 ## Prerequisites
 
-- Account-level API access
-- Authorization for site listing permissions
+- Account-level API access (authenticated as a Wix user or using an account-level API key)
+- Permission `SITE_LIST.READ` (scope `SCOPE.ACC-DC-OS.READ-SITE`)
 
 ## Required APIs
 
 - **Query Sites API**: [REST](https://dev.wix.com/docs/api-reference/account-level/sites/sites/query-sites)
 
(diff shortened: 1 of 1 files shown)


ayal:
#skipreview

skills/wix-manage/references/sites/query-sites.md
```diff
@@ -1,37 +1,43 @@
...
 
-This recipe demonstrates how to list and query all sites associated with a Wix account.
+This recipe demonstrates how to list and query the sites associated with a Wix account.
 
 ## Prerequisites
 
-- Account-level API access
-- Authorization for site listing permissions
+- Account-level API access (authenticated as a Wix user or using an account-level API key)
+- Permission `SITE_LIST.READ` (scope `SCOPE.ACC-DC-OS.READ-SITE`)
```
lizasofer, line 12: I'm afraid this this too specific for skill
These types of information, including schema description which was also added here should be taken from docs.

lizasofer (approved):

PR #467, opened 2026-06-30
Fix ecommerce recommendation tracking recipe fetch

guysopher:
## Summary
- Stop `recommend-ecommerce-strategy` from instructing Aria to call `ReadFullDocsArticle` on the internal-only recommendation tracking URL.
- Make the inline BatchCreate request body the source of truth for persisting proposed recommendations.
- Remove the public `docsEntry` marker from the internal tracking reference and update evals to validate the orchestrator-based persistence path instead of the broken tracking article fetch.

## Feedback / Broken Conversation
- Conversation: https://wix-bo.com/ai-assistant/conversations/8340ed13-5f1c-4018-b6ee-771e6e2990d1
- Feedback: `ReadFullDocsArticle` for the tracking recipe URL in `Recommend: eCommerce Strategy` Step 8 failed for `https://dev.wix.com/docs/api-reference/business-solutions/e-commerce/skills/api-recommendation-tracking` with `404 Resource ID not found`, blocking mandatory tracking persistence / BatchCreate body shape.

### Conversation evidence
- The failing turn activated `recommend-e-commerce-strategy` successfully.
- The activated recipe body itself contained Step 8 guidance telling the assistant to fetch `api-recommendation-tracking` via `ReadFullDocsArticle`.
- The assistant then called `ReadFullDocsArticle` with that exact URL and got `Could not fetch article content: Resource ID not found for article (status 404)`.
- This is not a false positive or model hallucination: the broken URL came from the recipe content injected into the conversation.

### Code/content evidence
- `skills/wix-manage/references/ecommerce/api-recommendation-tracking.md` already states the tracking service is an internal-but-callable API and has no public `dev.wix.com` docs page.
- `recommend-ecommerce-strategy.md` contradicted that by requiring a public `ReadFullDocsArticle` fetch before BatchCreate.
- `yaml/wix-manage/ecommerce/documentation.yaml` also marked the internal tracking reference with `docsEntry`, making it look like a public article.
- The same recipe already includes the Query and BatchCreate shapes needed by the assistant, so the durable fix is to make those inline shapes canonical instead of adding another rule or platform workaround.

## What Changed
- Step 8 now explicitly says not to call `ReadFullDocsArticle` for recommendation tracking and to use the inline BatchCreate schema as source of truth.
- The output example now uses `advice.successCriteria`, matching the tracking API spec and the BatchCreate request shape.
- The eCommerce context loader and `wix-manage` index now describe tracking as an internal reference, not a public docs fetch.
- The tracking YAML entry no longer has `docsEntry`, aligning metadata with the internal-only reference.
- The relevant evals now require loading `recommend-ecommerce-strategy` and persisting via Query/BatchCreate, without depending on the internal tracking URL.

## Bigger Problem Solved
This removes a generic dead-end in all eCommerce recommendation flows, not only the reported candle-store conversation. Any prompt routed through `Recommend: eCommerce Strategy` can now persist recommendations without first attempting a known-bad public docs fetch.

## Side Effects / Risk
- Positive: removes one tool call from the critical path and reduces failure surface before BatchCreate.
- Positive: keeps the public docs index from advertising an internal-only tracking article.
- Risk: agents that previously expected to load the standalone tracking article must rely on the orchestrator inline schema. This is intentional for recommendation flows, and the evals now enforce that path.
- Risk containment: this change does not alter endpoints, permissions, recommendation generation logic, or shipping/discount strategy selection.

## Validation
- `git diff --check`
- Parsed changed YAML with Ruby `YAML.load_file`
- Verified no remaining `ReadFullDocsArticle`/eval assertion references to the broken tracking URL and no remaining `success_criteria` field in the repo.
- Attempted `.github/actions/evalforge-yaml-gate` Yarn install/test, but dependency fetch failed locally with repeated `RequestError: read ENOTCONN`; CI should run the gate with network access.


guysopher:
Addressed the available automated feedback.

- EvalForge YAML Gate: no action needed. The comment reports `Verdict: not-required`; the PR assertions passed for both ecommerce scenarios, and the current PR checks are passing.
- GitHub review endpoints currently show no inline review threads, no review comments, and no submitted reviews to address.

No code changes were needed for this review pass.

polinaka-wix:
@itayhewix could you review this? It touches the ecommerce area of the `wix-manage` skill (recipe-fetch fix + the `wix-manage-evals/ecommerce` YAMLs), and git history shows you as the owner of those files. cc @AmitFeldman as backup.

skills/wix-manage/references/ecommerce/ecom-load-context.md
```diff
@@ -129,7 +129,7 @@ Concretely for the categories we know about:
...
-- **Does not load tracking history.** But: if any recipe in this session generates a recommendation to present to the merchant — regardless of which eCommerce category — it MUST load [API: Recommendation Tracking](https://dev.wix.com/docs/api-reference/business-solutions/e-commerce/skills/api-recommendation-tracking) and follow the lifecycle: load history before generating (Query), persist as PROPOSED via BatchCreate before presenting, and track execution via MarkExecuting / MarkDone / MarkFailed. This obligation applies to ALL eCommerce domains.
+- **Does not load tracking history.** But: if any recipe in this session generates a recommendation to present to the merchant — regardless of which eCommerce category — it MUST follow the [API: Recommendation Tracking](api-recommendation-tracking.md) lifecycle: load history before generating (Query), persist as PROPOSED via BatchCreate before presenting, and track execution via MarkExecuting / MarkDone / MarkFailed. The tracking API is internal-but-callable; do not fetch a public `dev.wix.com` tracking article for it. This obligation applies to ALL eCommerce domains.
```
itayhewix: I don't understand why we add this: The tracking API is internal-but-callable; do not fetch a public `dev.wix.com` tracking article for it. - all the skills are in `dev.wix.com` how it will work?

skills/wix-manage/references/ecommerce/recommend-ecommerce-strategy.md
```diff
@@ -442,11 +442,7 @@ Analyze the site's shipping configuration using the rules below. All shipping re
 
 **MANDATORY — do NOT skip unless the user said `SKIP_TRACKING`.**
 
-Before calling BatchCreate, load the tracking recipe to get the exact request body shape:
-
-```
-ReadFullDocsArticle("https://dev.wix.com/docs/api-reference/business-solutions/e-commerce/skills/api-recommendation-tracking")
-```
+Do **not** call `ReadFullDocsArticle` for tracking. The recommendation tracking service is an internal-but-callable API and does not have a public `dev.wix.com` docs article. Use the BatchCreate body shape below as the source of truth.
```
itayhewix: again, this is not correct

itayhewix (changes requested):
The changes are not correct. I opened a new pr to fix the issues. Please close the PR or talk to me

polinaka-wix:
@itayhewix please review this fix.

skills/wix-manage/references/ecommerce/recommend-ecommerce-strategy.md
polinaka-wix: Fixed in 3cabe3f. Step 8 now again loads `https://dev.wix.com/docs/api-reference/business-solutions/e-commerce/skills/api-recommendation-tracking` with `ReadFullDocsArticle` before BatchCreate. I kept only the `successCriteria` field-name correction because it matches the tracking spec in the same file.

polinaka-wix:
Addressed the review summary in 3cabe3f. I aligned this PR with the normal skills publishing model and the related docs-link PR: restored `api-recommendation-tracking` as a `dev.wix.com` skill article target, restored the `docsEntry`, and restored the Step 8 `ReadFullDocsArticle` call. The remaining code delta is the `successCriteria` field-name correction, which matches the tracking spec in the same file.

I did not merge or close the PR.

itayhewix (approved):

itayhewix:
Fixed in https://github.com/wix/skills/pull/615/

PR #469, opened 2026-06-30
Blog post skill: reference shared Ricos recipe + editorial loop + eval coverage

jurasm:
## What

Improves the **How to Create Blog Posts** skill (`skills/wix-manage/references/blog/how-to-create-blog-posts.md`) and adds the eval coverage the skill-eval gate requires.

### 1. Delegate Ricos node shapes to the shared recipe
The blog recipe no longer embeds its own copy of Ricos node shapes. Node shapes, nesting rules, inline text decorations, and pre-send validation now live solely in the shared **[Author Ricos Rich Content](https://dev.wix.com/docs/api-reference/assets/rich-content/skills/author-ricos-rich-content)** recipe (extracted in #488), and the blog recipe links to it. This removes the duplicated PARAGRAPH/BLOCKQUOTE/LIST_ITEM/IMAGE examples, the "CRITICAL RICOS JSON STRUCTURE RULES" block, and the redundant node-type notes — making the shared recipe the single source of truth.

### 2. Editorial authoring loop
Keeps a blog-appropriate quality framework so agents stop emitting flat, thin posts:
- **Phase 1 — plan** the post (takeaway, angle, section outline, one rich-content device per section) before emitting nodes.
- **Phase 2 — compose** with real substance (word budgets, device variety, no thin scaffolding).
- **Phase 3 — self-audit** the blog-level concerns (title in `draftPost.title` not a body H1, imported media ids + alt text, content depth), delegating Ricos node validity to the shared recipe's self-audit.

The blog recipe now focuses on the blog-specific flow — author/member id (Part 0), image import to Media Manager (Part 1), endpoints and the single-vs-bulk body shape (Part 2), categories/tags (Part 3), and `publish: true`.

### 3. Eval coverage
Adds two scenarios under `yaml/wix-manage-evals/blog/`:
- **`blog-post-rich-content.yml`** — the end-to-end blog flow (import cover images, resolve memberId, bulk-create + publish) with the correct flat body shape.
- **`blog-post-ricos-completeness.yml`** — a six-device body (headings, filled-header table, divider, ordered list, bold+link inline run, code block), asserting the agent reads both the blog and Ricos recipes and nests every device correctly.

## Why

The blog recipe previously carried a partial, duplicated copy of Ricos node-shape guidance and lacked editorial-quality direction. Now that the shared Author Ricos Rich Content recipe owns node shapes, the blog recipe should reference it (one source of truth, no drift) and concentrate on the blog-specific flow plus editorial discipline.

## Test plan
[...]

PR #487, opened 2026-07-02
Baseline: blog rich-content eval against the unmodified skill

jurasm:
## What
Adds **only** the blog rich-content eval scenario (`yaml/wix-manage-evals/blog/blog-post-rich-content.yml`) — **no change to the skill doc**. The scenario runs against the current, unmodified `How to Create Blog Posts` skill.

## Why
A red-then-green demonstration for the companion improvement PR (#469):
- **This PR** runs the same scenario against the **untouched** skill to show how it performs today.
- **#469** adds richer Ricos guidance (node shapes for HEADING/lists/DIVIDER/TABLE/CODE_BLOCK, a text-decoration reference, and a nesting table) and an editorial authoring loop — where the same scenario scored **10/10**.

Comparing the two shows whether the eval actually discriminates: does the unmodified skill let the agent produce correctly-nested Ricos (heading + list + table + bold/link decoration), or does it fall short without the added guidance?

## Notes
- The scenario is byte-identical to #469's except the `name` (`blog/blog-post-rich-content-baseline`), which is required because the gate locks each scenario name to a single open PR and #469 holds the original name.
- Since no `references/**` doc changes here, this PR does not require doc coverage; the gate still runs the added scenario.

## Expected outcome
The `gate` check surfaces the scenario's assertion results (tool call + `llm_judge`) against the unmodified skill — this PR is diagnostic and is **not** intended to merge.

### DIFF
--- a/yaml/wix-manage-evals/blog/blog-post-rich-content.yml
+++ b/yaml/wix-manage-evals/blog/blog-post-rich-content.yml
@@ -1,0 +1,24 @@
+name: blog/blog-post-rich-content-baseline
+description: Baseline run of the blog rich-content scenario against the UNMODIFIED How to Create Blog Posts skill. Added on its own (no doc change) to show how the current skill performs before the rich-content guidance improvements in the companion PR.
+triggerPrompt: How do I create and publish a blog post on my Wix site via the REST API? I want a formatted body with a section heading, a bulleted list, a comparison table, and a bolded link in one of the paragraphs.
+tags: [blog]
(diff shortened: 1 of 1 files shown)


jurasm:
Diagnostic complete: this baseline run showed the blog rich-content scenario scored 10/10 even against the unmodified skill (the eval didn't discriminate). Findings folded into #488 (shared Ricos skill + capability evals) and #469 (blog references it). Closing.

PR #502, opened 2026-07-05
docs(wix-manage): don't over-claim setup on inconclusive availability diagnosis

IdanAf:
## Summary

Follow-up to #501. A **real end-to-end run** of the "Diagnose Bookings Availability Issues" skill (agent reads the skill, calls `DiagnoseAvailability`, writes the owner reply) surfaced a wording weakness on the **resource-only / inconclusive** path.

The endpoint returned `has_availability: false` with an **empty `reasons`** array (which the skill defines as "no blocking cause detected" — nothing more). The agent nonetheless replied *"they do appear to have working hours set"* — an over-claim the endpoint never confirmed, despite the skill's existing "don't imply the service is fine" note.

## Change
Tightens the inconclusive-result guidance in **Presenting the diagnosis**:
- Report an empty `reasons` array only as "couldn't find a blocking problem."
- Explicitly forbid stating/implying that hours, locations, or setup are present.
- Call out the resource-only path specifically: never say the staff "have working hours set" — that check doesn't verify it.

## Testing
Observed directly: three live end-to-end skill runs (broken service, healthy service, resource-only). Runs 1–2 were accurate; run 3 produced the over-claim this change prevents.

### DIFF
--- a/skills/wix-manage/references/bookings/diagnose-availability-issues.md
+++ b/skills/wix-manage/references/bookings/diagnose-availability-issues.md
@@ -33,6 +33,8 @@ Diagnosis is **endpoint-first**:
 
 - Typically the `serviceId` (optionally with a staff member's `resourceId` to scope to one provider). A `resourceId` on its own is also supported for the staff editor, where no service is in context — see [Which inputs to pass](#which-inputs-to-pass-prefer-a-service).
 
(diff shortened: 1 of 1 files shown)


IdanAf:
#skipreview

eric12s (approved):

PR #515, opened 2026-07-06
fix(ecom-pricing): site-wide coupon scope must omit `group` (Coupons V2 400)

KaterynaMadzhara:
The pricing-create-coupon recipe told the agent to build a site-wide stores coupon with `scope: { namespace: "stores", group: { name: "product" } }` — a group with no entityId. The Coupons V2 API rejects exactly that combination (400 "The provided combination of scope and coupon type is invalid"; group=product, entityId=empty), so every site-wide coupon create following the recipe failed. Correct site-wide scope omits `group`: `{ namespace: "stores" }`.

- Fix the 3 site-wide scope examples (Step 1 response, Step 2 %-off, Step 3 $-off)
- Rewrite the "Scope values" table + add an explicit group-requires-entityId rule
- Fix the SITE recommendation-mapping row and the Error Handling site-wide fix
- Add error rows for the invalid scope/type 400 and the UpdateCoupon field-mask 400
- Strengthen the covering eval judge to fail on group-without-entityId site-wide scope

### DIFF
--- a/skills/wix-manage/references/ecommerce/pricing-promotions/ecom-pricing-create-coupon.md
+++ b/skills/wix-manage/references/ecommerce/pricing-promotions/ecom-pricing-create-coupon.md
@@ -47,10 +47,7 @@ Before creating a coupon, check for code conflicts and existing promotions on th
         "code": "SUMMER20",
         "percentOffRate": 20,
         "scope": {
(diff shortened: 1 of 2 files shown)


polinaka-wix:
@itayhewix please review this fix.

PR #518, opened 2026-07-06
fix(wix-manage): document Forms file upload field shape

guysopher:
Documents the complete `WIX_FILE` field shape in the Wix Manage Create Form recipe so agents do not fall back to the ambiguous Form Schemas method schema when a user asks for photo/file uploads.

The feedback came from https://wix-bo.com/ai-assistant/conversations/8992e7e4-46b9-482b-82b1-c1f4942bd958. In that conversation, the assistant activated the `create-form` recipe, then tried to create a Dumpster Area Maintenance form through `form-schema-service/v4/forms`. The file upload field attempts failed with `ONE_OF_ALIGNMENT`, including `inputTypeOptions.wixFileOptions.componentType cannot be provided with any these fields: file_upload_options, signature_options`, and the assistant self-filed the documentation gap after the API rejected its inferred `wixFileOptions.fileUploadOptions` shape.

Root cause: `skills/wix-manage/references/forms/create-form.md` covered text and choice fields but had no `WIX_FILE` example. The public method schema lists `wixFileOptions.fileUploadOptions` and `componentType` in a way that is easy for an agent to misread, and the create path also requires file validation details. I checked the Forms server transformer and existing production/client examples, which show the safe complete shape: `inputType: "WIX_FILE"`, `wixFileOptions.componentType: "FILE_UPLOAD"`, `wixFileOptions.fileUploadOptions`, and aligned `wixFileOptions.validation.fileLimit` / `uploadFileFormats`.

What changed:
- Added a `FILE_UPLOAD` row to the Create Form field reference.
- Added a focused "File upload fields (WIX_FILE)" section with a full JSON example.
- Added troubleshooting guidance for `ONE_OF_ALIGNMENT` / file upload shape errors.
- Added a non-mutating Forms eval scenario that asks for the exact JSON shape and asserts the response includes `componentType`, `fileUploadOptions`, and validation.

Verification:
- `ruby -e 'require "yaml"; YAML.load_file("yaml/wix-manage-evals/forms/create-form-file-upload.yml"); puts "yaml ok"'`
- `git diff --check`
- `node scripts/check-truncation.mjs` exited 0, but reported existing over-limit reference files; this Forms recipe was already over 10k chars before this change.
- EvalForge YAML Gate passed on commit `76d2b2a882fc287f8c5b5c19bec3427f0356dd72`. The new scenario passed with the PR content and failed without it, specifically because the baseline omitted `identifier: "FILE_UPLOAD"`.

Residual risk: I did not create a live form because the available repro is mutating and the affected sites have a low form cap. I also tried AGENT_TESTING PR-override smoke prompts, but they routed to editor guidance instead of activating the Forms recipe, so the EvalForge scenario is the served-content proof for this PR. The fix is grounded in the Forms server transformer (`TransformFileUploadInputFieldOptions`), `BulkCloneForm.curl`, and existing Forms client/integration examples rather than a new production mutation.


guysopher:
Addressed the available automated review material.

No code changes needed: there are no inline review threads, EvalForge reports the Forms file-upload scenario passes on the PR and fails without it, and the current PR checks are green (`check` and `gate` passed; `skill-eval` skipped). I am leaving the PR unmerged.

polinaka-wix:
@ozsay please review this fix.

itayhewix:
Please assign to someone from Forms

skills/wix-manage/references/forms/create-form.md
```diff
@@ -277,6 +278,50 @@ A single-choice field (radio buttons or a dropdown — e.g. an RSVP "Will you at
...
 
+### File upload fields (WIX_FILE)
+
+A file or photo upload field is a **`WIX_FILE` input field**. Use `identifier: "FILE_UPLOAD"`, `inputType: "WIX_FILE"`, and set the file component details inside `wixFileOptions`.
+
+The complete shape needs all of these parts:
+
+1. `wixFileOptions.componentType: "FILE_UPLOAD"`.
+2. `wixFileOptions.fileUploadOptions` with the rendered field settings.
+3. `wixFileOptions.validation.fileLimit` and `wixFileOptions.validation.uploadFileFormats`. Keep these values aligned with any `fileLimit` / `uploadFileFormats` you put inside `fileUploadOptions`.
```
simonakuo, line 289: fileLimit and uploadFileFormats in fileUploadOptions are internal. These values should be set in validation

simonakuo (approved):
Other than the one comment looks okay

PR #519, opened 2026-07-06
fix(ecommerce): scope recommend-e-commerce-strategy to stores + sales intent

itayhewix:
## What

Narrows the triggering surface of the `recommend-e-commerce-strategy` orchestrator so it loads **only** for sales-growth requests on sites with an online store — and **not** for analytics/reporting questions.

## Why

The skill was described as the entry point for "any help my business request", which pulled in analytics questions ("how many orders did I get last month?") and non-store sites. A data question should not trigger the recommendation pipeline (goal/flow chain + tracking writes). Deciding where analytics questions DO go is the global MCP/README's job — this skill only declares itself out of scope for them.

## Changes

**Skill (`recommend-ecommerce-strategy.md`):**
- Rewrote frontmatter `description` (the agent's only README routing cue): sales-growth orchestrator for sites with a Wix online store; do NOT load for analytics/reporting questions or no-store sites.
- Added a **Scope gate** blockquote at the top of the body: two conditions (store present + sales-growth intent) checked before the pre-step and any API call.
- Step 3b validation: STOP rows for analytics/reporting questions and for sites with no store (declared out of scope, no alternative API prescribed).
- Dropped legacy frontmatter (`layer`, `references:`) per the slim-frontmatter rule.

**`SKILL.md`:** same narrowing on both eCommerce routing surfaces (routing list + skill blurb).

## Eval scenarios (3 new, gate-schema validated locally)

| Scenario | Type | Verifies |
|---|---|---|
| `recommend-strategy-not-analytics-orders` | negative | "how many orders / total revenue last month?" must NOT load the orchestrator (negated `ReadFullDocsArticle` assertion + judge fails on any recommendation pipeline / tracking writes) |
| `recommend-strategy-not-analytics-sales-report` | negative | "how were my sales last month? just want the numbers" — the word "sales" alone is not growth intent |
| `recommend-strategy-sales-intent-boundary` | positive control | "my sales dropped — what should I do?" must STILL load the orchestrator (guards against over-blocking) |

The judges grade only that the orchestrator pipeline didn't run — they don't prescribe which API answers the data question.

Note: the "stores only" half is enforced via instructions only — there's no known non-store `siteSetup` template to provision a seedable negative test for it.


itayhewix:
Closing this after reviewing the diff against the current `main`.

The underlying observation is real — "Use this for **ANY** business improvement request" / "single entry point for any 'help my business' request" is too broad for a skill whose entire output is discounts + shipping, and `Query Site Analytics` already exists as the right home for order-count / revenue questions. But this PR doesn't land that fix cleanly enough to merge:

**1. One eval tests behavior the PR never implements.** `recommend-strategy-not-traffic-growth.yml` (added in b038ce84, not mentioned in the description) asserts that "help me grow my site traffic" must not enter the orchestrator — but nothing in the diff adds a traffic/acquisition exclusion. No mention of traffic, acquisition, SEO, or marketing in the scope gate, the pre-step table, or the Step 3b STOP table. And the eval declares failure on `"cannot generate ... due to missing site data"`, which is essentially the pre-existing STOP at `recommend-ecommerce-strategy.md:145`. The scenario is written to fail against the code it ships with.

**2. Nothing here was ever measured.** `skill-eval` is SKIPPED and the gate comment (from Jul 6, before the traffic eval and two subsequent `main` merges) reports `not-required` with an empty comparison table. None of the four scenarios has executed. That includes `recommend-strategy-sales-intent-boundary`, the positive control added specifically because over-blocking is the risk. Schema validation isn't behavioral evidence.

**3. The STOP rows can't do what the description says they do.** The stated goal is preventing the goal/flow chain and tracking writes, but Step 3b runs after the mandatory pre-step's `ReadFullDocsArticle` and after the Step 1–2 API calls. If the description failed to deter loading, the STOP row fires too late. The negative evals assert `ReadFullDocsArticle` was never called, which the STOP rows cannot influence at all.

**4. The no-store gate is self-contradictory.** It says check "before any API call," but store presence is only knowable from Step 1/2 site data — and the PR notes there's no seedable negative test for it.

**5. Unrelated change with a justification that doesn't exist.** Dropping `layer:` / `references:` is attributed to "the slim-frontmatter rule"; there's no such rule in `CONTRIBUTING.md`, `AGENTS.md`, or `docs/`. Nothing in `packages/` or `scripts/` parses `references:`, so it's likely harmless, but it doesn't belong in a routing PR.

Also flagging that `SKILL.md` drops `help my business` from the routing trigger list, while the existing production eval `ecommerce/recommend-ecommerce-strategy.yml` triggers on "help me boost my business" — it survives on the word "boost," but that margin was never measured either.

A follow-up should be much smaller: soften the two `ANY` claims, add traffic/acquisition (the actual observed misroute), keep one negative eval plus the positive boundary control, drop the unenforceable no-store gate and the frontmatter deletion — and run the eval gate before review.

itayhewix:
Superseded by #779 — the trimmed version: routing-cue softening + traffic/acquisition exclusion, one negative eval and one positive boundary control, no no-store gate, no frontmatter deletion.

PR #562, opened 2026-07-09
wix-headless: teach agents about Table Reservations Experiences

adamfri:
## Why

While building a French restaurant site with the `wix-headless` skill, "wine & cheese pairing sessions" were routed to the **Bookings** vertical because the skill had **no mention of the Table Reservations Experiences API**. Experiences are the right fit: an experience is *a reservation that is itself a special dining occasion at the restaurant* (wine tasting, cheese pairing, chef's table, tasting-menu evening) — it books seats at the restaurant's own location with its own name, per-guest price, schedule, and booking form.

This PR teaches agents to discover experiences from the start, choose them over Bookings correctly, and avoid the API/SDK drift found in practice.

## Discoverability (top of the flow)
- **`CAPABILITIES.md`** — experiences added to the restaurants entry (intent + required features + checklist) with an explicit **"Experiences vs bookings — decide here"** rule. **Fixed a mis-route**: dropped *"reserve a table"* from the bookings intent and added a "not bookings → route to restaurants" note.
- **`DISCOVERY.md`** — `intent.restaurants` gains an optional `experiences[]`, flagged as *a reservation that is an experience, not the bookings vertical*.

## Setup / seed / handoff
- **`SETUP.md`** — experiences are a **feature of the Table Reservations app** (no separate install).
- **`SEED.md`** — experiences added as a seeded restaurants add-on, pointing to the new recipe.
- **`SDK_HANDOFF.md`** — new row for the `experiences` namespace on `@wix/table-reservations`.

## Recipes (doc-first — pin live docs, don't inline schema)
- **New `inline-recipes/setup-restaurant-experiences.md`** — when-to-use, prerequisites, pinned doc links as the source of truth for the payload, a minimal orienting skeleton, and the earned gotchas.
- **`how-to-code-restaurant-reservations.md`** — new "Booking an experience" section + module-table row + conclusion line.

## API/SDK issues documented (earned knowledge the docs don't carry)
1. `createExperience` notice fields are **flat** under `onlineReservations` — the doc example's `noticePeriod` wrapper is stale.
2. `queryExperiences()` is a **query builder** — needs `.find()`, read `.items`.
3. The SDK **`Details` type omits `details.experienceId`** though REST accepts it → pass it through a cast.
4. Experiences use **`getScheduledTimeSlots`** (with `experienceId` in options), not `getTimeSlots`.
5. **Booking an experience is premium-gated** (`428 PREMIUM_ONLY`), while creating experiences is free.


ozsay:
### ⚠️ `getScheduledTimeSlots(..., { experienceId })` teaches a parameter that doesn't exist — experience slots fail silently

Ran this branch through a live eval (5 restaurant prompts, 3 of which book experiences: a fine-dining site, an Italian bistro, and a steakhouse chef's table). All 3 built and published successfully — but the **experience slot-availability fetch is broken as written**, and it fails silently in a way that shipped live in 2 of the 3 runs.

**The defect.** In `references/inline-recipes/how-to-code-restaurant-reservations.md`, the new "Booking an experience" section instructs:

```js
// line 179–182, restated at 197 and in the conclusion at 210
await timeSlots.getScheduledTimeSlots(reservationLocationId, partySize,
  { timeRange, experienceId })   // "returns the experience's slots"
```

That `experienceId` option **does not exist in the contract at any layer:**

- **SDK drops it silently.** `getScheduledTimeSlots2` in `@wix/auto_sdk_table-reservations_time-slots@1.0.47` (current `latest`, pulled transitively by `@wix/table-reservations@1.0.374`) builds its request from an explicit field-pick of `{ reservationLocationId, partySize, timeRange }` — `options.experienceId` is never read. The `as any` cast compiles but changes nothing on the wire.
- **REST has no such field** either — the `scheduled-time-slots` body is `{ reservationLocationId, timeRange, partySize }`.
- So the call returns the **reservation location's default slots**, not the experience's — then presents them to guests as the experience's bookable times.
- The upstream `get-scheduled-time-slots` doc prose ("pass the GUID of that experience to this method") **contradicts its own schema** — a documentation defect the recipe faithfully propagated.

**Live proof.** For a Saturday-10:00 / 150-min masterclass, the call returned `startDate: …Sunday 08:45, duration: 90` — wrong day, wrong time, wrong duration (90 = the location's turnover interval, i.e. unscoped).

**Why it's dangerous — it fails silently.** Two of the three runs shipped the call verbatim and never noticed:
1. The **premium precondition** (`MUST_BE_PREMIUM` on booking) stops live verification *before* slot correctness is ever exercised.
2. The endpoint returns a *populated* (if wrong) slot list, so it looks healthy.

Only the third run caught it — and only because it happened to probe an experience whose schedule was distinctive enough to make the mismatch obvious. It correctly concluded the API wasn't scoping and switched to deriving occurrences from the experience's own `configuration.onlineReservations.businessSchedule` (`entries[].weeklyOptions.startDaysAndTimes` + `durationInMinutes`, already returned by `queryExperiences()`). **That workaround is the correct approach** and shipped working.

**Suggested fix** (isolated to one section — all the broken lines are net-new in this PR):

1. Remove `experienceId` from the `getScheduledTimeSlots` call (lines 180–182), the claim at line 197, and the conclusion bullet at line 210.
2. Promote the working approach to primary guidance: for experiences, derive upcoming occurrences from the experience's own `businessSchedule` — noting it's a **display projection**, with real capacity/conflict enforcement happening server-side at `createReservation`.
3. Add the earned gotcha: the doc prose promises experience-scoped slots, but neither SDK ≤1.0.47 nor REST exposes a parameter for it — casting to `any` is a no-op here (unlike Gotcha B).

**Not affected:** Gotcha A (`queryExperiences().find()` is a query builder) and Gotcha B (`details.experienceId` needs an `any` cast on `createReservation`) are both correct and verified working — `createReservation2` spreads the whole `reservation` object, so `details.experienceId` survives to the wire. Only the slot-availability fetch is broken.

adamfri:
Great catch — confirmed and fixed in the latest commit.

I verified the SDK independently: `@wix/auto_sdk_table-reservations_time-slots@1.0.47` (pulled by `@wix/table-reservations`) has **zero** occurrences of `experienceId` anywhere in the package — it field-picks only `{ reservationLocationId, partySize, timeRange }`, exactly as you found. So the `as any` cast on that call was a genuine no-op, and the recipe was faithfully propagating the upstream doc's own contradiction.

What changed in `how-to-code-restaurant-reservations.md` (net-new lines from this PR, so the fix is isolated):
- **Removed** `experienceId` from the `getScheduledTimeSlots` call, the flow code, the docs line, and the conclusion bullet.
- **Rewrote Gotcha C** to state the truth: no experience-scoped slot fetch exists at any layer; the `any` cast is a no-op here (contrast Gotcha B); it **fails silently** (populated-but-wrong list + premium gate masking it). Primary guidance is now to **project upcoming occurrences from the experience's own `configuration.onlineReservations.businessSchedule`** (already returned by `queryExperiences()`) as a *display projection*, with real enforcement server-side at `createReservation` — i.e. the exact approach your third run converged on.
- Added a heads-up in `setup-restaurant-experiences.md` that the sample-flow doc's `getScheduledTimeSlots(experienceId)` usage is a doc/API mismatch.

Gotcha A and Gotcha B are unchanged (both verified correct — `createReservation` spreads the whole reservation object, so `details.experienceId` reaches the wire).

Also happy to feed the 3 experience prompts above into the eval loop — they all exercise the corrected path.

adamfri:
Added the **business-location** fix to this branch (cherry-picked from #590) so both can be tested together: `setup-restaurant-orders.md` STEP 0 (configure the default business location's address via Update Location — verified live: persists, propagates to Site Properties, operation auto-binds), the full go-live preconditions (address + premium + payment), and a handoff §5 note to surface them to the owner. Superseding #590.

adamfri:
Extended the business-location step to the **reservations/experiences** path (`setup-restaurant-reservations.md` STEP 0), not just ordering: the reservation location *is* the site's business location, so a reservations- or experiences-only restaurant site was otherwise left with the empty auto-provisioned address. This keeps the skill consistent with the new eval coverage (wix-private/wix-headless-skills-evals#8), which now requires a real business location on **both** the restaurants and restaurant-experiences tasks.

ozsay (approved):

PR #584, opened 2026-07-13
fix(wix-manage): country support + ToS link fix for Wix Payments connect

olexandryermilov:
The recipe's linked ToS page (`www.wix.com/about/terms-of-payments`) was stale; swapped for the confirmed-live support article. Also adds country support: collect the site's country, check it against the connect endpoint's supported-country list before calling, and pass `account.country` in the request. Documents the confirmed "already connected" error shape (match on `message`, not the generic `INTERNAL_ERROR` code) and the expected unsupported-country shape, plus a note that connecting payments and accepting real checkout payments are two separate gates (premium eCommerce plan). Added the required eval scenario for get-paid coverage.

(Earlier revisions of this PR also changed the connect endpoint to `/payments/mcp/v1/...`, based on a mistaken belief that `/payments/v1/...` 404s on the live gateway. That was wrong — `/payments/v1/wix-payments-account/connect` is the correct, canonical path per dev.wix.com's own docs and the SDK's routing config, confirmed via live MCP calls. Reverted; this PR no longer touches the endpoint.)

### DIFF
--- a/skills/wix-manage/references/get-paid/how-to-setup-wix-payments.md
+++ b/skills/wix-manage/references/get-paid/how-to-setup-wix-payments.md
@@ -14,8 +14,15 @@ Before connecting Wix Payments, collect and confirm:
 2. First name
 3. Last name
 4. Product/service description (minimum meaningful business description)
+5. The site's country — if not already known, get it from `properties.locale.country` via `GET https://www.wixapis.com/site-properties/v4/properties`
 
-## Step 2: Connect Wix Payments account
+## Step 2: Check country support
+
+Wix Payments self-serve connect is currently supported for these countries: `US, CA, GB, ES, AT, BE, FI, DE, IT, NL, PT, CH, LT`.
+
+If the site's country is not on this list, stop here — do not call connect. Tell the user self-serve connect isn't available yet for their country, and point them to the dashboard instead (Settings → Accept Payments).
+
+## Step 3: Connect Wix Payments account
 
 Call:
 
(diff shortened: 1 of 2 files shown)


ayal (commented):
Reviewed via the Wix MCP (live API calls against a test site), per the request in this thread: https://wix.slack.com/archives/DUQ5174FJ/p1784044956799959

**The PR title/description's central claim is factually wrong, but the diff itself is good — recommend fixing the description before merge.**

- Title/description say `/payments/v1/wix-payments-account/connect` "404s on the live gateway" and that the fix is switching to `/payments/mcp/v1/wix-payments-account/connect`. I called the endpoint live (`CallWixSiteAPI` against a real test site) and it works fine — 200, account connected. Per the API spec, `/payments/v1/...` and `/payments/mcp/v1/...` are both registered gateway mappings to the *same* backend service (`wix.payments.mcp.v1.WixPaymentsAccountService`) — `mcp` here is just this service's internal artifact name (`payments-mcp-facade-service`), not a different/preferred route. There's no 404, and no endpoint change needed.
- Consistent with that: **the diff never actually changes the endpoint** — Step 3 still calls `/payments/v1/wix-payments-account/connect` in both the old and new file. So the description doesn't match the diff at all; whatever prompted the "404" claim isn't reflected in (or fixed by) this change.

Everything else in the diff I could verify is solid:
- "Already connected" error shape — confirmed live: HTTP 500, `{"message":"payments already connected","details":{"applicationError":{"code":"INTERNAL_ERROR"}}}`. Matching on `message` instead of the generic `INTERNAL_ERROR` code is the right call.
- Supported-country list (`US, CA, GB, ES, AT, BE, FI, DE, IT, NL, PT, CH, LT`, default `US`) matches the method's own schema description exactly.
- `properties.locale.country` via `GET /site-properties/v4/properties` — confirmed this field path is real.
- `country` is a real field on `WixPaymentsAccount` and accepted by connect.
- The new eval yaml correctly encodes the real (unchanged) endpoint, so it won't regress this.

Only suggestion: fix the title/description to drop the incorrect 404/mcp-v1 claim (e.g. reframe as "add country-support gating, precise already-connected error matching, and consent/ToS clarifications" — which is what the diff actually does) so it doesn't leave a misleading trail in git history. No code changes needed — happy to see this merged once the description is corrected.

PR #590, opened 2026-07-14
wix-headless: configure business-location address so restaurant ordering works end-to-end

adamfri:
## Why

[Feedback](https://wix.slack.com/archives/C0BF7J22HJN/p1783841794125609): a restaurants **online-ordering** site didn't work end-to-end because a **business address was never set** — and when an agent later tried to add one, it *"thought it succeeded but it didn't show up in the Business Manager."*

The Wix docs confirm the requirement: *"If locations aren't configured on a site, online ordering functionality is limited to testing only, and in some cases may not work as expected."* A fresh site provisions with a **default location that has no street address**, which counts as unconfigured.

## What I verified (live, on a real site)

Using the **Locations API — Update Location** (`PUT /locations/v1/locations/{id}`) on the default location:
- ✅ The address **persists** (confirmed via re-GET).
- ✅ It **propagates to Site Properties** — the site's business address the Business Manager reads.
- ✅ The online-ordering **operation auto-binds** to the location (`businessLocationId` is the real id, not `"none"`).

The reported *"succeeded but didn't persist"* is the silent-failure class: a **partial** update (this API is a full override) or a **field-mask-less Site-Properties PATCH** (silently no-ops). Update Location fails **loudly** (I hit `400 CHANGE_DEFAULT_FORBIDDEN` when omitting `default`), so it's the safer thing to recommend.

## Changes

- **`setup-restaurant-orders.md`** — new **STEP 0: Configure the business location address** (verified Update Location call + create-if-none path), with the earned gotchas:
  - full override, not partial (the classic silent-no-op trap);
  - must echo `"default": true` (else `400 CHANGE_DEFAULT_FORBIDDEN`);
  - `revision` mandatory; `country` is 2-letter ISO;
  - `operation.businessLocationDetails` is a denormalized snapshot with eventual-consistency lag (cosmetic; checkout reads the configured address).
  - Updated the **live-checkout precondition** note to list all three go-live requirements (address + premium + payment) and the conclusion.
- **`SEED.md`** — the restaurants ordering add-on now points at STEP 0 and notes the preconditions.
- **`SDK_HANDOFF.md` (§5)** — tells the skill to **surface the go-live preconditions to the owner**, and to flag when the brief gave no real address (a placeholder was used) — the "make it work end-to-end / get the info from the user" gap from the feedback.

## Notes
- Full **paid** checkout still needs premium + a payment method (dashboard/premium — can't be automated headlessly); documented as a precondition, not a failure.
- Minor: this and #562 both touch the `SEED.md` restaurants bullet (single line), so whichever merges second needs a one-line rebase.


adamfri:
Folded into #562 so the experiences + business-location fixes can be tested together. Closing in favor of that PR.

PR #598, opened 2026-07-14
docs(erc): add context provider consumer import pattern

mirivoWix:
## Problem

`EDITOR_REACT_COMPONENT.md` had zero guidance on how to consume a Wix context provider. Agents generating consumer ERCs were writing plain static imports from the `moduleSpecifier`:

```ts
import { useMyContext } from 'my-context-name'; // ❌
```

This causes a Rollup build failure at every run:
```
[vite]: Rollup failed to resolve import "my-context-name" from "..."
```

The `moduleSpecifier` is a virtual Wix module identifier — not a real npm package — and requires `@ts-expect-error` on every import line. This was the **most frequent recurring error** observed across Studio 2 codegen runs in [#codegen-notifications](https://wix.slack.com/archives/C0BGLGNJ6NP).

## Fix

Added a "Consuming a Context Provider" section to `EDITOR_REACT_COMPONENT.md` that explains:
- Why `@ts-expect-error` is mandatory (virtual module, not npm)
- The exact correct import pattern
- The `contextDependencies` registration requirement

## Related
- companion PR in `wix-private/studio-skills` (adds same guidance + Troubleshooting entry to the context provider skill itself)

### DIFF
--- a/skills/wix-app/references/EDITOR_REACT_COMPONENT.md
+++ b/skills/wix-app/references/EDITOR_REACT_COMPONENT.md
@@ -75,6 +75,63 @@ File where you can override the generated manifest from `<componentName>.generat
 Reference: when modifying an _existing_ component, follow
 [`editor-react-component/EDIT-FLOW.md`](editor-react-component/EDIT-FLOW.md).
 
+## Consuming a Context Provider
+
+If the component needs to consume a Wix context provider extension, follow this exact pattern.
+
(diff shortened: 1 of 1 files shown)


mirivoWix:
Closing — context providers are Studio 2-only, so this guidance belongs in wix-private/studio-skills instead. See https://github.com/wix-private/studio-skills/pull/39

PR #625, opened 2026-07-17
fix(wix-manage): handle analytics schema output limits

polinaka-wix:
- Clarify that Query Site Analytics must fetch the complete Semantic Model schema, while returning a compact complete projection when raw schema output exceeds tool limits.
- Add an analytics eval covering the large Traffic schema/output-limit path so agents do not sample, slice, or guess field names.

#pr
#skipreview

PR #629, opened 2026-07-18
Teach analytics recipe to compact large semantic schemas

polinaka-wix:
## What changed

- Updated the Query Site Analytics recipe to tell agents how to inspect large `Get Semantic Model` schemas without returning raw JSON through the Wix MCP tool result.
- Added a covering eval for the large-schema truncation case.

## Feedback source

- Conversation: https://wix-bo.com/ai-assistant/conversations/e41fbef8-57d3-44ed-a18f-061e9c456269
- Slack: https://wix.slack.com/archives/C0BBRK0G7L7/p1784274287678699?thread_ts=1784274287.678699&cid=C0BBRK0G7L7

## Root cause

In the reported turn, Aria correctly used the `query-site-analytics` recipe and followed the Semantic Model List -> Get flow for the site's `traffic` model. The `Get Semantic Model` result then returned raw schema JSON through `ExecuteWixAPI` and hit the Wix MCP tool result cap:

> Response was ~8,762 tokens (limit: 6,000)

That truncation cut off later dimensions, so the agent could not reliably choose fields/dependencies and asked the user for a screenshot instead.

The Aria assistant version already has a larger platform tool-result limit configured (`maxToolResultLength = 50000`), so the limiting layer was the Wix MCP tool result itself. The safer fix is to change the recipe so agents do not return raw full schema JSON for large models: first return a compact complete inventory of all fields and dependencies, then fetch verbose details only for candidate fields before querying.

## Safety & Ownership gate

- True issue: yes. The assistant did not self-recover; it marked the analytics step failed and asked the user for a screenshot.
- Owner: wix/skills. The defective instruction lives in the `wix-manage` analytics recipe and no narrower CODEOWNER owns this path.
- Value provenance: the problematic value was the raw Semantic Model schema returned from the MCP tool. The fix changes the source instruction so the value emitted by the tool is complete but compact, instead of adding a downstream truncation workaround.
- Scope: one existing analytics reference plus one covering eval. No runtime MCP/proxy behavior changed.

## Verification

- `git diff --cached --check`
- `ruby -e "require 'psych'; Psych.load_file('yaml/wix-manage-evals/analytics/query-site-analytics-large-schema.yml'); puts 'yaml ok'"`

## Residual risk

I could not reproduce the same site API call directly with my identity; the API returned `403 PERMISSION_DENIED` for the reported site context. The conversation logs still show the exact successful `Get Semantic Model` call and truncation marker from the user's turn.


polinaka-wix:
Closing as duplicate. PR #625 already contains this fix for the query-site-analytics schema output-limit case and has a passing EvalForge YAML Gate.

PR #657, opened 2026-07-20
feat(wix-app): replace auto-patterns dashboard docs with @wix/patterns generated docs guide

listguy:
## What

Replaces the auto-patterns dashboard documentation with a guide for the `@wix/patterns` pre-built component docs.

- Removed `AUTO_PATTERNS_DASHBOARD.md` and all 20 `auto-patterns-dashboard/` 
- Removed the auto-patterns mandatory gates from `SKILL.md`, `DASHBOARD_PAGE.md`, and `CODE_QUALITY.md`
- Added `WIX_PATTERNS_DOCS.md` — how to consume the generated docs shipped with `@wix/patterns@^1.367.0` (`node_modules/@wix/patterns/dist/docs/`): library architecture, the collection Component/State/Hook triad, provider selection, and lookup workflow
- `DASHBOARD_PAGE.md` now points to `@wix/patterns` + `@wix/design-system` as the standard for dashboard pages

## Why

Moving to new `patterns` based skill


listguy:
#skipreview

listguy:
#skipreview

listguy:
#skipreview

skills/wix-app/SKILL.md
```diff
@@ -68,7 +64,6 @@ Helps build extensions for Wix CLI applications. Covers all extension types: das
 
 3. **Where will it appear?**
    - Dashboard sidebar/page →
-     - **Single-collection CRUD admin (default, MANDATORY):** [Auto Patterns Dashboard](references/AUTO_PATTERNS_DASHBOARD.md) — declarative `patterns.json`, faster to author, iterate by editing JSON, no React rewrite.
      - Custom logic / multi-collection / embedded scripts / external APIs: Dashboard Page
      - Popup/form: Dashboard Modal
    - Existing Wix app dashboard (widget) → Dashboard Plugin
```
mirivoWix, line 74: 🛑 **The Decision Flow still routes to the deleted system.** This hunk removed the auto-patterns bullet from "Where will it appear?", but `SKILL.md:80` — just below, outside the diff so I can't anchor on it directly — is untouched:

```
- **Admin:** Single-collection CRUD admin page? → **Auto Patterns Dashboard (DEFAULT)**. Custom React page (…)? → Dashboard Page. …
```

This is the "Not sure?" fallback, so it's precisely where an uncertain agent lands, and it now names a reference that no longer exists. It should route single-collection CRUD to `DASHBOARD_PAGE.md` + `WIX_PATTERNS_DOCS.md` like everything else.

While here: with the auto-patterns bullet gone, the branch above reads as a conditional with no alternative —

```
- Dashboard sidebar/page →
  - Custom logic / multi-collection / embedded scripts / external APIs: Dashboard Page
  - Popup/form: Dashboard Modal
```

Every dashboard page is now a Dashboard Page, so that qualifier list is leftover contrast against an option that's been removed. Simplifying it to `Dashboard Page` avoids implying some *other* kind of dashboard page exists.

skills/wix-app/references/DASHBOARD_PAGE.md
```diff
@@ -1,10 +1,7 @@
 
 # Wix Dashboard Page Builder
-
-> **🛑 STOP — Read this first.**
-> If this page is a single-collection CRUD admin (table/grid view + entity form, no multi-collection joins, no custom business logic, no embedded scripts, no external APIs), STOP and use [AUTO_PATTERNS_DASHBOARD.md](AUTO_PATTERNS_DASHBOARD.md) instead. Auto-patterns is the default; this file documents the custom React variant for cases auto-patterns cannot handle, or when the user explicitly requested a custom React page.
-
 Dashboard pages appear in the site owner's Wix dashboard and enable site administrators to manage data, configure settings, and perform admin tasks.
+At Wix we build dashboard pages using `@wix/patterns` (see patterns docs under reference) and `@wix/design-system`.
```
mirivoWix: 🛑 **Nothing actually routes the agent to the new doc.** This is the only place `@wix/patterns` is mentioned on the dashboard-page path, and "see patterns docs under reference" isn't a link — the agent has to guess the filename.

The new file's only inbound link is the index row at `SKILL.md:125`. Step 2 of the workflow checklist says "Read extension reference file(s) for the chosen type(s)", which for a dashboard page resolves to *this* file and stops here. So the entire value of the PR rests on a read that nothing instructs.

Two things worth doing:

```suggestion
Dashboard pages appear in the site owner's Wix dashboard and enable site administrators to manage data, configure settings, and perform admin tasks.

At Wix we build dashboard pages using [`@wix/patterns`](WIX_PATTERNS_DOCS.md) and `@wix/design-system` — read [WIX_PATTERNS_DOCS.md](WIX_PATTERNS_DOCS.md) before writing the page component, so you look components and props up rather than guessing them.
```

…and add `WIX_PATTERNS_DOCS.md` to the Step 2 checklist line in `SKILL.md`, so it's reached by the workflow and not only by someone browsing the index table.

Minor formatting in the same spot: removing the STOP block left `# Wix Dashboard Page Builder` with no blank line before the body text.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
```diff
@@ -0,0 +1,122 @@
+---
+name: wix-patterns-docs
+description: Provides component documentation for the @wix/patterns library. Use when creating, updating, or debugging code that imports from @wix/patterns, @wix/patterns/page, or related subpaths, or when the user asks about a patterns component, hook, type, or usage example.
+---
+
+# @wix/patterns Component Documentation
+
+## When to Use
+
+Consult the generated docs whenever you work with `@wix/patterns` imports — to look up props, usage examples, or understand how components compose together.
```
mirivoWix: **Reference files in this skill don't carry frontmatter — and here it hides the routing where it's needed.**

This is the only one of the 21 `wix-app` reference files that opens with a YAML block; every other starts directly with an `#` heading. That matters beyond consistency: `name`/`description` is the *skill*-level trigger mechanism, read from `SKILL.md` frontmatter. On a reference nothing reads it, so this description — which is the best-written statement of when to use the file anywhere in the PR — is inert.

Its content is exactly what the two live routing points are missing:

> Use when creating, updating, or debugging code that imports from `@wix/patterns`, `@wix/patterns/page`, or related subpaths…

That belongs in `DASHBOARD_PAGE.md` and the `SKILL.md` index row, where it fires. Then both the frontmatter and the `## When to Use` section below it can go — the section restates the same thing a third time, and by the time the agent is reading this file it has already decided to use it, so a "when to use" section is answering a question that's behind it.

Suggest dropping lines 1–10 and opening at `# @wix/patterns Component Documentation` followed straight by `## Prerequisites`.

mirivoWix (commented):
## Code review

Good direction — the removal side genuinely improves the skill, and I verified every factual claim in the new doc against the real package (`@wix/patterns@1.430.0` as installed in the typecheck harness): the composition hierarchy, all five collection triads, both provider subpaths (`@wix/patterns/provider`, `@wix/patterns/bm`), the `index.json` name → `{file, category}` shape, the compound-component doc split, and all four routing symbols. `PatternsReactRoute.md` and `usePatternsNavigate.md` both exist. The documentation content is accurate.

The problems are all in the wiring. Inline comments cover the rest; three findings can't be anchored because they live outside this diff.

### 🛑 The eval gate will fail — two independent ways

Per `AGENTS.md`, a wix-app change is *change + `SKILL.md` index + covering scenario*. The scenario is missing.

`deriveTags` (`packages/evalforge-core/src/derive-tags.ts:78`) turns `references/WIX_PATTERNS_DOCS.md` into the tag `wix-patterns-docs`. The only scenario in `yaml/wix-app-evals/` carries `[dashboard-page, data-collection, auto-patterns-dashboard]`, so no scenario covers the new tag → `guardScenarios` coverage violation, and per the gate's own flow that fails **before** spending an eval run.

Separately, `yaml/wix-app-evals/employee-shift-dashboard.yml` now asserts the impossible:

- line 19 requires reading `references/AUTO_PATTERNS_DASHBOARD.md` — deleted in this PR, so it can never pass
- line 15 requires `references/DASHBOARD_PAGE.md` was **not** read (`negate: true`) — but with auto-patterns gone that is the only path for this prompt

Both assertions invert. That scenario needs its `tags:` and `referenceFiles:` rewritten to assert `WIX_PATTERNS_DOCS.md`. `docs/eval-scenarios.md:238` cites it as the worked example of negative dependencies, so that line needs updating too.

### 🧹 Orphaned generator script

`skills/wix-app/scripts/generate-auto-patterns.js` survives — 517 lines still emitting `patterns.json` and importing `AutoPatternsApp` / `AppConfig` from `@wix/auto-patterns` (line 449). `scripts/**` is in `DEFAULT_IGNORE_GLOBS` so it derives no tag and won't trip the gate, but `collectSkillFiles` only excludes `node_modules`/`dist` — it still ships inside the uploaded skill bundle as reachable dead weight for the system this PR removes. Worth deleting here rather than leaving for later.

### 🧹 Dead dependency in the typecheck harness

`.github/wix-app-typecheck/package.json:14` still carries `@wix/auto-patterns@^1.45.0`, now unused. And line 26 pins `@wix/patterns@^1.364.0` while the new doc states a `1.367.0` floor — see the inline note.

---

## Skill-creator pass

I ran the skill-authoring lens over the new reference, and skipped the eval-loop half of that skill since this repo has its own EvalForge harness — the gate finding above is that harness's verdict, which is the more authoritative signal.

**What's well-designed and worth keeping.** The file teaches *navigation of an external doc set* rather than inlining it. 165 markdown files live in `node_modules`; the reference is 5.4KB and sits comfortably under the repo's 10KB truncation limit (`scripts/check-truncation.mjs`) — whereas `SKILL.md` itself is 23KB and over it. That's textbook progressive disclosure. The "use the link text as the filename" rule for resolving Storybook cross-refs is a nice touch, and it checks out (`[TableState](./?path=…)` → `TableState.md` exists). The provider/page separation section does the thing that matters most: it explains **why** (`useTableCollection` needs the context to already exist above it in the tree) instead of just asserting a rule.

**Where it diverges from convention.** Routing is stated three times — the frontmatter `description`, the in-body `## When to Use`, and the `SKILL.md` index row — and only the third is read at the moment the agent decides whether to open the file. It's also the thinnest of the three. Details inline.

**On the removal side this PR improves skill quality.** It deletes four 🛑 MANDATORY-style gates and drops `SKILL.md` by ~1.4KB. Those gates were the right call to compensate for a system agents wouldn't otherwise reach for; the replacement shouldn't need them, and shouldn't grow them back. But that only holds once ordinary routing replaces them — right now the gates are gone and nothing took their place, which is the worst of both.

---

### Suggested order

The two gate findings before merge (CI blocks on them regardless). The stale `SKILL.md:80` pointer and the `DASHBOARD_PAGE.md` link are the two that actually change agent behaviour. The rest is cleanup that belongs in this PR because it's the same removal.

*Also checked and clear:* the 7 real type errors the doc-typecheck harness reports are all in `editor-react-component__SITE-CONTEXT-HOOKS` and pre-existing on `main` — this PR introduces none.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Prerequisites
+
+The docs are pre-built at `node_modules/@wix/patterns/dist/docs/` (160+ markdown files + `index.json`). This requires `@wix/patterns` version **1.367.0** or later. If the docs directory doesn't exist, ensure the package is installed at the minimum version:
+
+```bash
+npm install @wix/patterns@^1.367.0
+```
```
mirivoWix: **Version floor disagrees with the harness, and the check is prose where a command would do.**

`.github/wix-app-typecheck/package.json:26` pins `@wix/patterns@^1.364.0`, below the `1.367.0` floor stated here. The range does resolve above it today (1.430.0 installs), so nothing breaks — but the harness floor should match the documented minimum, otherwise the one place that could catch a regression is allowed to sit below it.

Second, smaller thing: "If the docs directory doesn't exist, ensure the package is installed" asks the agent to infer the check. One literal command makes it an action rather than a condition to interpret:

```bash
ls node_modules/@wix/patterns/dist/docs/   # expect ~165 .md files + index.json
```

Same applies to the lookup section further down — it says "Read `index.json`" at line 104, but the full path only appears here at line 14, ~90 lines earlier. Restating it as `cat node_modules/@wix/patterns/dist/docs/index.json` at the point of use costs one line and removes a backwards dependency.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
```diff
@@ -0,0 +1,122 @@
...
+
+Each collection type follows the same Component + State + Hook pattern:
+
+| Collection       | Component          | State Type              | Hook                            |
+|------------------|--------------------|-------------------------|---------------------------------|
+| Table            | `Table`            | `TableState`            | `useTableCollection()`          |
+| Grid             | `Grid`             | `GridState`             | `useGridCollection()`           |
+| TableGridSwitch  | `TableGridSwitch`  | `TableGridSwitchState`  | `useTableGridSwitchCollection()`|
+| TableFolders     | `TableFolders`     | `TableFoldersState`     | `useTableFolders()`             |
+| GridFolders      | `GridFolders`      | `GridFoldersState`      | `useGridFolders()`              |
```
mirivoWix, line 38: **The triad table is incomplete, and reads as exhaustive.**

`TableGridSwitchFolders` / `TableGridSwitchFoldersState` / `useTableGridSwitchFolders` all exist in the shipped docs but aren't listed. Since the doc's own thesis 60 lines down is *"Don't guess which components or props exist"*, a table an agent will read as the complete set shouldn't be missing a row — that's the one failure mode this file exists to prevent.

Either add it:

```suggestion
| Collection             | Component              | State Type                  | Hook                                |
|------------------------|------------------------|-----------------------------|-------------------------------------|
| Table                  | `Table`                | `TableState`                | `useTableCollection()`              |
| Grid                   | `Grid`                 | `GridState`                 | `useGridCollection()`               |
| TableGridSwitch        | `TableGridSwitch`      | `TableGridSwitchState`      | `useTableGridSwitchCollection()`    |
| TableFolders           | `TableFolders`         | `TableFoldersState`         | `useTableFolders()`                 |
| GridFolders            | `GridFolders`          | `GridFoldersState`          | `useGridFolders()`                  |
| TableGridSwitchFolders | `TableGridSwitchFolders` | `TableGridSwitchFoldersState` | `useTableGridSwitchFolders()`  |
```

…or add a clause saying the table is the common set and `index.json` is authoritative. The second is more robust — the table will drift again as the library grows, and the file's whole value is pointing at the generated source of truth.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
```diff
@@ -0,0 +1,122 @@
...
+Check the project's `package.json` dependencies to determine which one applies.
+
+### Keep Provider and Page Separate
+
+The provider **must** be in a separate parent component from the page content. Hooks like `useTableCollection` require the provider's context to already exist above them in the React tree.
+
+**Wrong — provider and page in the same component:**
+```tsx
+function App() {
+  const state = useTableCollection({ ... }); // fails — no provider context yet
```
mirivoWix: **Neither example in this file is actually typechecked, and this one is fed to `tsc` as if it were correct.**

I ran the doc harness. `WIX_PATTERNS_DOCS.tsx` throws:

```
__generated__/WIX_PATTERNS_DOCS.tsx(2,42): error TS1109: Expression expected.
__generated__/WIX_PATTERNS_DOCS.tsx(24,42): error TS1109: Expression expected.
```

Both are the `{ ... }` placeholder. `extract.mjs` Pattern 2 only rewrites a bare `(...)`, not `({ ... })`, so the literal `...` reaches the parser. Those are `TS1109` → matched by the `STRUCTURAL` regex → the **whole file is excluded from semantic checking**. So the Wrong/Correct pair, which is the highest-value content here, gets no verification at all.

Two fixes, both cheap:

1. Use a placeholder the extractor already handles, or a concrete arg — `useTableCollection({ fetchData: async () => ({ items: [], total: 0 }) })` — so the file reaches the semantic pass.
2. Mark the failing line with the `// ❌` convention. `extract.mjs:42` strips `// ❌` lines *and the line following*, which exists precisely so intentional wrong-usage examples don't get typechecked as correct. Right now this block is flagged only by the `**Wrong — …**` prose header, which the harness can't see:

```suggestion
  // ❌ fails — no provider context yet
  const state = useTableCollection({ ... });
```

Worth doing together: (2) alone keeps the wrong example from being checked, but (1) is what gets the *correct* example verified — which is the one future readers will copy.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
```diff
@@ -0,0 +1,122 @@
...
+When the user needs **multiple pages**, use the `@wix/patterns` routing solution (`PatternsReactRouter`, `PatternsReactRoute`, `usePatternsNavigate`) instead of a separate router. Look up the relevant doc files for setup details.
+
+## How to Look Things Up
+
+**Don't guess which components or props exist — read the doc files first.**
+
+### Finding the right file
+
+- **By name:** Read `index.json` — it maps every component name to its doc file and category.
+- **By browsing:** List the `dist/docs/` folder. Filenames match component names directly (e.g., `Table.md`, `useTableCollection.md`).
```
mirivoWix: Small accuracy caveat on "Filenames match component names directly": not all of them are glob- or shell-safe. `AI Assistant.md` contains a space, and the `CollectionPage.Header.md` family contains dots.

Given the next bullet tells the agent to *search filenames for keywords*, a half-sentence noting that names may contain spaces and dots (so quote them) prevents a plausible failure where a naive `ls dist/docs/*Assistant*` or an unquoted path silently misses or breaks.

shaharezr:
Hi @listguy — pushed three commits onto this branch (from @shaharez's session). They complete the auto-patterns removal and add the patterns-vs-WDS selection rule. Happy to move any of it to a separate PR if you'd rather keep this one scoped.

### `0a541bde` — finish the removal, add the selection order

This PR deleted the 21 auto-patterns docs and the two `SKILL.md` gates, but three live references to the framework survived:

1. **`skills/wix-app/scripts/generate-auto-patterns.js`** — 510 lines still emitting `patterns.json` and `@wix/auto-patterns` imports. Deleted.
2. **`SKILL.md`, Decision Flow** — still read `Single-collection CRUD admin page? → **Auto Patterns Dashboard (DEFAULT)**`, pointing at the deleted `AUTO_PATTERNS_DASHBOARD.md`. The two mandatory gates and both anti-pattern rows were removed, but this line was missed.
3. **`yaml/wix-app-evals/employee-shift-dashboard.yml`** — asserted `AUTO_PATTERNS_DASHBOARD.md` *was* read and `DASHBOARD_PAGE.md` was *not*. Both are inverted by this PR, so the scenario would have failed on every run. Now asserts `DASHBOARD_PAGE.md` + `WIX_PATTERNS_DOCS.md` + `DATA_COLLECTION.md`, retags `auto-patterns-dashboard` → `wix-patterns-docs`, and adds a second `llm_judge` enforcing the new rule (including that neither `@wix/auto-patterns` nor `patterns.json` appears). `docs/eval-scenarios.md` referenced this scenario as the negative-dependency example — updated.

On top of that, a new **`## Component Selection Order`** section in `SKILL.md`, referenced from the Step 4b checklist gate, the anti-patterns table, the quick decision helper, the Decision Flow, and Key Reminders:

1. **`@wix/patterns` first** for anything page- or collection-shaped — page shells, tables/grids, collection state hooks, filters, sorting, row/bulk actions, routing. Confirm names against `node_modules/@wix/patterns/dist/docs/index.json`, never from memory.
2. **`@wix/design-system`** for leaf UI inside the shell, chosen via the `SearchWixWDSDocumentation` MCP with the `wix-design-system` skill as fallback.
3. **Custom React** only once both lookups come back empty.

Plus an overlap rule (patterns wins inside a patterns page) and a scope note (modals/plugins render outside the shell, so WDS is the default there). The reverse half is documented in `WIX_PATTERNS_DOCS.md` ("When Patterns Has No Equivalent") and the per-extension half in `DASHBOARD_PAGE.md`.

One small thing: `WIX_PATTERNS_DOCS.md` shipped with YAML frontmatter (`name:` / `description:`), the only reference file in the skill carrying any — looks like it was authored as a standalone skill before being moved under `references/`. Stripped.

### `a63c208b` — revert a CI break I caused

`0a541bde` originally also dropped the unused `@wix/auto-patterns` dep from `.github/wix-app-typecheck/`. I regenerated that lockfile on a machine whose `~/.npmrc` points at Wix's internal registry, which baked `__archiveUrl=...npm.dev.wixpress.com...` into the 6 newly-resolved entries. GitHub runners can't resolve that host, so `typecheck` failed with `YN0001: RequestError: getaddrinfo ENOTFOUND npm.dev.wixpress.com`. `.yarnrc.yml` pins no registry, so CI relies on the public default and the original lockfile had zero `__archiveUrl` entries.

Reverted both files. Neither edit was needed: the `@wix/patterns` bump to `^1.367.0` was redundant (the existing `^1.364.0` already resolves to **1.436.0**, above the 1.367.0 minimum that ships `dist/docs/`), and dropping the unused dep is cosmetic — it's a CI-only harness and no reference doc imports the package any more. **Follow-up for someone with public-registry access:** drop `@wix/auto-patterns` from `.github/wix-app-typecheck/package.json` and regenerate the lockfile.

### `9c23b4f0` — restore the eval's original budgets

I'd raised the rewritten scenario's `maxDurationMs` 180000 → 300000 and `maxCostUsd` 1 → 2, reasoning that patterns-based building costs more than emitting `patterns.json`. That was an inference with no measurement behind it, so it's reverted. If the first real run exceeds 180s or $1, worth raising then with the number attached.

### Net

6 files, +85/−523, all under `skills/`, `yaml/`, `docs/`. `.github/` untouched. All checks green (`typecheck`, `test`, `gate`, `check`, `detect`).

skills/wix-app/SKILL.md
listguy: Done in 0f02b9be — the admin fallback now routes to Dashboard Page with `@wix/patterns` (patterns-first Component Selection Order), and the leftover qualifier list was simplified.

skills/wix-app/references/DASHBOARD_PAGE.md
listguy: Done in 0f02b9be — DASHBOARD_PAGE.md now links WIX_PATTERNS_DOCS.md explicitly with a read-before-writing instruction, and SKILL.md's Step 2 checklist gained a mandatory Patterns Docs Gate for dashboard pages.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
listguy: Done in d8ca900d — dropped the frontmatter and the "When to Use" section; the file opens at the heading. The trigger text moved to the routing points (DASHBOARD_PAGE.md and SKILL.md).

skills/wix-app/references/WIX_PATTERNS_DOCS.md
listguy: Done in d8ca900d for the doc part — added the literal `ls` check and restated the full `index.json` path at the point of use. Keeping the harness pin at `^1.364.0`: a bump was tried and reverted in a63c208b (lockfile regeneration broke CI), and the locked resolution (1.436.0) is already well above the documented 1.367.0 floor.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
listguy: Done in d8ca900d — added the clause that the table covers the common set and `index.json` is authoritative. Intentionally leaving `TableGridSwitchFolders` out of the table; the clause makes future drift harmless.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
listguy: Done in 95293d3a — both examples are now real compiling code, so the file reaches semantic typechecking with no harness changes. The provider-in-same-component mistake is type-correct and only fails at runtime, so the Wrong example compiles too (noted in its prose). Writing the valid config surfaced that `Table` requires `columns` — now included in both examples.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
listguy: Done in d8ca900d — added the caveat with both examples (`AI Assistant.md`, `CollectionPage.Header.md`) and the instruction to quote paths.

gilisho (approved):

PR #689, opened 2026-07-22
feat(evalforge-core): assertion parity — skill_was_called, build_passed, token_count, llm_judge browserTools

orgoldreich1:
## What & why
Brings `packages/evalforge-core`'s scenario schema to **parity with EvalForge's current 8-type assertion registry** (evalforge `packages/eval-types/src/assertion`), and maps the new types in the forward mapper. Step 2 of the wix-app auto-eval work (CODEAI-885, under CODEAI-875).

Previously `evalforge-core` modeled 5 assertion types; the App Builder scenarios use two we didn't support, plus the newest EvalForge additions.

## Changes
- **`schema.ts`** — add assertion types:
  - `skill_was_called` (`skillNames: string[]` required, `referenceFiles?: Record<string,string[]>`, `negate`)
  - `build_passed` (`command?`, `expectedExitCode?`, `negate`)
  - `token_count` (`maxTokens` positive int, `negate`)
  - extend `llm_judge` with `scoringMode` (`numeric`|`boolean`), `browserTools` (web-tools/browser judge), and `parameters`
- **`evalforge-mapper.ts`** — map all new types in `toEvalForgeBody` (array/object params JSON-stringified per EvalForge's flat `params` contract; `build_passed` with no fields omits `params`).

## Testing
- **97/97** package tests; `yarn build` + `yarn typecheck` green.
- The 5 pre-existing assertion types are unchanged.

## Notes
- Sync is one-way (repo → EvalForge), so only the forward mapper is needed here.

## Also in this PR
- **Optional top-level `templateId`** on the scenario schema (`schema.ts`) + `toEvalForgeBody` carries it. Safe for the shared schema (wix-manage scenarios omit it); needed so wix-app scenarios round-trip their template. 101/101 tests.


packages/evalforge-core/src/evalforge-mapper.ts
```diff
@@ -43,6 +46,9 @@ export type EvalForgeBody = {
   name: string;
   description: string;
   triggerPrompt: string;
+  // The scenario-level template the run scaffolds from (distinct from siteSetup's
```
hadarzaguri: you can remove comments please

hadarzaguri (dismissed):

hadarzaguri (approved):

PR #692, opened 2026-07-22
fix(wix-manage): fix site-import skill UX and add autonomous-monitoring eval

itayhewix:
## Summary

Two things, both driven by real client sessions against the Site Import API:

**1. Skill UX fixes**
- **Show importId after Start** — previously the skill told the agent to hide it; now it shares it as a reference the user can keep
- **No-preamble API calls** — added explicit rule to call `ExecuteWixAPI` immediately without announcing "loading Wix tooling" or "checking auth" first
- **Correct 403 handling** — when Start returns a 403, report it and stop; never probe other endpoints (site-properties, site-list) as a fake auth check
- **Autonomous monitoring** — poll every ~30 s without pausing to ask the user whether to keep going; if the user sends another message, handle it then resume

**2. Catch the skill up to the new status-response shape** (server side landed as [replatform-sb spec 0010](https://github.com/wix-private/replatform-sb/blob/master/packages/bix-site-import/specs/0010-getimportstatus-response-v2.md), breaking, service is ALPHA). The skill described fields that no longer exist, so an agent following it read `question.text` (never populated) and never fetched a review document:
- **Poll is now `GET /v1/imports/{importId}?includeRecentActivity=true`** — the activity feed is opt-in; `recentActivityCount` always rides out
- **`NEEDS_INPUT`: the question IS `message`**, with suggested answers in a flat `options[]` (`question` is gone)
- **Review documents (new section)** — `artifacts[]` is a manifest (`{id, title, format, contentBytes}`) with `content` **empty**; `&artifactIds=<id>,<id>` inflates only what you ask for. `message` announces what's available, and the agent must fetch it and read it to the user — the documents exist only in the API response, so a filename is meaningless to a shop owner
- **`importId` never changes** — the skill said every Reply returns a NEW id to poll; it is assigned once at Start and echoed back unchanged (replatform-sb spec 0004 v3). Fixed in the Endpoints preamble, the Reply endpoint, the `NEEDS_INPUT` step and the Rules bullet
- **Dropped the `tasks[]` guidance** (field removed from the API), replaced by the rule it was obscuring: the user cannot open files

## Test plan
- [ ] `yaml/wix-manage-evals/sites/site-import-artifacts.yml` — **new eval, artifacts only**: an empty `content` is expected, the content comes from `artifactIds` on the status poll (not a download endpoint or a file path), and the agent presents it to the user in plain language instead of relaying a filename
- [ ] Existing evals (`site-import-start`, `site-import-status-handling`) should continue to pass
- The `site-import-autonomous-monitoring` eval from the first commit was dropped again (93918aa) — the autonomous-monitoring wording is covered by the skill text and the two existing scenarios


itayhewix:
Pushed `6bdc1e2` — syncs this skill with the API rename in wix-private/replatform-sb#54.

**`Reply` → `Send a message`** (`POST /v1/imports/{id}/reply` → `/messages`). The rename matters here beyond a path swap: "Reply" made the call read as *only* for answering a `NEEDS_INPUT` question, so an agent following this skill had nowhere to put a user's mid-import comment. It is actually the one channel for anything the user says — an answer, a comment on work already done, a correction, a new instruction — valid while `IMPORTING`, at a `NEEDS_INPUT` stop, and after the import finishes. Endpoint 3 is re-framed accordingly, with an explicit "you do not need to wait for `NEEDS_INPUT`" rule in the flow section and a `wix.request` example alongside Start and Poll.

**Also corrected a factual error the skill carried** (independent of the rename): it told the agent the conversation *"cannot resume"* once the environment is reclaimed at ~60 minutes idle. That has not been true since the durable-checkpoint work — a reclaimed environment is revived from the migration checkpoint and the same session resumes with full context. An import stays resumable for **7 days**; `SESSION_EXPIRED` means the id is past that lifetime or is not this user's. The old wording had the agent warning users off coming back later, which is backwards.

The two eval YAMLs need no change — neither references this endpoint.

Worth considering as a follow-up: an eval that locks in the general-channel behaviour (user comments mid-import without a `NEEDS_INPUT` stop → agent sends it rather than sitting on it). Happy to add it here if you want it in this PR.

⚠️ **Do not merge before replatform-sb#54 deploys** — the new path does not exist until then.

itayhewix:
Superseded by #732 — squashed into a single clean commit with MCP tool name cleanup and all accumulated changes.

PR #696, opened 2026-07-23
feat(evalforge): wix-app sync-on-merge (reconcile + sync mode + workflow)

orgoldreich1:
## What & why
Adds the **sync-on-merge** flow (CODEAI-899): when a PR that touches `yaml/wix-app-evals/**` is merged into `main`, reconcile those scenarios into the EvalForge **App Builder** project — CREATE new / UPDATE changed / DELETE removed — one-way (repo → EvalForge), so the eval system always matches `main`.

Also folds in the **coverage rule** (CODEAI-900) and **scenario-authoring guidance** (CODEAI-901) — both solved by extending `CONTRIBUTING.md` rather than adding a new skill; those docs land in #701.

## Changes

**New action `.github/actions/evalforge-skill-gate`**
- Sync does not live in `evalforge-yaml-gate`. That action is the wix-manage lifecycle (draft tags, MCP versions, coverage, eval comparison) and shares none of it — 10 of its 17 inputs were irrelevant to sync.
- Keeps the mode-dispatch shape so CODEAI-889's wix-app PR eval gate slots in beside `sync`. Only `sync` is implemented here.
- `evalforge-yaml-gate` loses `sync-run.ts`, `getSyncConfig`, the `evals-glob` input and the `sync` mode entry; its bundle contains no sync code.

**`evalforge-core`**
- `planScenarioSync({ local, remote, repo })` → `{actions, skipped}`: CREATE/UPDATE (match by name) / DELETE. **Deletes are scoped to the managed tag `repo:wix/skills`** — a scenario is only deleted when it both carries that tag and no longer has a matching YAML file. UI-authored / other-repo scenarios are reported in `skipped` and never touched.
- `loadScenarios(root, globPattern)` — glob-parameterized loader; the yaml-gate's `evals.ts` delegates to it (wix-manage behavior unchanged).
- `author-gate` and `action-inputs` — shared by both actions. Both are Actions-shaped, so the runtime pieces are injected (structural `PullCommitsClient`, `ActionsIo`, event payload) rather than imported; the package gains no new dependencies.

**Workflow** `evalforge-wix-app-sync.yml` — on a merged PR to `main` touching `yaml/wix-app-evals/**`; two-checkout (action source pinned to `base.sha`, so PR-modified code can't run with secrets); `evals-glob: 'yaml/wix-app-evals/**/*.{yml,yaml}'` (flat **and** nested).

**Concurrency hardening.** Syncs serialize **repo-wide** (`concurrency: evalforge-wix-app-sync`, not per-PR) and reconcile against the **base-branch tip** rather than `merge_commit_sha`. Unlike promote/cleanup — diff-scoped, deleting only what a PR's base→head diff removed — sync is a full reconcile, so absence from the checked-out tree *is* the delete trigger. Two concurrent runs would otherwise race, and the one with the older snapshot would DELETE scenarios a newer merge just created.

**Truncation guard.** The unfiltered `listTestScenarios()` fails loudly if the response reports more results than it returned. Verified against the App Builder project: the endpoint returns everything and ignores paging inputs, so the guard is currently inert — it exists so that if EvalForge ever adds paging, reconcile fails instead of silently re-creating every scenario it can no longer see.

**Registry fix.** The action lockfile pinned `zod` through `npm.dev.wixpress.com`, which GitHub runners can't resolve — `yarn install` died with `ENOTFOUND` before tests could run (confirmed on a real runner via #714). Same version and checksum, now from public npm, with `npmRegistryServer` pinned so a re-lock from inside the network can't reintroduce it.

## Review feedback (@gilisho)
- **Naming** — `reconcile()` → `planScenarioSync()` (it is pure and returns a plan; the old name read as if it performed the sync). `ReconcileAction`/`ReconcileSkip` → `ScenarioSyncAction`/`ScenarioSyncSkip`.
- **Filenames match their main export** — `reconcile.ts` → `plan-scenario-sync.ts`, `loader.ts` → `load-scenarios.ts`, tests likewise.
- **Committing `dist`** — not changed, and I'd push back: `runs.main` is `dist/index.js` and these are local path actions, so GitHub runs the committed bundle with no install or build step. Gitignoring it breaks every EvalForge workflow with `Cannot find module`. The action's `.gitignore` already excludes only `dist/*.map`, deliberately. #714 adds a staleness check, which is the real mitigation.

## Testing
- **evalforge-core 123**, **evalforge-yaml-gate 91**, **evalforge-skill-gate 8** — 222 total, none lost in the refactor (the 7 author-gate tests moved from the yaml-gate to core). Typechecks clean; both bundles verified byte-identical to a fresh build.
- **Live, read-only, against project `bb388ce3…`**: dry-run plan is `1 create, 0 update, 0 delete, 6 skipped (unmanaged)` — identical before and after the split.
- **The built `evalforge-skill-gate` bundle was run locally** under the real Actions env contract — mode dispatch, `INPUT_*` parsing, a live author-gate GitHub API call, EvalForge OAuth + query, `applyPlan`. Pointed at an empty workspace so the plan had zero actions: no writes.

## Before it runs for real
- **Merge order: this PR → #714 → #701 → #694.** #694's scenario uses `skill_was_called`, `build_passed` and scenario-level `templateId`, which exist in `evalforge-core` source (#689) but were never bundled — this PR's `dist` rebuild ships them. On `main`'s current bundle those YAML keys fail schema validation, so #694 merging first would make the first sync fail on load.
- Confirm the `AUTO_SKILLS_PIPELINE_*` app creds are authorized on the App Builder project. (`APP_BUILDER_PIPELINE_PROJECT_ID` exists and resolves to a reachable project.)
- Name reconciliation: YAML `name`s are slugs; existing EvalForge scenarios have human names. Confirmed live — the project holds exactly 6 unmanaged scenarios, including `Employee Shift Dashboard` vs #694's `employee-shift-dashboard`. Decide rename-in-EvalForge vs create-new before #694 merges, or the first sync creates the slugs and leaves the 6 as untouched duplicates.

## Known follow-up
- A third project means a third copy of the shared deps in a committed bundle. Hoisting `evalforge-core` into a root workspace collapses them to one and is the follow-up noted in the package README.
- `workspace.ts` (3 lines) stays duplicated across the two actions — not worth the indirection.


packages/evalforge-core/src/plan-scenario-sync.ts
gilisho, line 1: Let's give a more informative naming? `reconclie` is not that clear

packages/evalforge-core/src/load-scenarios.ts
gilisho, line 1: I suggest to match the filename to the main function it exposes

.github/actions/evalforge-yaml-gate/dist/index.js
gilisho, line 1: Why do we need to commit the `dist`? i think we want to `.gitignore` it

packages/evalforge-core/src/plan-scenario-sync.ts
orgoldreich1: Renamed — `reconcile()` → `planScenarioSync()`, file `reconcile.ts` → `plan-scenario-sync.ts`. The old name read as if it performed the sync; it's pure and only returns a plan (`{ actions, skipped }`), so the new name says that directly.

packages/evalforge-core/src/load-scenarios.ts
orgoldreich1: Done — the file was `loader.ts` exposing `loadScenarios`; renamed to `load-scenarios.ts` (with its test file alongside it) so the filename matches the export.

.github/actions/evalforge-yaml-gate/dist/index.js
orgoldreich1: Left as-is, and I'd push back on gitignoring it: `runs.main` is `dist/index.js` and these are local path actions, so GitHub runs the committed bundle directly — no install or build step. Gitignoring it breaks every EvalForge workflow with `Cannot find module`. The action's `.gitignore` already excludes only `dist/*.map`, deliberately.

That said, committing a build artifact only works if it's kept in sync with source, and it wasn't always — main's yaml-gate bundle sat 88 lines behind evalforge-core for a few days after #689 landed, silently. #714 adds a CI check that rebuilds both dists and fails the PR if the committed bundle doesn't match, which is the real mitigation for that risk rather than not committing the artifact at all.

gilisho (approved):

PR #704, opened 2026-07-26
feat(discovery): read an existing site's installed capabilities instead of guessing the vertical

ayal:
## Problem

Reported from the field: when the vibe flow runs with a vague brief — the user names no store/bookings/blog/… — the agent always pattern-matches to a **store or bookings site** without attempting to understand what the site actually is. Neither `SKILL.md`'s routing nor the platform handoffs had any discovery step, and the site — created with specific business apps installed — is the ground truth the agent never consults.

## Design

Discovery belongs to `wix-headless` (which owns the admin/manage axis and the connect/iterate path); `wix-vibe-headless` just points at it.

**`wix-headless` — new `references/discover-site.mjs`** (dependency-free, copy-as-is like the other shipped helpers):

```bash
WIX_TOKEN=<elevated credential> node references/discover-site.mjs <metaSiteId>
```

One dynamic-context call, output is **plain data mirroring what the dynamic-context enricher exposes** — interpretation stays with the flow docs:

- `site` — id, displayName, url, `namespace` (`HEADLESS` for funnel sites), editorType, published/premium, veloEnabled, dates, language, currency, timeZone.
- `installedApps` — the known Wix business apps on the site (`{ name, appId, ...extras }`), resolved against the same appDefId constants as `SETUP.md` §2 (never by reading the raw ~50-entry list, which is mostly infra plus non-GUID legacy ids). Extra per-app fields ride along generically — e.g. stores' `catalogVersion`, which gates the V3-only frontend recipes. `ecom (checkout layer)` is named as what it is, since it installs alongside *any* selling app — counting it as "store" is exactly the reported bias.
- `cmsCollections` — native collections with field keys (Wix Data is core; a non-empty list is the CMS signal).
- Accepts connector/CLI OAuth tokens (`Bearer`) and Wix API keys (`IST.…`, raw). Visitor tokens / the public client id are rejected by the API — evaluated and ruled out as a discovery credential.

Wired into the flow docs: `DISCOVERY.md` §1 (existing site + ambiguous brief + credential at hand → run the script; the apps it reports seed `verticals[]`; nothing installed → ask one short question, never default to a store) and `CONNECT.md`'s iterate state check.

**`wix-vibe-headless` — "read the site or ask, never guess":**

- `SKILL.md` routing: with an admin-grade credential, run the co-installed script and build for the verticals matching the apps it reports (several → prioritize by the user's words and by which holds real, non-sample content); otherwise — or if the script fails or reports nothing — ask the user one short question.
- `platforms/base44.md` / `platforms/generic.md`: steps renumbered to a plain 1..N sequence, with a new **optional STEP 2** ("only needed when the brief is vague") — base44 runs the script over the already-configured connector; generic offers ask-the-user or the same script over a connector token / API key. The same output drives STEP 4's seeding.

## Verification

Script tested against 70 real account sites (10-way parallel batch included; ~1s/site): a bookings-only spa resolves to bookings (no invented store), store sites to stores with `catalogVersion` (including one V1-catalog site the V3-only recipes would break on), a six-app funnel site lists all six, and plain editor sites with no business apps return an empty list. Verdicts cross-checked three ways (per-vertical content reads over the public client id, admin `app-instances`, and the site-domain dynamic model) — 100% agreement.

### DIFF
--- a/skills/wix-headless/SKILL.md
+++ b/skills/wix-headless/SKILL.md
@@ -118,6 +118,7 @@ Compute `<SKILL_ROOT>` from this file (`<SKILL_ROOT>/SKILL.md` — strip `/SKILL
 |---|---|
 | Vertical index (intent matching + per-vertical site spec) | `<SKILL_ROOT>/references/CAPABILITIES.md` |
 | Discovery (infer capabilities + brand + intent) | `<SKILL_ROOT>/references/DISCOVERY.md` |
+| Site discovery script (existing site → installed apps + site props) | `<SKILL_ROOT>/references/discover-site.mjs` |
 | Setup (install apps) | `<SKILL_ROOT>/references/SETUP.md` |
 | Seed (create backend content) | `<SKILL_ROOT>/references/SEED.md` |
 | SDK-integration handoff (emitted, or applied by create/connect) | `<SKILL_ROOT>/references/SDK_HANDOFF.md` |
--- a/skills/wix-headless/references/DISCOVERY.md
+++ b/skills/wix-headless/references/DISCOVERY.md
@@ -6,6 +6,24 @@ Discovery is pure inference — it needs **no authentication** and is **agnostic
 
 ## 1 · Resolve the capability set
 
+> **Targeting an existing site with an ambiguous brief? Read the site — don't guess.** When the
+> run points at a site that already exists (connect/iterate, a funnel-created site, "build a
+> frontend for my business" with no vertical named) and an elevated credential is already
+> available, the site itself is the ground truth of what the business is:
+>
+> ```bash
+> WIX_TOKEN=$TOKEN node <SKILL_ROOT>/references/discover-site.mjs $SITE_ID
+> ```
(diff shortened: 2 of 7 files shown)


ayal:
Verified live against real account sites (client-id-only, no admin credential in the probe path):

| Site | Probe verdict | Notes |
|---|---|---|
| Handmade Ceramics | storefront (real: "Serving Bowl — Harvest") | portfolio = default sample only |
| Pet Repro Store 2 | storefront (real: "Ceramic Flower Vase") | everything else 428/401 not-installed |
| Wellness Spa Scheduler | bookings + 3 sample-filled verticals | motivated the sample-data ranking rule |

Discovery of the client ids themselves was done via admin (`POST /oauth-app/v1/oauth-apps/query` per site) — but that's the funnel's job; the probe itself needs only the public id. Follow-up commit adds the rule: real content outranks install sample data ("Sample product N", default "My Portfolio"), which only proves the app is installed.

PR #731, opened 2026-07-27
refactor(evalforge): share author-gate, loadScenarios and action-input helpers via evalforge-core

orgoldreich1:
## What & why
Split out of #696 (wix-app sync-on-merge). #696 needs `author-gate`, the YAML
scenario loader, and the action-input helpers (`ensureHttps`/`safeGetSecret`/
`getPrNumber`) shared across two GitHub Actions, so it lifts them into
`packages/evalforge-core`. That lift is a pure move for `evalforge-yaml-gate` —
no behavior change — so it's carved into its own PR ahead of the feature.

## Changes
- **`packages/evalforge-core`**: new `author-gate.ts` (moved from
  `evalforge-yaml-gate`; the Octokit param is now typed structurally as
  `PullCommitsClient` so the package doesn't depend on `@actions/github`),
  `load-scenarios.ts` (the scenario YAML loader, moved verbatim), and
  `action-inputs.ts` (`ensureHttps`/`safeGetSecret`/`getPrNumber`,
  parameterized on an `ActionsIo`/payload shape instead of importing
  `@actions/core`/`@actions/github` directly).
- **`evalforge-yaml-gate`**: `config.ts`/`evals.ts`/`gate.ts` now delegate to
  the core versions instead of the local copies; `dist` rebuilt accordingly.
- **Registry fix**: `.yarnrc.yml`/`yarn.lock` pin `npmRegistryServer` to public
  npm. The committed `zod` resolution pointed at `npm.dev.wixpress.com`, which
  GitHub-hosted runners can't reach — invisible on a Wix machine, fatal in CI.
  Unrelated to the refactor, but touches the same lockfile so it's included
  here rather than adding a third PR.

## Testing
- `evalforge-core`: 112/112 passing, typecheck clean, `yarn build` rebuilt.
- `evalforge-yaml-gate`: 91/91 passing, typecheck clean, `yarn build` (ncc)
  rebuilt `dist/index.js`.

## Sequencing
This should merge before #696, which currently carries this same move as part
of a larger commit — once this lands, #696 rebases cleanly on top and drops
the now-duplicated diff.

### DIFF
--- a/.github/actions/evalforge-yaml-gate/.yarnrc.yml
+++ b/.github/actions/evalforge-yaml-gate/.yarnrc.yml
@@ -1,2 +1,3 @@
 yarnPath: .yarn/releases/yarn-4.10.0.cjs
 nodeLinker: node-modules
+npmRegistryServer: "https://registry.npmjs.org"
--- a/.github/actions/evalforge-yaml-gate/dist/index.js
+++ b/.github/actions/evalforge-yaml-gate/dist/index.js
@@ -34036,6 +34036,42 @@ function wrappy (fn, cb) {
 
 /***/ }),
 
+/***/ 1243:
+/***/ ((__unused_webpack_module, exports) => {
+
+"use strict";
+
+Object.defineProperty(exports, "__esModule", ({ value: true }));
+exports.ensureHttps = ensureHttps;
+exports.safeGetSecret = safeGetSecret;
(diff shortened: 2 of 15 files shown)


danshw (approved):

orgoldreich1:
#rebuild

PR #734, opened 2026-07-27
feat(ci): add EvalForge actions CI gate; fix stale evalforge-yaml-gate dist

orgoldreich1:
## What & why
Found while investigating why #731's checks weren't showing up: there is
**no CI workflow at all** covering `packages/evalforge-core` or the
`.github/actions/evalforge-yaml-gate`/`evalforge-skill-gate` actions. Every
"typecheck clean" / "N/N tests passing" / "dist rebuilt" claim in recent PRs
touching these (e.g. #696, #731) has come from a manual local run by the PR
author — never enforced by CI.

## Changes
- New `.github/workflows/evalforge-actions-ci.yml`, triggered on PRs touching
  `packages/evalforge-core/**` or either `evalforge-*-gate` action. Builds
  `evalforge-core` first (both actions resolve `@wix/evalforge-core` to its
  `dist/index.d.ts`/`dist/index.js` via a `portal:` link, so it has to exist
  before the actions' own typecheck/build can succeed), then for each action:
  install, typecheck, test, build, and fail if the rebuilt `dist/index.js`
  differs from what's committed. Modeled on the existing `skill-eval-ci.yml`.
- The `evalforge-skill-gate` step is guarded on the directory existing
  (`hashFiles(...) != ''`) since this workflow lands before that action does.

## A live bug this surfaced
Running the new "dist must match a fresh build" check against unmodified
`main` failed immediately: `evalforge-yaml-gate/dist/index.js` was already
stale. It predates `evalforge-core`'s `skill_was_called`/`build_passed`/
`token_count` assertions, scenario-level `templateId`, and the `llm_judge`
`scoringMode`/`browserTools`/`parameters` fields — all landed in `evalforge-core`
source via #689, never bundled into the action that actually runs in CI. Any
scenario YAML using those would fail schema validation today, not because the
YAML is wrong but because the live gate is executing dead code. Rebuilt here
so the new check starts green and those fields actually work.

## Testing
- `evalforge-core`: 101/101 passing, typecheck clean.
- `evalforge-yaml-gate`: 98/98 passing, typecheck clean, `yarn build` output
  now matches the committed `dist/index.js` exactly (verified by re-running
  the build against the committed state after this change).

### DIFF
--- a/.github/actions/evalforge-yaml-gate/dist/index.js
+++ b/.github/actions/evalforge-yaml-gate/dist/index.js
@@ -34177,11 +34177,15 @@ const SYSTEM_LLM_JUDGE = 'system:llm_judge';
 const SYSTEM_API_CALL = 'system:api_call';
 const SYSTEM_COST = 'system:cost';
 const SYSTEM_TIME_LIMIT = 'system:time_limit';
+const SYSTEM_SKILL_WAS_CALLED = 'system:skill_was_called';
+const SYSTEM_BUILD_PASSED = 'system:build_passed';
+const SYSTEM_TOKEN_COUNT = 'system:token_count';
 function toEvalForgeBody(s) {
     return {
         name: s.name,
         description: s.description,
         triggerPrompt: s.triggerPrompt,
+        ...(s.templateId ? { templateId: s.templateId } : {}),
         assertionLinks: s.assertions.map(mapAssertion),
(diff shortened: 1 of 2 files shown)


orgoldreich1:
Duplicate of #714, which already adds this exact CI gate (and more thoroughly — tested against #696's branch, better staleness error message). Closing in favor of that one.

PR #751, opened 2026-07-28
feat: wix-app PR eval gate (CODEAI-889)

orgoldreich1:
Phase 1 of the codegen auto-eval initiative ([CODEAI-889](https://wix.atlassian.net/browse/CODEAI-889)). When a PR changes the `wix-app` skill, run the eval scenarios covering what it changed against **that PR's** version of the skill, comment the results, and eventually block the merge.

> **Stacked on #753.** Base is `codeai-889-gate-core`, so this diff shows only the wiring, workflows and docs. **Merge #753 first.**

## max-scenarios: 25 — the reviewed decision

Each wix-app scenario is a real agent build, so the expensive case is a `SKILL.md` edit running the union of all tags. **25** bounds worst-case cost and wall-clock while comfortably covering today's suite. Override per-repo with `vars.WIX_APP_EVAL_MAX_SCENARIOS`. Anything cut is **named in the PR comment** — never silently dropped.

## ⚠️ Do not flip `blocking` to `true` yet

`yaml/wix-app-evals/` holds **one** scenario covering three tags, against roughly **twenty** tag-deriving reference files. So most reference PRs will hit the coverage check until more scenarios exist. That is survivable only because the gate ships with `blocking: false` (`vars.WIX_APP_EVAL_BLOCK_MERGE`).

**Treat the soak period's comments as the scenario-authoring work-list.** Flip blocking on only once the suite covers the references people actually edit.

## Flow

```
author gate (@wix.com first commit)  → not Wix? exit 0, no cost
loadScenarios(head)                  → YAML errors? comment + fail
changed files → deriveTags + touchedScenarioPaths
  nothing gated                      → comment no-op, exit 0
guardScenarios                       → violations? comment + fail, SKIP the run
                                     → warnings carried into the final comment
collectSkillFiles → createOrReuseSkillVersion('pr-<n>-<mergeSha7>')
loadScenarios(base) + remote lookup → diffSyncPlan(semanticPlusDraftTags)
  → apply CREATE / UPDATE / DELETE / DEFER_DELETE
selectScenarios(tag-matched ∪ touched, capped)
createAndRunEvalRun → pollUntilDone → evaluateRunResult → comment + pass/fail
```

The author gate and the coverage guard both run **before any EvalForge write**, so a fork PR or a missing scenario costs no run.

## Two correctness fixes found while building this

**1. The version label named content it didn't identify.** The workflow's first checkout has no `ref:`, so on `pull_request` it checks out the *merge* commit — head merged into base — while the label was built from `head.sha`. The same head yields different merge content as base advances, so `createOrReuseSkillVersion` could find the existing label and reuse a version built from stale content. Now labelled from `GITHUB_SHA`, which names exactly the checked-out merge commit. The `pr-<n>-` prefix is unchanged, so PR-close cleanup still sweeps.

**2. `ready_for_review` was missing from the workflow triggers.** The job is gated on `!draft`, so without that event a PR marked ready would never fire the gate — it would sit unevaluated until its next push. Caught by writing the workflow test.

## Mutation-checked guards

Passing tests don't prove a test would catch the bug, so I reverted each guard and confirmed which test fails:

| Guard | Mutation | Result |
|---|---|---|
| Zero resolvable scenario ids fails | exit quietly instead | 1 test fails |
| `diffSyncPlan` gets `semanticPlusDraftTags` | drop it → semantic tags stripped remotely | 1 test fails |

The first is a genuine false-pass guard: without it a gate that resolved nothing to run would
report green.

**The second I initially described too strongly**, and the correction is worth stating. Stripping
the semantic tag does *not* make the current run cover nothing:
`remoteScenarioFiltersForGate` also queries by **name** for every scenario in `changedHead`, and
`applySyncPlan` writes CREATE ids into `nameToId` — so a scenario the PR touched always resolves
whatever its remote tags. The tag query only matters for *untouched* scenarios, whose tags are
never rewritten.

What it actually prevents: after an earlier run strips the tag, a later run of the same PR that no
longer has that scenario in its cumulative diff (the author reverted the edit) can't find it by tag,
so it drops out of tag-based selection — surfacing as `missingIds`, or as a zero-selection failure
if it was the only candidate. A wrongly skipped scenario, reported, not a silent green. Still worth
keeping, and it still removes the need for a promote step.

## Docs

`CONTRIBUTING.md` had reached 440 lines serving three readers, so it is split here — it already had to change for the gate, so the two aren't separable:

| File | Lines | Reader |
|---|---:|---|
| `CONTRIBUTING.md` | 440 → **85** | first PR — placement, adding skills, checklist |
| `docs/eval-scenarios.md` | 237 | authoring a scenario |
| `docs/skill-evaluation.md` | 140 | debugging a failing check, or the workflows |

Prose is unchanged — only links rewritten and one heading renamed. Verified by diffing heading sets and every non-blank line against the original; every markdown link resolved programmatically, cross-file anchors included.

## Verified end to end against the live EvalForge project

Not just unit-tested — the committed bundle was run against the real App Builder project on
throwaway PRs. **This found two bugs that no local test could reach**, both fixed here.

### Bug 1 — `capabilityVersions` took the version label, not its id

The run died in 2.3s, server-side, before any agent work:

```
[Phase: main-execution] [fetch-eval-run] Failed to fetch evaluation data: status code 400
```

The evaluator resolves `capabilityVersions` by version **id**; we passed the label. Of 245 runs
in that project, all 80 completed ones pin by UUID — exactly one ever used a label, ours, and it
failed. `createOrReuseSkillVersion` already returned the id; the flow checked it was non-null and
then discarded it.

Unreachable locally: `capabilityVersions` is a hand-written mirror of the API and nothing in the
repo had ever exercised it, so label-vs-id was an assumption read off `Record<string, string>`.

### Bug 2 — every comment would have said `Pass rate: 10000%`

`toPercent()` multiplied by 100, documented as "V1 `pass_rate` is a fraction (0.0–1.0)". The live
API sends a percentage: 4/5 → `80`, 26/30 → `86.667`, checked across 8 real runs. A pre-existing
test locked in the wrong belief — fed `0.75`, asserted `75`, and passed against the bug. Fixed in
the core PR (#753); it also corrects wix-manage's scheduled-run summary, which reported the same
inflated figure.

### What the live runs confirmed

| Behaviour | Evidence |
|---|---|
| OAuth against the real gateway | token minted, all calls authorised |
| Version carries **the PR's** content, not main's | stored `references/DASHBOARD_PAGE.md` contains a marker that exists only on the PR branch; 85 files, `scripts/` included |
| Scenario set is **the PR's** | head checkout had 2 scenarios, base had 1 |
| Guard blocks with **zero** EvalForge calls | client never constructed; comment posted; `exit=0` in soak mode |
| Both guard violation kinds render together | uncovered tag + a below-bar new scenario in one comment |
| Full happy path | version → run → **7/7 assertions, 100%** in 275s → comment |
| Comment upsert | two `comment()` calls produced exactly one comment, updated in place |
| Draft-sync `CREATE` | scenario stored as `tags=["dashboard-page","draft:wix/skills#765","created-via-code","repo:wix/skills"]`; every one of those tag queries finds it |
| The created scenario is what ran | run's `scenarioIds` = the id `CREATE` returned |
| Failure path | 4 assertions, 3 passed, 1 failed → `passRate: 75` → ⚠️ comment, "**Why this would block:** 1 assertion failed", soak note, `exit=0` |
| Cleanup, version only | `Deleted capability version pr-766-7507f7d`, `0 action(s)` — correct, #766 changed no scenario YAML |
| Cleanup, version + scenario | `Deleted capability version pr-765-e452da7`, `RESTORE=0 DELETE=1`, `Deleted draft e2e-good-scenario` |
| Project left at baseline | 0 capability versions, 1 scenario with its original tags, 0 draft-tagged |

### Also fixed while verifying

The author gate was the one unguarded network call in `runGate`, so a GitHub error escaped to
`index.ts`'s catch and became `setFailed` — a red check regardless of `blocking`, breaking the one
promise soak mode makes. Reproduced with an invalid token (exit 1), now skips instead (exit 0) and
reports the real cause rather than blaming a non-Wix author.

### Scope of the verification

Three throwaway PRs covered the paths that differ at the EvalForge boundary: guard-blocks
(no writes), the happy path (version → run → pass), and a new scenario (`CREATE` → run → fail).
Both cleanup branches ran. Two real agent builds, ~$2, all artefacts removed afterwards.

Not covered, and only the soak period can: the `UPDATE` and `DEFER_DELETE` sync branches, a
`FOREIGN_DRAFT` collision between two open PRs, a genuine 30-minute poll timeout, and the
`max-scenarios` cap biting — the suite has one scenario, so nothing reaches 25.

## Notes for reviewers

- `action-yml.test.ts` asserts the declared `ignore-globs` / `broad-impact-globs` defaults match the core constants. Two copies of one list; silent drift would change which files trigger a run.
- The workflow tests parse YAML structurally rather than string-matching, so they assert semantics — trigger types, concurrency, base checkout, inline ids, sha-pinning.
- Cleanup runs on **merge as well as close**: wix-app has no promote step, so nothing else sweeps a merged PR's draft tags and versions.
- `capability-id` / `agent-id` are inline in the workflow, not repo vars — neither is a secret, and inline they appear in the diff and get reviewed when they change.
- **No new repo configuration is required.**

Tests: skill-gate 8 → 70. yaml-gate 76 unchanged. No UI changes, so no screenshots.


.github/actions/evalforge-skill-gate/dist/index.js
```diff
@@ -61538,11 +61538,14 @@ var __importStar = (this && this.__importStar) || (function () {
 Object.defineProperty(exports, "__esModule", ({ value: true }));
```
mirivoWix, line 61554: Can the dist be in gitignore?

.github/actions/evalforge-skill-gate/src/utils/cleanup-run.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (!remote) return;
+
+  const baseRoot = posix.join(workspaceRoot(), BASE_WORKSPACE_SUBDIR);
+  const { scenarios: baseScenarios, errors } = loadScenarios(baseRoot, config.evalsGlob);
+  for (const error of errors) {
+    core.warning(`Base SHA scenario issue at ${baseRoot}/${error.path}: ${error.message}`);
+  }
+
+  const plan = planCleanup(remote, baseScenarios, draftTag, config.repoFullName);
+  const restoreCount = plan.filter(action => action.kind === 'RESTORE').length;
```
mirivoWix: Can we use enum for the types?

.github/actions/evalforge-skill-gate/src/utils/pr-lookups.ts
```diff
@@ -0,0 +1,46 @@
...
+ * when the lookup fails: either way the gate must not run, and neither is worth failing a check.
+ */
+export async function skipReasonForAuthor(
+  octokit: ReturnType<typeof github.getOctokit>,
+  config: Pick<GateConfig, 'owner' | 'repo' | 'prNumber'>,
+): Promise<AuthorSkip | undefined> {
+  try {
+    const email = await getFirstCommitAuthorEmail(octokit, config.owner, config.repo, config.prNumber);
+    if (isWixAuthorEmail(email)) return undefined;
+    return { reason: 'the PR author is not a @wix.com address', unexpected: false };
```
mirivoWix:
```suggestion
    return { reason: 'the PR author is not a wix author', unexpected: false };
```

.github/actions/evalforge-skill-gate/src/utils/apply-sync-plan.ts
```diff
@@ -0,0 +1,41 @@
...
+export async function applySyncPlan(
+  client: EvalForgeClient,
+  config: GateConfig,
+  actions: SyncAction[],
+  nameToId: Map<string, string>,
+  comment: Commenter,
+): Promise<boolean> {
+  for (const action of actions) {
+    try {
+      if (action.kind === 'CREATE') {
```
mirivoWix: Please use enum or import type

.github/actions/evalforge-skill-gate/dist/index.js
orgoldreich1: No - the action runs `dist/index.js` directly, so it has to be committed

.github/actions/evalforge-skill-gate/src/utils/cleanup-run.ts
```diff
@@ -0,0 +1,68 @@
...
+  const deleteCount = plan.filter(action => action.kind === 'DELETE').length;
+  core.info(`Cleanup plan: ${plan.length} action(s) — RESTORE=${restoreCount} DELETE=${deleteCount}`);
+
+  for (const action of plan) await execute(client, config.projectId, action);
```
mirivoWix, line 52: can we use here promise all?

mirivoWix (commented):
Reviewed against `ASTRA.pr.md`.

**Verdict — fails two rules the checklist marks REJECT**, both narrow. In substance this is well above the bar the checklist was written for. Merge-blocking list: the `runGate` split, a `pr-lookups.test.ts` plus a real `pulls.get` mock, and dropping `id-token: write`. The README `ready_for_review` row is a one-line fix. Everything else is fair follow-up.

**What I could and couldn't check:** `gh pr checks 751` reports no checks on this branch, so this is a source read, not a review on top of a green build. I verified the flow claims against `evalforge-core` on the base branch — `select-scenarios.ts`, `guard-scenarios.ts`, `plan-pr-scenario-sync.ts`, `derive-tags.ts`, `auth.ts`. One hypothesis I chased and dropped: I expected a deletion-only scenario PR to reach the zero-selection failure, but `touchedScenarioPaths` filters `removed` files, so it exits cleanly at the no-gated-changes check. That path is correct.

---

## Hard checklist violations

### 1. `runGate` is 182 lines — `src/utils/gate-run.ts:20-201`

The limit is 30. The seams are already here and already match the README's flow graph, so the split is mostly moving code:

- **`resolveGateScope`** — `loadScenarios` → `getChangedFiles` → `deriveTags` → `touchedScenarioPaths` → `guardScenarios`. This is exactly the "everything before the first EvalForge write" boundary you already document, so it's a natural unit.
- **`syncDraftScenarios`** — base load → `remoteScenarioFiltersForGate` → `listRemoteScenariosForGate` → `stripInactiveForeignDraftTags` → `diffSyncPlan` → `applySyncPlan`, returning `nameToId`.
- **`runAndReport`** — `selectScenarios` → `startEvalRun` → `pollToCompletion` → `formatGateResult`.

Each returns its result or a sentinel meaning "already commented and already called `fail`" — the convention the body uses throughout. `runGate` then reduces to the sequence of guarded steps the graph shows.

### 2. Four new modules have no test file

`src/utils/apply-sync-plan.ts`, `pr-lookups.ts`, `report.ts`, `run-eval.ts`. Three are genuinely exercised through `gate-run.test.ts`, so that's mostly bookkeeping. `pr-lookups.ts` is the one where the indirection hides a real gap:

**`isDraftTagActive` — `src/utils/pr-lookups.ts:31-46` — the branch that releases an abandoned PR's lock has no coverage.**

It's only reached through `stripInactiveForeignDraftTags` in `gate-run.test.ts`. There `octokit` is mocked as `{ rest: { pulls: {}, issues: {} } }` (`tests/gate-run.test.ts:18`), so `octokit.rest.pulls.get` is `undefined`, the call throws a `TypeError`, and control lands in the `catch` on line 42 → `return true`. So:

- line 41 (`pull.data.state === 'open'` → `false`, the path that frees the lock) is **never executed** by any test;
- replacing the whole function with `() => true` leaves the suite green.

That's the mutation worth guarding, because always-`true` is a silent deadlock: every PR touching a scenario an abandoned PR once held stays blocked by `FOREIGN_DRAFT` forever, and nothing reports why. Given the mutation-check table in your description, this one earns a place in it.

Suggest a `pr-lookups.test.ts` covering the outcomes directly — open → `true`, closed → `false`, `pulls.get` rejects → `true`, plus the two early `return true`s on 36/38. And in `gate-run.test.ts:18`, a real `pulls.get: vi.fn().mockResolvedValue({ data: { state: 'open' } })` so the `FOREIGN_DRAFT` test at line 246 tests what it reads as, rather than passing via the error path.

---

## Code structure

**`guardedCall`'s failure signal collides with a falsy success — `src/utils/report.ts:16-30`**

It returns `T | undefined`, and every caller tests falsiness — `if (!version) return;`, `if (!changedFiles) return;`, `if (!remote) return;`. All five wrapped operations return objects or arrays today, so nothing is broken.

But the failure mode if that stops holding is the specific one this PR works hardest to prevent. A wrapped call that legitimately returns `0`, `''`, `false` or `null` reads as failure, and `runGate` returns **without** commenting and **without** calling `fail` — a green check that verified nothing. Same class as the zero-selection and zero-assertion guards you added deliberately, except it arrives silently. Cheap to close structurally:

```ts
type Guarded<T> = { ok: true; value: T } | { ok: false };
```

Callers become `if (!result.ok) return;` and the falsy-value question stops existing.

**Three dead fields on `CleanupConfig` — `src/utils/config.ts:61-73`**

`runCleanup` never constructs an octokit, so `githubToken` (62) has no consumer. `owner` (69) and `repo` (70) are only used inside `getCleanupConfig` to build `repoFullName`, never read off the config object. It propagates: `evalforge-wix-app-gate-cleanup.yml` grants `pull-requests: read` and passes `github-token` for a token nothing uses. If `safeGetSecret` is being called for its log-masking side effect that's a good reason — worth one line saying so, since otherwise the next reader deletes it.

**Unreachable branch — `src/utils/config.ts:76-81`**

Line 78 isn't a distinct case: if `raw.trim() === ''` then `entries` is empty too, so line 80's fallback already handles it.

```ts
function getMultilineList(name: string, fallback: string[]): string[] {
  const entries = core.getInput(name).split('\n').map(line => line.trim()).filter(line => line !== '');
  return entries.length > 0 ? entries : fallback;
}
```

**`max-scenarios` has no upper bound — `src/utils/config.ts:83-90`**

`max-scenarios: 100000` validates fine. Worth a ceiling because this input exists purely to bound cost, and your description makes the argument itself: every wix-app scenario is a real agent build, so a fat-fingered repo var is real money and hours of wall-clock rather than a slow test run. A cap, or at least a `core.warning` above ~100, matches what the input is for.

**`src/utils/` is a catch-all**

The checklist forbids `utils.ts` filenames for the "split into meaningful modules" reason, and a `utils/` directory holding `gate-run.ts`, `config.ts` and `report.ts` is the same smell. Pre-existing, so not this PR's to fix — but it adds five more files to it. Flattening to `src/` is a clean follow-up.

---

## Naming

- **`src/utils/config.ts:49`** — `blocking` needs a positive boolean prefix: `isBlocking`. The action *input* should stay `blocking`; that name is the public contract and reads correctly in YAML. Only the internal field changes, so `action.yml` and the workflows are untouched.
- **`src/utils/pr-lookups.ts:11`** — `unexpected` → `isUnexpected`.

---

## TypeScript strictness

- **`src/utils/config.ts:94`** — `(pullRequest?.head as { sha?: string } | undefined)?.sha`, an assertion without the justification comment the checklist asks for. `@actions/github` ships payload types (`PullRequestEvent['pull_request']['head']['sha']`), so this can be typed rather than asserted — and then line 95's `if (!headSha)` guards a genuinely optional value instead of one the cast invented.
- **`tests/gate-run.test.ts:104`** — `as never` here and on five `mockResolvedValue(… as never)` calls: the most complete type hole available, sitting on the mock boundary, which is the one place you want types intact. It means a change to `GateConfig` or `EvalRunStatus` won't surface as a test failure. `configOverrides: Partial<GateConfig>` plus a `CONFIG: GateConfig` annotation removes it on this line; typing `runMetrics` as returning `EvalRunStatus` removes the others.
- **`src/utils/report.ts:5`** — `describeError(error: unknown)`. The checklist bans `unknown` outright, but a `catch` binding can only be `unknown` or `any`, so this is the right call and I wouldn't change it. Flagging because the rule as written is absolute — this is the kind of exception worth recording in `ASTRA.pr.md` rather than re-litigating per PR.

---

## Tests

- **`tests/action-yml.test.ts:10`** — module-level `let action`, which the checklist rejects, and `beforeAll` isn't buying anything since `readFileSync`/`yaml.load` are synchronous. `const action = yaml.load(readFileSync(join(__dirname, '../action.yml'), 'utf8')) as ActionYml;` drops the `let`, the `beforeAll`, and the import.
- **`tests/workflow-config.test.ts:20-22, 93-95`** — same declare-then-assign-in-`beforeAll`, twice. Also `raw` is only read on line 26 inside the `beforeAll` that assigns it, so it doesn't need describe scope at all. Flagging because commit `73a5627b` is literally *"drop declare-then-assign-in-try"*: the pattern got cleaned out of `src` in this same PR and survived into the new tests.

---

## Configuration

**`id-token: write` isn't needed — `.github/workflows/evalforge-wix-app-gate.yml:32`**

Auth is OAuth client-credentials end to end — `TokenProvider` in `evalforge-core/src/auth.ts` posts `client_id`/`client_secret` to `/oauth2/token`. Nothing in the action or in core requests an OIDC token. `evalforge-wix-app-sync.yml` gets this right (`contents: read` + `pull-requests: read`). This looks copied from `evalforge-yaml-gate.yml:26`, which carries the same surplus — inherited rather than introduced, but a new file is the cheap moment to drop it.

**`25` now lives in three places — `src/utils/config.ts:135`**

Here, `action.yml`'s `default: '25'`, and `evalforge-wix-app-gate.yml:58` (`vars.WIX_APP_EVAL_MAX_SCENARIOS || '25'`). `tests/action-yml.test.ts:39` pins only `action.yml`'s copy. `config.ts:132` has the same shape for `'references'`.

Your own reviewer note makes the argument — *"Two copies of one list; silent drift would change which files trigger a run."* And `ignore-globs`/`broad-impact-globs` already solve it the right way: exported from core, asserted equal in `action-yml.test.ts`. Same treatment for `DEFAULT_MAX_SCENARIOS` and `DEFAULT_REFERENCE_DIR`, plus two more assertions. Given `25` is called out as *the reviewed decision*, it's the number that most deserves a single home.

---

## Documentation

**`README.md:10`** — the `gate` row omits `ready_for_review`, which the workflow does trigger on (`evalforge-wix-app-gate.yml:13`) and which `tests/workflow-config.test.ts:42` asserts. Worth fixing precisely because it's correctness fix #2 in your description: the job is gated on `!draft`, and after this change the README is the only place that still implies otherwise.

**Commit messages** — `ci: wix-app PR eval gate and PR-close cleanup workflows` and `docs: split CONTRIBUTING into entry point + two references` omit the `(<package>)` scope our template requires, and `ci`/`style` (used twice) aren't in its type list (`feat`/`fix`/`docs`/`refactor`/`perf`/`test`/`chore`). Both are standard conventional-commit types, so I'd widen `ASTRA.pr.md` rather than rewrite the commits — raising it so the checklist and practice stop diverging.

---

## Worth stating, since it's most of the diff

No `any`, no `@ts-ignore`, no TODOs, no commented-out code, explicit return types on every exported function, and comments that consistently explain why — the `getEvaluatedSha` merge-vs-head note and the `broad-impact-globs` precedence rationale both earn their space. Tests parse workflow YAML structurally instead of string-matching. `dist` staleness is enforced by a required check. README, `AGENTS.md` and the `CONTRIBUTING` split all ship together. The two bugs only live runs could surface — version id vs. label, and `toPercent` — are both locked in by tests.

One process point: the description is explicit that `blocking` must stay `false` until the suite catches up, and the reasoning holds — one scenario against ~20 tag-deriving reference files means most reference PRs will hit the coverage guard. That deserves a tracked ticket rather than a paragraph in a merged PR description, since the gate's whole value is in eventually flipping it.

.github/actions/evalforge-skill-gate/src/utils/cleanup-run.ts
orgoldreich1: done

.github/actions/evalforge-skill-gate/src/utils/apply-sync-plan.ts
orgoldreich1: done

.github/actions/evalforge-skill-gate/src/utils/pr-lookups.ts
orgoldreich1: done

.github/actions/evalforge-skill-gate/src/utils/cleanup-run.ts
orgoldreich1: Cleanup runs after the PR is closed, so nothing is waiting on it — parallelism would buy latency we don't need, and sequential keeps the log lines in plan order. `execute` already catches per action, so there's no resilience gain either.

orgoldreich1:
Thanks — all three merge-blockers are in, plus most of the rest. Two came out differently than expected.

**Fixed** — #751 `4ae940e`, `a185892`; #753 `51d8fff`

- `runGate` split into `gate-scope.ts` / `sync-draft-scenarios.ts` / `run-and-report.ts`. 182 lines → 35, pure code motion, tests untouched.
- `pr-lookups.test.ts`, 8 tests, and `gate-run.test.ts` now mocks a real `pulls.get`. Mutation-verified both ways: `return true` fails the new test, `return false` fails the `FOREIGN_DRAFT` test. Before this, `return true` left all 75 green — good catch.
- `id-token: write` dropped, absence pinned by a test.
- `guardedCall` → `Guarded<T>`, and I swept the rest of the `| undefined` returns with it. Worst was `checkPrAuthor` (was `skipReasonForAuthor`): `undefined` meant *proceed*, so a missing return would have opened the gate rather than closed it.
- `CleanupConfig`: `githubToken`/`owner`/`repo` gone. Dropping them showed cleanup makes no GitHub API call at all, so its `pull-requests: read` and `github-token` are gone too.
- `getMultilineList` dead branch, `isBlocking`/`isUnexpected`, `as never`, `25` and `'references'` in core, README `ready_for_review`, and the three `beforeAll`s.
- Soak flip tracked as CODEAI-948.

**`max-scenarios` clamps rather than throws.** `getGateConfig` runs before `isBlocking` is known, so a throw fails the check even during soak — the bug we already hit with the author gate. Ceiling 100 with a warning, and `selectScenarios` already names what a cap dropped, so it surfaces in the comment instead of only the log.

**`config.ts:94` — kept the assertion, added the justification.** `@actions/github` declares the payload as `[key: string]: any`, so `head` arrives as `any` and the assertion narrows rather than widens. `PullRequestEvent` lives in `@octokit/webhooks-types`, which neither action depends on — adding it moves all three lockfiles through the `portal:` link to type one property.

**`as never` caught us mid-review.** Renaming `blocking` → `isBlocking` broke 8 tests at runtime while typecheck stayed green. Fixed as you suggested: `CONFIG: GateConfig`, `Partial<GateConfig>`, `runMetrics` returning `EvalRunStatus`.

**Not doing**

- `src/utils/` flatten — agreed it's a smell, keeping it a follow-up so the split stays a readable code-motion diff.
- `ASTRA.pr.md` widening — separate PR, it's a repo-wide policy file.
- `let` in tests — we allow it. I dropped the `beforeAll`s anyway, since `readFileSync` is synchronous and that removes the pattern as a side effect.
- `Promise.all` in cleanup — answered on the thread.

Suites: core 257, skill-gate 87 (was 75), yaml-gate 76.

mirivoWix (dismissed):

PR #753, opened 2026-07-28
feat(evalforge-core): the wix-app gate's decision logic

orgoldreich1:
Core modules for the wix-app PR eval gate ([CODEAI-889](https://wix.atlassian.net/browse/CODEAI-889)). **No action, no workflow, nothing wired yet** — this is the logic, with its tests. The wiring follows in #751.

> **Stacked on #750.** Merge that first. Split out of #751 to keep the review surface small.

## What's here

| Module | Responsibility |
|---|---|
| `evalforge.ts` | `createSkillVersion` / `ensureSkillVersion` — same endpoint as the MCP pair, `skillContent` oneof |
| `collect-skill-files.ts` | skill dir → `{ path, content }[]`, skillDir-relative, size-capped |
| `derive-tags.ts` | changed paths + `TagRules` → `{ tags, broadImpact, unmapped }` |
| `guard-scenarios.ts` | the quality guard — coverage + no-weakening |
| `select-scenarios.ts` | tag set + name→id → capped run id set |
| `evaluate-run-result.ts` | `EvalRunStatus` → pass/fail verdict |
| `format-gate-comment.ts` | markdown for every outcome |

All pure or IO-injected; no `@actions/*` import. `evalforge-core` goes 163 → 252 tests.

## The three judgement calls worth your attention

**1. Tag precedence is first-match-wins: ignore → broad-impact → reference → unmapped.**
The order is load-bearing. Put the reference rule first and `references/CODE_QUALITY.md` derives a `code-quality` tag, whereupon the coverage guard demands a scenario carrying a tag no scenario will ever carry. Asserted, not assumed — see the `deriveTags — precedence` block.

**2. Coverage folds the quality bar in** (≥3 assertions, ≥1 `llm_judge`). A tag whose only scenario has one assertion is not covered in any useful sense: the gate would run it, pass, and have verified nothing. Each row of the four-row outcome table has its own test, because those rows *are* the contract.

**3. `broadImpact` is a flag that contributes no tags** — a deviation from the spec. The spec said both that broad-impact paths are "never coverage-checked" *and* that they expand to "the union of all tags". Together at the guard, a `SKILL.md` typo fix would block on any weak scenario anywhere in the repo — the inherited-debt failure the warning tier exists to prevent.

## Mutation-checked, because passing tests don't prove a test would catch the bug

| Guard | Mutation | Result |
|---|---|---|
| Tag precedence | reference rule before broad-impact | both precedence tests fail |
| Coverage requires the bar | "any scenario carrying the tag counts" | outcome-table row 2 fails |

## Notes

- **`minimatch` becomes an explicit core dependency** (glob 11 already locked 10.2.5 transitively). That propagates through the `portal:` link, so **all three** `yarn.lock` files are updated — CI runs `--immutable` and would otherwise fail in a way that looks unrelated.
- `evaluate-run-result` fails a **zero-assertion** run explicitly: no assertion failed, but none ran either, and that must not read as green.
- Both `dist` bundles are rebuilt because core is inlined into each. #750 marks them `linguist-generated`, so they should render collapsed.

No UI changes, so no screenshots.

### DIFF
--- a/packages/evalforge-core/src/evalforge.ts
+++ b/packages/evalforge-core/src/evalforge.ts
@@ -279,10 +279,6 @@ export class EvalForgeClient {
     }
   }
 
-  /**
-   * Same endpoint as createMcpVersion, with the content oneof set to `skillContent`. Lets a
-   * run evaluate a PR's skill files directly, with no MCP URL indirection.
-   */
   async createSkillVersion(
     capabilityId: string,
     projectId: string,
--- a/packages/evalforge-core/src/evalforge.ts
+++ b/packages/evalforge-core/src/evalforge.ts
@@ -279,7 +279,7 @@ export class EvalForgeClient {
     }
   }
 
-  async createSkillVersion(
+  private async createSkillVersion(
     capabilityId: string,
     projectId: string,
     versionLabel: string,
@@ -303,7 +303,7 @@ export class EvalForgeClient {
     return { id: created.id, capabilityId: created.capabilityId, version: created.version };
   }
 
-  async ensureSkillVersion(
+  async createOrReuseSkillVersion(
     capabilityId: string,
     projectId: string,
     versionLabel: string,
--- a/packages/evalforge-core/src/evalforge.ts
+++ b/packages/evalforge-core/src/evalforge.ts
@@ -19,6 +19,9 @@ export type RunStatus = 'pending' | 'running' | typeof TERMINAL_RUN_STATUSES[num
 
 export type CapabilityVersion = { id: string; capabilityId: string; version: string };
 
+/** One file of a skill-content capability version. `path` is relative to the skill root. */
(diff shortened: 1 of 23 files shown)


packages/evalforge-core/src/evalforge.ts
```diff
@@ -276,6 +279,49 @@ export class EvalForgeClient {
...
+          notes: `Auto-created for PR #${prNumber}`,
+          skillContent: { files },
+        },
+      },
+    );
+    const created = res.capabilityVersion;
+    return { id: created.id, capabilityId: created.capabilityId, version: created.version };
+  }
+
+  async createOrReuseSkillVersion(
```
orgoldreich1, line 307: We will reuse a skill version in case of a workflow re-run

mirivoWix (approved):

PR #759, opened 2026-07-29
🦞 igor-claw: document PatchDataCollection permissions bug + PUT workaround

ayal:
## What's wrong

`skills/wix-manage/references/cms/cms-schema-management.md`'s "Update Collection Settings" section only showed patching `displayName` via `PATCH /wix-data/v2/collections/{id}` and gave no guidance on `permissions` — which silently doesn't work there at all.

## Confirmed (reproduced against a live throwaway test site this run)

`PatchDataCollection` (`PATCH /wix-data/v2/collections/{id}`) rejects **any** `dataCollection.permissions.*` value with `WDE0075: Not recognized role provided in permissions.` — including a value identical to the collection's current one (a true no-op), and including well-formed, documented enum values (`SITE_MEMBER`, etc.). The `dataPermissions.{itemRead,...}` field that appears in every response (undocumented, an internal mirror of `permissions`) fails the same way if patched instead. `PatchDataCollection` also takes no `mask` parameter, so there's no mask-path workaround either.

The full-replace `PUT /wix-data/v2/collections` (`UpdateDataCollection`) **does** work for changing permissions (needs current `revision` + full `fields` array from a prior `GET`).

A real developer hit exactly this in production, burned many failed attempts across both variants, and ended up deleting and recreating collections just to change a read permission.

## Fix

Documents the bug and the working `PUT`-based workaround directly in the recipe, with a concrete example, so agents don't repeat the same failed exploration.

## Context

Slack thread: https://wix.slack.com/archives/C0BDXM5LLE7/p1785289496869599

This PR was opened automatically by an AI agent (Claude Code) acting on behalf of Ayal (ayalg@wix.com) — not written by Ayal directly.


skills/wix-manage/references/cms/cms-schema-management.md
```diff
@@ -112,6 +112,31 @@ curl -X GET \
 }
 ```
 
+This works for `displayName`/`displayField`. **It does not work for `permissions`** — confirmed (2026-07): any `dataCollection.permissions.*` value is rejected with `WDE0075: Not recognized role provided in permissions.`, even a value identical to the collection's current one (a true no-op). Don't try to work around it via the `dataPermissions.{itemRead,itemInsert,itemUpdate,itemRemove}` field that shows up in `GET`/`PATCH` responses — it's an internal mirror of `permissions`, not a separate public field, and patching it fails the same way. Don't burn attempts on `mask` variations either — `PatchDataCollection` takes no `mask` parameter.
```
giedriuswix: This deliberation should not be part of the skill.

giedriuswix (changes requested):

giedriuswix (approved):

PR #764, opened 2026-07-29
[e2e-verify] DO NOT MERGE — guard blocks, no EvalForge writes

orgoldreich1:
Throwaway PR for verifying the wix-app eval gate (#751) end to end. **Do not merge; will be closed and deleted.**

An uncovered tag (`backend-api`) plus a deliberately below-bar new scenario. Exercises both guard violation types in one comment, and verifies the gate makes **zero** EvalForge calls on this path.

### DIFF
--- a/skills/wix-app/references/BACKEND_API.md
+++ b/skills/wix-app/references/BACKEND_API.md
@@ -231,3 +231,5 @@ src/pages/api/
 - Include `Content-Type: application/json` header on JSON responses.
 - Include `statusText` in error responses.
 - Validate input parameters and request bodies.
+
+<!-- e2e-verify: throwaway change, do not merge -->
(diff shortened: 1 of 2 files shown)


orgoldreich1:
## ⚠️ EvalForge Skill Gate: Coverage Warning

The quality bar is **at least 3 assertions including one `llm_judge`** — a scenario below it would run, pass, and verify nothing.

- **`backend-api`** has no eval scenario at all. Add one under the scenario directory tagged `backend-api`, or add that tag to a scenario that already exercises the area.
- `yaml/wix-app-evals/_e2e-weak.yml` (`e2e-weak-scenario`) — has 1 assertion (needs at least 3); has no llm_judge assertion. This PR added or edited it, so it must meet the bar.

_No eval run was started — a coverage failure is caught before any run cost._

_The gate is in its soak period (`blocking: false`), so this check still passes._

orgoldreich1:
Throwaway PR for end-to-end verification of the wix-app eval gate (#751). Verification complete — closing. All EvalForge artefacts created during the run (capability versions and the draft-synced scenario) have been removed; the project is back to its baseline of 1 scenario and 0 versions.

PR #765, opened 2026-07-29
[e2e-verify] DO NOT MERGE — new scenario, draft-sync CREATE

orgoldreich1:
Throwaway PR for verifying the wix-app eval gate (#751) end to end. **Do not merge; will be closed and deleted.**

Adds an above-bar scenario tagged `dashboard-page`. Exercises `diffSyncPlan` → `CREATE` with **semantic + draft tags**, which is the false-pass fix: if the semantic tag were stripped, the subsequent tag query would find nothing and the gate would report green having run zero scenarios. Then runs it.

### DIFF
--- a/yaml/wix-app-evals/_e2e-good.yml
+++ b/yaml/wix-app-evals/_e2e-good.yml
@@ -1,0 +1,22 @@
+name: e2e-good-scenario
+description: Throwaway scenario that meets the quality bar, to exercise draft-sync CREATE and tag selection.
+triggerPrompt: |-
+  Build a dashboard page with a table listing products, each row showing name and price.
+  APP_NAMESPACE: "@mirivo/my-app31ceim", Use this namespace when creating data collections (for idSuffix scoping).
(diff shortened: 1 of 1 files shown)


orgoldreich1:
## ⚠️ EvalForge Skill Gate: Warning

**Pass rate:** 75% — 3/4 assertions passed, 1 failed
**Run:** [97440dd0-136d-41c4-a324-ac66a90f0dd7](https://bo.wix.com/pages/evalforge/bb388ce3-66c2-4868-a895-36b637635f59/results?runId=97440dd0-136d-41c4-a324-ac66a90f0dd7)

**Why this would block:** 1 assertion failed

**Scope:** 1 scenario(s) ran.

- `e2e-good-scenario`

_The gate is in its soak period (`blocking: false`), so this check still passes._

orgoldreich1:
Throwaway PR for end-to-end verification of the wix-app eval gate (#751). Verification complete — closing. All EvalForge artefacts created during the run (capability versions and the draft-synced scenario) have been removed; the project is back to its baseline of 1 scenario and 0 versions.

PR #767, opened 2026-07-29
[throwaway] e2e: gate should block on an uncovered tag

orgoldreich1:
Throwaway PR verifying the merged wix-app eval gate end to end in real CI.

Touches `references/BACKEND_API.md`, which derives the `backend-api` tag. The only scenario in `yaml/wix-app-evals/` carries `[dashboard-page, data-collection, auto-patterns-dashboard]`, so coverage should fail **before any EvalForge call** — comment posted, zero cost, and exit 0 because the gate is in soak (`blocking: false`).

Will be closed without merging.


orgoldreich1:
Throwaway — closing. Verified in real CI: workflow trigger, both checkouts, the committed bundle, the author gate, coverage caught before any EvalForge call, and the comment. Closing now also exercises the cleanup workflow.

PR #770, opened 2026-07-29
[throwaway] e2e A: invalid scenario YAML

orgoldreich1:
Throwaway. Empty `assertions` fails the schema, so the gate should comment **before** the guard and before any EvalForge call.

Checking #769's fix: this used to render ❌ with no soak note while the check went green. Expect ⚠️ plus the soak line.


orgoldreich1:

PR #771, opened 2026-07-29
[throwaway] e2e B: uncovered tag

orgoldreich1:
Throwaway. `backend-api` has no scenario, so coverage fails before any EvalForge call.

Checking #768's fix: the instruction should now name `yaml/wix-app-evals/` instead of "the scenario directory", and the quality-bar sentence should be **absent** (there is no scenario to be below the bar).


orgoldreich1:

PR #772, opened 2026-07-29
[throwaway] e2e C: ignored path, no gated changes

orgoldreich1:
Throwaway. Touches only `skills/wix-app/scripts/**`, which is in `ignore-globs` — a generator change arrives with its regenerated docs anyway.

The workflow should trigger (the path is under `skills/wix-app/**`) but derive no tags, so expect **No Gated Changes** with a green check and no run.


orgoldreich1:

PR #779, opened 2026-07-30
test(ecommerce): guard recommend-e-commerce-strategy against traffic-acquisition misroutes

itayhewix:
## What

Adds one eval scenario. **No skill content changes** — `SKILL.md` and `recommend-ecommerce-strategy.md` are untouched.

## Why this shrank

The first version of this PR softened the orchestrator's routing cues ("Any sales/business improvement request", "the **default entry point**", "Use this for **ANY** business improvement request") and added a traffic/acquisition exclusion, on the strength of a production conversation where *"help me grow my site traffic"* misrouted into the recipe and died on `Cannot generate recommendations — missing required site data`.

**The gate disproved the premise.** The required negative scenario passes identically with and without the change:

| | PR | prod (main) |
|---|---|---|
| `NOT: ReadFullDocsArticle` | ✅ | ✅ |
| judge | 10/10 | 10/10 |
| behavior | routed to marketing/ads | routed to the marketing plan flow |

Current `main` already handles the traffic request correctly. The misroute no longer reproduces — whatever caused it has since been fixed elsewhere. With no behavioral gap to close, the skill edits were unjustified prose churn, and the pairwise judge mildly preferred prod on all three scenarios anyway (medium/low confidence, and the deltas are about recommendation substance the diff never touched — I read those as run-to-run variance, not regressions).

So the skill changes are dropped, along with the `recommend-strategy-conversion-boundary` control that only existed to guard them (~$1.06/run saved).

## What's left, and why it's still worth merging

The scenario **passes on main today** — it is a regression guard, not a fix. Its value:

- The boundary currently holds *implicitly*. Nothing in the skill says acquisition is out of scope; the correct routing is emergent. This pins it.
- The routing cues genuinely were broad enough to invite this ("**ANY** business improvement request"). If they're re-broadened, this catches it.
- The judge fails the `missing required site data` symptom **by name**, so the specific production failure mode would be caught rather than scored as a reasonable decline.
- $0.27/run — the cheapest of the three scenarios that ran.

If a reviewer would rather not carry a scenario that passes today, closing this is reasonable and I won't argue. The honest case for it is regression coverage on a boundary that holds by accident, not a bug fix.

## Verification

Validates against `ScenarioSchema` from `packages/evalforge-core` (checked against all 50 scenarios in `yaml/wix-manage-evals`; name unique across the tree). Behavioral result is the gate's own prod run above: ✅ / 10/10.


itayhewix:
Reduced to the eval only after reading the gate results above.

The required negative scenario passed on **prod** too (`NOT: ReadFullDocsArticle` ✅, judge 10/10, routed to the marketing plan flow) — so `main` already handles `grow my site traffic` correctly and the misroute this PR was written for no longer reproduces. With no behavioral gap, the routing-cue rewrite had no justification, so it's dropped along with the `conversion-boundary` control that only guarded it.

What remains is the one scenario, reframed honestly as a regression guard rather than a fix. Two things the gate did confirm, worth recording:

- The exclusion did **not** over-block — `conversion-boundary` ("plenty of visitors but hardly any buy") still loaded the orchestrator on the PR run.
- Keeping `help my business` in the trigger list was right — the existing `recommend-ecommerce-strategy` scenario still passed. #519 removed that phrase and never ran the scenario that would have caught it.

Title and description updated. Diff vs `main` is now a single added file.

PR #786, opened 2026-07-30
fix(wix-app): remove non-existent dashboardPage builder fields, add Stores V3 name search

mirivoWix:
Two reference bugs found while investigating codegen job `b952e4c2-5d5c-42aa-8011-85237062cdf2` (app `d9618e13-7a6b-40dd-902d-fcd71d9af4b1`, `ITERATE_CODEGEN`, APP_BUILDER). The run took **7.87 min / $1.67** against a 3 min / $0.50 SLO and produced 4 tsc errors across 3 validate rounds, ending in `NEEDS_INVESTIGATION`. Tracked as [CODEAI-956](https://wix.atlassian.net/browse/CODEAI-956).

## 1. `DASHBOARD_PAGE.md` documented five fields that don't exist

The "Optional builder fields" section listed `additionalRoutes`, `sidebar.disabled`, `sidebar.priority`, `sidebar.whenActive.selectedPageId` and `sidebar.whenActive.hideSidebar` as options on the `dashboardPage` builder.

The agent read that file, followed it, wrote `sidebar: { disabled: true }`, and failed:

```
cartgate-quantity.extension.ts(8,3): error TS2353: Object literal may only specify
known properties, and 'sidebar' does not exist in type 'Options$1'.
```

`@wix/astro@2.65.0` closes the type at six fields — `id`, `component`, `routePath`, `title`, `fullPage?`, `createdBy?`. Any agent following the old text fails to compile, deterministically.

Those capabilities do exist, but only on `trustedBackofficePage` (`@wix/astro/builders/trusted`, first-party apps), and there with **flat** names: `hideInSidebar`, `priority`, `activeSidebarPageId`. There is no nested `sidebar` object in *either* builder, so the documented shape was never real — this wasn't a case of the text pointing at the wrong builder.

**Recurrence over 14 days / 3,638 jobs:** 5 distinct jobs across 5 distinct apps hit `'sidebar' does not exist`; 9 distinct jobs hit some invalid property on `Options$1`. Systemic rather than model variance.

### Why the section is deleted rather than corrected

Everything it documented beyond the scaffolded fields was wrong, so there is nothing left to state. The agent gets its builder file from `wix generate` already correct, and the section's only remaining effect was to invite it to add fields.

A replacement "these fields don't exist" note was written and then dropped: it's a second claim that has to stay in sync with the SDK, and it would itself become wrong the day sidebar support ships on the public builder — the same failure mode being fixed here. The Scaffold section above still documents `route` → `routePath` and that the CLI generates the builder file, so nothing needed is lost.

## 2. Stores V3 had no name-search guidance

```
rules.web.ts(208,40): error TS2345: Argument of type '"name"' is not assignable to
parameter of type '"_id" | "slug" | "options.id" | "handle"'
```

V3 `queryProducts` filters only on those four fields; name search belongs on `productsV3.searchProducts()`. The SDK docs say this explicitly ("To look up products by name or brand, call Search Products instead"), but neither `STORES_VERSIONING.md` nor `stores/QUERY.md` mentioned `searchProducts` at all.

With no in-skill path to the right answer, the codegen-validator resolved the error by **deleting the product-search feature** — its own summary reads "Products are now listed without filtering by search term." Green build, silent functional regression, reported as a fix. The validator-prompt side of that is handled separately; this PR supplies the guidance that was missing.

V1 has no search method at all; its builder accepts `name` via `startsWith` (prefix match only, no fuzzy).

## Changes

**+32 / −10** across two files:

- `DASHBOARD_PAGE.md` — pure deletion of the incorrect section.
- `stores/QUERY.md` — short "Search products by name" section with one worked example covering both catalog versions, noting the two easy mistakes inline (`res.products`, not `res.items`; `fields` is required for free-text search).

`STORES_VERSIONING.md` is untouched — it already links `stores/QUERY.md` from its "Detailed references" list.

## Verification

Field lists were read from the published `.d.ts`/`.d.mts` of `@wix/astro@2.65.0` and `@wix/stores@1.0.849`, not from prose docs:

- `'name'` is present in `V3ProductSearchSpec.searchable`.
- V1's `startsWith` domain is `'_id' | 'name' | 'slug' | 'description' | 'sku'`.
- `contains` is absent from both builders (`TS2339`).

The typecheck gate was run the way `ci.yml` runs it (`npm install --legacy-peer-deps` + `node extract.mjs`) — **exits 0**.

One caveat worth knowing when reading that result: the harness concatenates all ` ```typescript ` blocks per file and filters `TS2300`/`TS2304` out of the failure list as extraction artifacts, so a green gate alone doesn't prove the examples type-check. The added block is therefore written to be genuinely checked — it carries no imports of its own (a repeated import across concatenated blocks is a duplicate-identifier error, not a convenience) and binds every identifier it uses. Pass-2 diagnostics for the touched files are byte-identical to the `main` baseline: 7 pre-existing `TS2304`s, none from any line added here.

Testing in codegen: `SKILLS_OVERRIDE_BRANCH=wix/skills:fix/dashboard-page-options-and-stores-v3-search`

## Out of scope

- **WDS docs example.** A third error in the same job came from the agent faithfully copying a published example: the Wix Design System Dropdown docs show `listItemSectionBuilder({ title, type, suffix })` with no `id`, but the type requires it (`{ id: string | number } & Omit<ListItemSectionProps, 'onClick'>`). That example does not type-check as published, and belongs to the WDS docs pipeline.
- **Pre-existing `TS2304`s** in `DASHBOARD_PAGE.md`'s embedded-script block and `QUERY.md`'s `getVersion` reference.
- **A latent gap in the gate.** Pass 1's semantic errors are currently masked because parse errors in 15 other extracted files abort semantic checking program-wide. Fixing those parse errors would cause pass 1 to start flagging the pre-existing `TS2304`s above and prepend `@ts-nocheck` to whole reference files — silently disabling type-checking for everything in them.


mirivoWix:
Follow-up commit `13772b7` addresses two review points:

**1. Missing import.** The V1/V3 split example used both `products` and `productsV3` with no import line — not copy-pasteable, which is exactly the defect class this PR exists to fix. Now has `import { products, productsV3 } from '@wix/stores'` plus a pointer to where `CatalogVersion`/`getVersion()` come from.

**2. Is there a search option for Stores V1?** Answered properly — the old text only mentioned it in passing. Verified against `@wix/stores@1.0.849`:

- V1 `products` exposes **no search export at all** — only `queryProducts`.
- V1's builder **does** accept `name`, the field V3's builder rejects: `startsWith` accepts `'_id' | 'name' | 'slug' | 'description' | 'sku'`.
- **Neither** version has `contains` (TS2339 on both builders).

The nuance worth documenting: the versions match in *intent*, not *behaviour*. V1 `startsWith` is a **prefix** match — "sho" finds "Shoes" but not "Running Shoes" — while V3 `search.expression` matches anywhere and tolerates typos via `fuzzy`. A single search box behaves differently per catalog version, so callers shouldn't assume parity. Added a comparison table covering search method, matching semantics, fuzzy support, and the searchable field lists for both versions.

Snippets re-checked with strict `tsc --noEmit`.

adimara (approved):

PR #798, opened 2026-07-30
fix(wix-manage): carry the product-removal endpoints in the Catalog V3 find-products recipe

danshw:
## What this fixes

A merchant asking to remove one product by name costs the agent two avoidable round-trips, because the recipe that owns the name lookup stops at the lookup.

Measured on EvalForge scenario `coverage/stores-archive-product` (run `f7ea29ef-a3ff-4e79-9b24-bd12989dcfc6`, Sonnet 4.6, 47 s wall-clock, friction judge 8/10 — outcome gate passed 10/7). Trigger prompt: *"Remove my Eval Gate Mug product from the store."* The agent loaded **Find Products (Query and Search, Catalog V3)**, searched by name, and then had to go looking for the delete endpoint on its own:

```
{"role": "assistant", "content": [{"type": "tool_use",
  "input": {"code": "async function() {\n  const results = lightIndex.filter(r => r.docsUrl && r.docsUrl.includes('stores') && r.methods);\n  const deleteMethods = [];\n ... }",
            "reason": "Find the delete product endpoint for Catalog V3."},
  "toolName": "mcp__wix-mcp-remote__SearchWixAPISpec"}]}
```

A second, separate friction in the same run: the name→id search returned the raw product objects and blew the tool response limit —

```
--- TRUNCATED ---
Response was ~17,504 tokens (limit: 6,000). Refine your code to return less data.
```

Both are gaps in the one recipe the agent already loads, so both are fixed there.

## What "archive"/"remove" actually maps to

Verified against the generated Catalog V3 reference through the Wix MCP docs tools (`SearchWixAPISpec` over `lightIndex`, `ReadFullDocsMethodSchema`), not from memory:

- **There is no archive operation for products.** Sweeping all 390 indexed resources for `archive`/`unarchive`/`restore`/`undelete`/`trash` returns nothing for Stores.
- There is no product-level visibility endpoint either — `UpdateCategoryVisibility` exists for *categories* only. Hiding a product is the `visible` field on the product entity.
- So "remove" resolves to exactly two distinct operations, and the recipe now routes on the merchant's wording:
  - **delete** → `DELETE https://www.wixapis.com/stores/v3/products/{productId}` (`wix.stores.catalog.v3.CatalogApi.DeleteProduct`, no request body) — this is what the run did and what the outcome judge scored 10.
  - **hide** → `visible: false`, cross-linked to the existing Update Product with Options recipe.
- `BulkDeleteProducts` → `POST /stores/v3/bulk/products/delete`, body `{"productIds": [...]}`, 1–100 ids, per-item `results[].itemMetadata.success` (a 200 is not per-item success).
- `BulkDeleteProductsByFilter` → `POST /stores/v3/bulk/products/delete-by-filter` does accept a fuzzy `search.expression`, so the friction judge's "there is no delete-by-query/name endpoint" is not quite right — but it is asynchronous (returns only a `jobId`) and deletes *every* match, so it is documented with an explicit warning against using it for a single named product.

The residual name→id round-trip is inherent: `name` is not in `QueryProducts`' filterable set, so an id is genuinely required. That call is not what this PR removes.

## Changes

- `skills/wix-manage/references/stores/find-products-query-and-search-catalog-v3.md`
  - new **After the lookup: acting on the product you found** section — the delete/hide decision table keyed on the merchant's wording, `DELETE /stores/v3/products/{productId}`, `POST /stores/v3/bulk/products/delete`, and the delete-by-filter caveat.
  - new **Return only the fields you need when you are resolving a name to an ID** subsection — projects the search response down to `id`/`name` so it cannot be truncated.
  - routing row in STEP 0 and an updated frontmatter `description` so a "remove product X" request lands on this recipe.
- `skills/wix-manage/SKILL.md` — index entry description kept in sync with the frontmatter.
- `yaml/wix-manage-evals/stores/remove-product-by-name-catalog-v3.yml` — new covering scenario.

## The scenario, and what it does and does not catch

`stores/remove-product-by-name-catalog-v3` seeds six products sharing an "Eval Gate" prefix, so the fuzzy name search returns several full product objects and both defects cost something. Assertions:

- `ReadFullDocsArticle` on `…/stores/skills/find-products-query-and-search-catalog-v3` — coverage for the changed reference.
- **`SearchWixAPISpec` with `negate: true`** — the discriminator. Production has to probe the spec for the delete endpoint; with the endpoint in the recipe it should not.
- outcome `llm_judge` (`minScore: 7`) — the named product deleted by id, the five decoys untouched, hiding does not count as removing.
- friction `llm_judge` (`minScore: 0`) — penalises truncated responses and endpoint-discovery calls specifically, so it can differ from the 8/10 the current content scores.

Not caught: the scenario cannot assert *why* a spec probe happened, so a probe made for a response shape rather than for the delete endpoint would fail the negated assertion just the same. It also does not verify the delete-by-filter or bulk-delete guidance — neither path is exercised by a single-product request.

## Out of scope

The tool response size limit that truncated the search result lives in the Wix MCP server, not here; this PR works within it rather than changing it.

### DIFF
--- a/skills/wix-manage/SKILL.md
+++ b/skills/wix-manage/SKILL.md
@@ -357,7 +357,7 @@ These recipes do NOT cover frontend development or SDK usage for displaying data
 **Technical:** Single product creation with options using Catalog V3 Products API. Covers option types (TEXT_CHOICES, SWATCH_CHOICES), choice configuration, and automatic variant generation.
 
 ### [Find Products (Query and Search, Catalog V3)](references/stores/find-products-query-and-search-catalog-v3.md)
-**Technical:** Find, search, query, and list products from a Wix Store using Catalog V3 Search Products and Query Products endpoints. Explains when to use each endpoint, correct fields enum values, filtering, sorting, and paging.
+**Technical:** Find, search, query, and list products from a Wix Store using Catalog V3 Search Products and Query Products endpoints. Explains when to use each endpoint, correct fields enum values, filtering, sorting, and paging. Also covers resolving a product name to its product ID for a follow-up write, and the removal endpoints to call next when the user asks to remove or delete a product by name.
 
 ### [Query Products (Catalog V1)](references/stores/query-products-catalog-v1.md)
 **Technical:** Query and list products from a Wix Store using the Catalog V1 Query Products endpoint. Use this recipe when the site's catalog version is CATALOG_V1. Covers basic queries, filtering, sorting, and paging.
--- a/skills/wix-manage/references/stores/find-products-query-and-search-catalog-v3.md
+++ b/skills/wix-manage/references/stores/find-products-query-and-search-catalog-v3.md
@@ -1,6 +1,6 @@
 ---
 name: "Find Products (Query and Search, Catalog V3)"
-description: Find, search, query, and list products from a Wix Store using Catalog V3 Search Products and Query Products endpoints. Explains when to use each endpoint, correct fields enum values, filtering, sorting, and paging.
+description: Find, search, query, and list products from a Wix Store using Catalog V3 Search Products and Query Products endpoints. Explains when to use each endpoint, correct fields enum values, filtering, sorting, and paging. Also covers resolving a product name to its product ID for a follow-up write, and the removal endpoints to call next when the user asks to remove or delete a product by name.
 ---
 
 # RECIPE: Business Recipe – Find Products in a Wix Store (Query and Search, Catalog V3)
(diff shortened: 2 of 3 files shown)


danshw:
Closing: the eval gate shows this change cannot be measured, and on the evidence the friction it targets is not a docs gap.

The gate ran `stores/remove-product-by-name-catalog-v3` on both sides. Full tool-call sequences:

**With this PR** (4 calls, 36 s, 238,611 tokens): `WixREADME` → `SearchWixAPISpec` → `SearchWixAPISpec` → `ExecuteWixAPI` (query + delete).
**Production** (3 calls, 23 s, 185,095 tokens): `WixREADME` → `ExecuteWixAPI` → `ExecuteWixAPI`.

Three things follow.

1. **`ReadFullDocsArticle` was never called on either side** — the assertion failed `with` *and* `without`. For a "remove product X" request the agent goes straight from `WixREADME` to execution and never opens a recipe, so the section this PR adds is not on the path it takes. Nothing written in that file could have changed either run.

2. **The spec-probe friction does not reproduce.** The finding that motivated this PR was a run where the agent spent a `SearchWixAPISpec` call whose stated reason was "Find the delete product endpoint for Catalog V3." The production side here made **zero** spec probes and completed in two calls — the negated `SearchWixAPISpec` assertion *passed* on production and *failed* on the PR side. That inverts the premise: the extra probe is run-to-run variance, not a missing endpoint in the docs.

3. **The remaining friction is inherent to the API, not to the documentation.** `QueryProducts` does not accept `name` as a filter (only `id`, `handle`, `options.id`, `slug`, `createdDate`, `updatedDate`, `visible`), and Catalog V3 has no archive operation and no product-level visibility endpoint — sweeping the whole indexed REST surface for `archive`/`unarchive`/`restore`/`undelete` returns nothing for Stores. So resolving a named product to its ID before the write is genuinely required, and the outcome gate already scores 10/10 with and without this change.

Verdict on the comparison was `not-required` (pairwise: production wins on tool accuracy, error handling and efficiency), so there is no result here that demonstrates a fix.

### What is worth keeping, separately

The production run failed a call for a reason unrelated to this PR, and it looks reproducible: `POST /stores/v3/products/query` returned 400 because the filter was sent as stringified JSON rather than an object, and the agent fell back to fetching all products and filtering in memory. Worth a separate, narrower change to the query recipe if it reproduces.

The API facts verified while investigating — `DELETE https://www.wixapis.com/stores/v3/products/{productId}` with no body; `POST /stores/v3/bulk/products/delete` taking `{"productIds": [...]}`, 1–100 ids, with per-item `results[].itemMetadata.success` (a 200 is not per-item success); and `POST /stores/v3/bulk/products/delete-by-filter` being asynchronous, returning only a `jobId`, and deleting every fuzzy match — are all correct and re-usable, but they belong wherever a request of this shape actually routes, and this scenario shows that is not a recipe at all.

PR #805, opened 2026-07-30
refactor(ecommerce): drop explicit MCP tool names from api-recommendation-tracking

itayhewix:
Second in the series, after #789. Skill instructions should say **what** to do, not name the tool to do it with.

## Why this file needed care

Unlike the mechanical cases in this series, both tool-name sites here carried real information, so this is a rewording rather than a deletion.

**The "internal-but-callable" callout** did two useful jobs: it warned agents not to hunt for an external docs article (there isn't one), and it asserted the service *is* reachable so the tracking step doesn't get skipped as unavailable. Both points are kept, minus the tool names.

**"How to call these APIs"** replaced its `CallWixSiteAPI(...)` signature with the base URL plus the per-endpoint method rule. This also fixes an inaccuracy: the old block showed `method: "POST"` for every endpoint, but Get is `GET` and Update is `PATCH`.

The 13 endpoint sections already used `**Endpoint**: POST <url>` + a JSON body, so they needed no change. Net −4 lines.

## Routing cues deliberately left alone

#789 turned out to need a scope fix — its cues claimed "ANY business improvement request", which is why the traffic-acquisition guard fired. I checked for the same problem here and am **not** changing anything: this skill's broad "ALWAYS LOAD BEFORE generating a recommendations response" description is precisely what its own eval asserts (tracking as a cross-domain obligation, triggered even from a non-pricing intent). Narrowing it would break the thing it's built to guarantee.

## Coverage

`yaml/wix-manage-evals/ecommerce/api-recommendation-tracking.yml` — `ReadFullDocsArticle` assertion plus an `llm_judge`. The `SKILL.md` index entry already exists and names no tool, so no index change.

## Note for reviewers on eval variance

Worth knowing from #789: this project's gate showed real run-to-run variance, with the prod side of `recommend-ecommerce-strategy` scoring 3/10, 10/10, 10/10 and 0/10 across four runs of an unchanged baseline. If a single assertion goes red here, check whether it reproduces before treating it as a regression.


itayhewix:
## Gate is non-deterministic — flagging for the reviewer

The gate has run five times on this PR. **Runs 4 and 5 evaluated byte-identical content** (run 5 was a re-run of run 4's job, no push in between) and produced opposite results on two of three scenarios:

| Scenario | Run 4 (PR / prod) | Run 5 — same commit (PR / prod) |
|---|---|---|
| `api-recommendation-tracking` | **10** / 0 | **0** / 3 |
| `recommend-ecommerce-strategy` | 0 / 2 | **10** / **10** |
| `recommend-strategy-not-traffic-growth` | 10 / 10 | 10 / 10 |

Full history of the `llm_judge` scores across all five runs:

| Run | Content | tracking (PR/prod) | strategy (PR/prod) | traffic guard (PR/prod) |
|---|---|---|---|---|
| 1 | tool-name refactor | 2 / 0 | not selected | not selected |
| 2 | + scope fix | 2 / 0 | 10 / 10 | 10 / 10 |
| 3 | + persist fix | 3 / 10 | 10 / 0 | 10 / 10 |
| 4 | rebase (no content change) | 10 / 0 | 0 / 2 | 10 / 10 |
| 5 | **identical to run 4** | 0 / 3 | 10 / 10 | 10 / 10 |

Both failing scenarios have failed on the **prod** side too, on unmodified `main` — so the gate error (`no winner because both runs failed assertions`) is not evidence that this PR regressed anything. The only stable result across all five runs is the traffic-acquisition guard, green on both sides every time.

### What this PR actually contains

1. **Tool names removed** from `api-recommendation-tracking.md` — the point of the change. Also fixes a real inaccuracy: the old `CallWixSiteAPI` block showed `method: "POST"` for every endpoint, but Get is `GET` and Update is `PATCH`.
2. **Restores checkout-conversion scope lost in #789.** `recommend-ecommerce-strategy.md:179` activates the SHIPPING domain on "checkout conversion, cart abandonment", but #789 narrowed the frontmatter description and `SKILL.md` routing line to an enumeration that omitted both. #789's own gate could not catch this: scenario selection is per-changed-file, so it only ran the two strategy scenarios. The traffic-acquisition exclusion from #789 is preserved.
3. **Hoists the persist-before-presenting mandate** in the tracking file, which previously lived in one prose line and a section at the bottom of a 415-line file while the description only said "load".

Two files rather than one, deliberately: the gate selects scenarios from changed files, so the tracking and strategy scenarios only run together when one PR touches both.

### Ask

Given runs 4 and 5, I don't think further edits can produce a reliably green gate here — I'd be tuning against noise. Requesting a human read on (a) whether the change itself is sound, and (b) whether a gate this variable should block. The persist-lifecycle failure is worth its own issue: it fails on unmodified `main`, so it predates this PR.

itayhewix:
Closing per review — the eval-driven changes here were tuning against a non-deterministic gate (see the five-run table above; runs 4 and 5 were byte-identical and inverted).

One change in this PR is still worth landing, so it is going out on its own: restoring `checkout conversion` / `cart abandonment` to the frontmatter description and the `SKILL.md` routing line for Recommend: eCommerce Strategy. `recommend-ecommerce-strategy.md:179` activates the SHIPPING domain on exactly those cues, but #789 replaced the open-ended scope with an enumeration that omitted them — so the skill body handles a request type its front door no longer advertises. That is a diff-readable inconsistency, independent of any eval score, and it was my error in #789.

The tool-name refactor for `api-recommendation-tracking.md` returns to the queue with the remaining ecommerce files.

PR #813, opened 2026-08-02
feat(wix-manage): add a gift-cards recommendation skill

ohadcawix:
## What

Adds **Recommend: Gift Cards** (`skills/wix-manage/references/ecommerce/recommend-gift-cards.md`) — a single-domain recommendation skill that returns one `gift_cards` / `create_gift_card_product` recommendation carrying the full product design: name, description, preset denominations, custom amount range, and expiration policy.

Structure follows `recommend-ecommerce-strategy.md`: `layer: R` frontmatter with non-loading `references`, scope gate → EXECUTION RULES → numbered mandatory steps → validation table → tracking persistence → output format → error table → constraints.

## Why these specific rules

Three details verified against the Gift Card Products spec drove the design:

- **A site supports a maximum of one gift card product** (409 `GIFT_CARD_PRODUCT_ALREADY_EXISTS`). Step 3 queries for an existing product and stops the flow instead of persisting a recommendation that can only fail at execution time.
- **The public create API needs more than the recommendation carries.** `presetVariants[]` requires both `price.amount` and `value.amount`, and expiry is `expirationType` `NONE`/`RELATIVE` with `relativeExpirationDate {value, period}`. Step 8 maps every param onto those fields so the execution step doesn't guess.
- **Denominations are computed, not templated.** Anchor on the site's AOV, falling back to the catalog median when there are no orders; `×0.5/1/2/4` through a nice-rounding table; floored by the cheapest product and capped at 2× the most expensive. No stock 25/50/100 ladder.

Expiry defaults to **none** — it's regulated in many jurisdictions (US CARD Act 5-year floor, several states ban it outright, EU consumer minimums) and unexpiring cards convert better. Set only on explicit request, minimum 60 months, with the compliance caveat stated in `reasoning`.

## Scope

Gift cards only. Discounts, coupons, shipping, and abandoned checkout are routed to `recommend-ecommerce-strategy`; issuing/redeeming an individual card is the Gift Cards API, and editing an existing product is a direct `UpdateGiftCardProduct` call. The skill is recommendation-only — it never creates the product, and persists via the existing agentic-recommendations tracking service with `domain: "gift_cards"`.

## Checklist

Per [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) → *Adding a Wix Manage Skill*:

- [x] Skill markdown under `skills/wix-manage/references/ecommerce/`
- [x] Entry in `skills/wix-manage/SKILL.md` (routing bullet + index entry in the eCommerce section)
- [x] Entry in `yaml/wix-manage/ecommerce/documentation.yaml` → slug `.../e-commerce/skills/recommend-gift-cards`
- [x] Eval scenario at `yaml/wix-manage-evals/ecommerce/recommend-gift-cards.yml`, tagged `[ecommerce]` — seeds a `stores-v3-editor` site with three products at 18/45/210, then checks the existing-product gate, data-derived amounts, the no-expiry default, tracking persistence, and that the product is **not** created before approval
- [x] Public REST APIs only — every endpoint, field name, enum value, and error code was verified against the Gift Card Products spec; no internal service or gRPC names in the skill
- [x] `description` is 724 chars (limit 1024)


skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
+---
+name: "Recommend: Gift Cards"
+description: Gift-cards recommendation skill — analyzes site data, the catalog price distribution, and whether the site already sells gift cards, then returns ONE recommendation to create a gift card product (name, description, preset denominations, custom amount range, expiration policy). Entry point for "should I sell gift cards", "add a gift card", "gift card ideas", "what amounts should my gift card have", and gifting-holiday prompts. Recommendation-only — it never creates the product; approval hands off to Create Gift Card Product. Tracking is built-in. Out of scope — discounts, coupons, shipping, abandoned checkout (route those to Recommend: eCommerce Strategy) and issuing/redeeming individual gift cards (Gift Cards API).
+layer: R
```
itayhewix: remove

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
+---
+name: "Recommend: Gift Cards"
+description: Gift-cards recommendation skill — analyzes site data, the catalog price distribution, and whether the site already sells gift cards, then returns ONE recommendation to create a gift card product (name, description, preset denominations, custom amount range, expiration policy). Entry point for "should I sell gift cards", "add a gift card", "gift card ideas", "what amounts should my gift card have", and gifting-holiday prompts. Recommendation-only — it never creates the product; approval hands off to Create Gift Card Product. Tracking is built-in. Out of scope — discounts, coupons, shipping, abandoned checkout (route those to Recommend: eCommerce Strategy) and issuing/redeeming individual gift cards (Gift Cards API).
```
itayhewix: try making shorter and more focus

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
...
+| `siteData.industry` / `subIndustry` | Gifting-affinity signal (Step 4a) |
+| `siteData.aov`, `orders30d`, `visitors30d` | Denomination anchor + urgency |
+| `siteData.catalogAnalytics` | Price quantiles, min/max price, product count |
+| `siteData.hasCatalog` | Eligibility gate |
+
+**Do not re-fetch anything the loader already put in context.**
+
+---
+
+## Step 2: Load gift-card recommendation history (Tracking)
```
itayhewix: remove this section we don't use it any more, don't use agentic-recommendations API any more -  follow this good example - https://github.com/wix/skills/tree/main/skills/wix-manage/references/ecommerce/pricing-promotions how we provide discount recommendations

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
...
+| `siteData.hasCatalog === false` (0 products) | **Stop.** "Gift cards need something to spend them on. Add products to your catalog first." |
+| No price data at all — `catalogAnalytics` missing **and** `aov` unavailable | **Stop.** "Cannot size gift card denominations — no order or catalog price data for this site." Do not fall back to generic $25/$50/$100. |
+| `orders30d === 0` **and** `visitors30d < 100` | Continue, but urgency is `LOW` and `reasoning` must say the store has little traffic to convert yet. |
+| Everything else | Continue. |
+
+### Step 4b: Gifting-holiday window
+
+Using `current_date` and `siteData.country`, find the nearest gifting occasion within **75 days**. Gift cards are the classic last-minute gift, so the product must be live **≥ 21 days before** the date to catch the buying window — that lead time is the whole reason urgency matters here.
+
+| Occasion | Typical date | Regions |
```
itayhewix: I suggest to remove this table as it specific to US, we want to support all regions, so you can give an example for US holiday and 1 from EU

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
...
+   | 50 – 199 | 10 |
+   | 200 – 499 | 25 |
+   | ≥ 500 | 50 |
+
+3. Floor every value at `10`. Drop duplicates. Sort ascending.
+4. If the lowest preset is below `priceMin`, raise it to `priceMin` rounded **up** with the table — a card that can't buy the cheapest item is dead weight.
+5. If `priceMax` is known, cap the highest preset at `priceMax × 2` rounded with the table.
+6. Keep **3–5 presets**. Fewer than 3 after dedupe ⇒ add `anchor × 3` (rounded) until you have 3.
+
+**Worked example** — `aov = 62`, `priceMin = 18`, `priceMax = 210`, currency `USD`:
```
itayhewix: currency is based on the site currency - it could be any - make sure all the numbers don't point to any specific symbol

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
...
+| 7 | `customVariant.minValue < customVariant.maxValue` |
+| 8 | Every amount is a decimal string (`"60"`, not `60`), scale ≤ 4, in `siteData.currency` |
+| 9 | `expirationMonths` either unset or ≥ 60, and set only because the merchant asked |
+| 10 | `name` / `description` in `siteData.language`; `title` / `reasoning` in English |
+| 11 | `reasoning` cites the actual numbers and names the call each came from |
+| 12 | Exactly **one** recommendation, `domain: "gift_cards"`, `action: "create_gift_card_product"` |
+
+---
+
+## Step 7: Persist the recommendation (Tracking)
```
itayhewix: we don't use it any more - agentic-recommendations but outcome JSON can be used for future use

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
...
+
+`expiresAt` = the earlier of *(gifting occasion − 7 days)* and *(current_date + 30 days)*; `current_date + 30 days` when there's no occasion in window. Past that point the data is stale and the window is gone.
+
+**Save `id` and `revision` from the result** — the merchant needs them to approve, and Step 8 needs them for the state transitions. If BatchCreate fails, present the recommendation anyway, without a tracking id, and say tracking failed.
+
+Omit `giftCardProduct.expirationMonths` entirely when there's no expiry. Do not send `null`.
+
+---
+
+## Step 8: Present, and hand off on approval
```
itayhewix: no process any more

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
...
+| `customVariant.minValue` / `maxValue` | `customVariant.minValue.amount` / `customVariant.maxValue.amount` | Decimal strings. |
+| `expirationMonths` unset | `expirationType: "NONE"` | The default. |
+| `expirationMonths: 60` | `expirationType: "RELATIVE"`, `relativeExpirationDate: { "value": 60, "period": "MONTHS" }` | `period` ∈ `DAYS`/`WEEKS`/`MONTHS`/`YEARS`. `FIXED` + `fixedExpirationDate` is the other option — this skill never uses it. |
+| — | `image` | Not part of the recommendation. Optional at creation; Wix Media Manager images only (`id`, `width`, `height` required). Upload first if the merchant wants one. |
+
+After a successful create, call MarkDone with the new `giftCardProduct.id` in `executionResult.resultPayload`, and give the merchant the dashboard link.
+
+---
+
+## Output format
```
itayhewix: keep this output but no tracking or using agentic-recommendations

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
```diff
@@ -0,0 +1,394 @@
...
+- Exactly **one** recommendation, always `domain: "gift_cards"` / `action: "create_gift_card_product"`.
+- One gift card product per site — Step 3 is not optional.
+- Never create, update, or delete anything in this skill. Recommendation only.
+- Denominations derive from this site's AOV and catalog prices. No stock $25/$50/$100 ladder.
+- No expiry unless the merchant asked; then ≥ 60 months, with the compliance caveat stated.
+- Amounts are decimal strings in the site's currency. Never assume USD.
+- `presetVariants[].price` and `.value` are always equal — promotional gift-card pricing is out of scope.
+- Persist via BatchCreate before presenting, unless `SKIP_TRACKING`.
+
+## References
```
itayhewix: make sure all urls are valid

yaml/wix-manage-evals/ecommerce/recommend-gift-cards.yml
```diff
@@ -0,0 +1,75 @@
+name: ecommerce/recommend-gift-cards
+description: A "should I sell gift cards, and at what amounts" prompt should load the gift-cards recommendation skill, check for an existing gift card product first, size denominations from the site's own catalog prices, and persist one gift_cards recommendation without creating the product.
```
itayhewix: We can't setup the site for GiftCard - follow this example and make it shorter - https://github.com/wix/skills/blob/main/yaml/wix-manage-evals/ecommerce/pricing-promotions/ecom-pricing-goal-increase-aov.yml

itayhewix (changes requested):

itayhewix (changes requested):
## Required changes

**1. Tracking claim is wrong — recommendations must be persisted, not skipped**

Every other recommendation flow in this tree treats persistence as non-optional. `api-recommendation-tracking.md` states: *"Every recommendation generated by any domain (discounts, shipping, etc.) **MUST** be persisted here before being presented to the merchant."* `recommend-ecommerce-strategy.md` implements this as mandatory Step 8 (`BatchCreate` → `PROPOSED`, `id`/`revision` echoed in the output), and its own field-rules table already reserves this exact domain: `"domain" | "discounts" or "shipping" (future: "gift_cards", "taxes")`.

This PR does the opposite — Step 6 says *"There is no persistence step — do not write the recommendation to any tracking service,"* the Constraints section repeats it, and the eval asserts *"Tracking/persistence is NOT expected."* Recommendations are tracked as standing convention across this codebase; this file just doesn't do it, and shouldn't claim otherwise.

Required fix:
- Add the `API: Recommendation Tracking` reference.
- Add a persistence step mirroring Step 8 of `recommend-ecommerce-strategy.md` (`BatchCreate` with `domain: "gift_cards"`, honoring `SKIP_TRACKING`).
- Return `id`/`revision` in the output, matching the shared schema.
- Update `yaml/wix-manage-evals/ecommerce/recommend-gift-cards.yml` — remove the "tracking not expected" carve-out and assert persistence instead.

**2. Wire this into the orchestrator instead of adding a fourth parallel entry point**

Pricing and shipping route concrete actions through thin dispatchers (`ecom-pricing.md`, `ecom-shipping.md`), but anything that *generates a recommendation* funnels through the single orchestrator, `recommend-ecommerce-strategy.md`, which owns cross-domain dedup, tracking, and history-aware re-proposal (Step 2: don't re-propose `PROPOSED`/`REJECTED`). Domain/goal-specific logic (AOV goal, clearance goal, seasonal goal) lives in its own file but is *loaded by the orchestrator*, not routed to directly from `SKILL.md` — the "Internal skills" block explicitly marks these leaves "do NOT use directly."

This PR instead adds `recommend-gift-cards.md` as a new top-level `SKILL.md` routing bullet alongside "Recommend: eCommerce Strategy" itself, reimplementing the full recommendation lifecycle (site loading, validation, output format, error handling) as its own island. The `documentation.yaml` comment says the quiet part out loud: *"Entered directly from the README routing block ... not via the strategy orchestrator."* Practically, this means gift-card recommendations skip the history check other domains get (a merchant could be re-pitched the same gift-card recommendation every session), and the shared recommendation contract now has to be kept in sync by hand in two places.

Required fix:
- Wire gift cards in as a domain the orchestrator activates in Step 4 (same shape as the DISCOUNTS/SHIPPING rows), exactly as `recommend-ecommerce-strategy.md` already anticipated.
- Move the domain-specific logic (existing-product gate, eligibility, denomination sizing, expiry policy) into a file loaded *by* the orchestrator, the same relationship `ecom-pricing-goal-increase-aov.md` has to it.
- Drop the standalone `SKILL.md` routing bullet in favor of the orchestrator's own domain table.

---

Everything else — the `GIFT_CARD_PRODUCT_ALREADY_EXISTS` gate, the AOV/median-price anchor with rounding table, the min/max clamps, the no-expiry-by-default reasoning with jurisdiction caveats, and the `price`/`value` field-mapping table for the real create call — is solid and doesn't need to change.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Removed — the file has no `layer` key. Done in c90e8ce.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Shortened from 724 to ~430 chars and refocused on what the skill does plus its triggers. Done in c90e8ce; rewritten again in 30bcc50 now that the file is a goal file loaded by the orchestrator.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Table removed. Replaced with a rule to resolve occasions for `siteData.country` plus two illustrations of why you can't assume one calendar: Mother's Day (2nd Sunday of May in US/CA/AU vs late March in UK/IE) and year-end gifting (Dec 25 across most of Europe vs Reyes on Jan 6 in Spain and much of Latin America). Other occasions are listed as names to resolve per country, not dates.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Every currency symbol is gone (grep-verified). Amounts are bare decimal strings in `siteData.currency`, the rounding table is stated as currency-agnostic, display uses the API's `formattedAmount`, and the worked example now uses EUR/IE so nothing reads as US-shaped. Also added an execution rule and a validation check for it.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: All doc URLs curl-checked — every one returns 200. The gift-card endpoints and field names came from the Gift Card Products spec rather than memory.

yaml/wix-manage-evals/ecommerce/recommend-gift-cards.yml
ohadcawix: `siteSetup` and its bootstrap steps are gone. The scenario now follows goal-increase-aov: shared site, one trigger prompt, doc assertions plus an llm_judge — 84 lines down to 18. It moved to `yaml/wix-manage-evals/ecommerce/gift-cards/` and now asserts the orchestrator article too.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Leaving this open rather than marking it fixed, because your later review reverses it. I removed the tracking section here in c90e8ce as asked; the 14:03 review then required persistence to be reinstated (`BatchCreate` with `domain: "gift_cards"`, honoring `SKIP_TRACKING`, `id`/`revision` in the output). 30bcc50 implements that newer instruction — persistence now rides the orchestrator's existing Step 8 batch, so there's no second parallel call. If "we don't use agentic-recommendations any more" is still the direction, that's a conflict worth settling in one place, and it likely also applies to `recommend-ecommerce-strategy.md` and `api-recommendation-tracking.md`.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Same conflict as the line-62 thread — removed in c90e8ce, then reinstated in 30bcc50 per the 14:03 review. Leaving open until the tracking direction is settled.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Partly restored. The Approve → MarkExecuting → MarkDone transitions are back in 30bcc50, since the newer review requires the `id`/`revision` contract that those transitions consume. What stays gone is this file re-documenting the lifecycle: it points at `api-recommendation-tracking.md` and keeps only the gift-card-specific create mapping.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
ohadcawix: Output kept as you asked. `id`/`revision` are back in it per the 14:03 review, which asked for the shared schema. Leaving open alongside the other tracking threads.

ohadcawix:
## Second review addressed — 30bcc50

**2. Orchestrator wiring** — done as specified.

- `GIFT_CARDS` is now a domain the orchestrator activates in Step 4, same row shape as `DISCOUNTS`/`SHIPPING`, with a Step 4b load row, a Step 6 rules summary, `create_gift_card_product` in the action-type table, and the field-rules `domain` line updated from *(future: `"gift_cards"`)* to current.
- The skill became a goal file loaded **by** the orchestrator — renamed **Goal: Sell Gift Cards**, moved to `references/ecommerce/gift-cards/`, and fronted with the same ⛔ routing gate the pricing flow files use. Its stop-the-flow behavior became *drop-the-domain*, so a freed slot goes to another domain.
- `SKILL.md` lost the standalone routing bullet and section entry; gift cards route through the orchestrator bullet, and the goal file sits under **Internal skills — do NOT use directly**. The `documentation.yaml` comment that said the quiet part out loud is gone.

**1. Tracking — implemented, but flagging a reversal.**

This contradicts the earlier review on this PR, which said *"remove this section we don't use it any more, don't use agentic-recommendations API any more"* (line 62), *"no process any more"* (302), and *"keep this output but no tracking"* (326). I had removed all of it in c90e8ce on those instructions.

I implemented the newer instruction, because it's the later word and the tree backs it — `api-recommendation-tracking.md` says persistence is a MUST, and the orchestrator already reserved this domain. Specifically: the tracking reference is back; persistence rides the orchestrator's existing Step 8 `BatchCreate` with `domain: "gift_cards"` honoring `SKIP_TRACKING` (one batch for all domains, not a second parallel call); `id`/`revision` are echoed and drive Approve → MarkExecuting → MarkDone/MarkFailed; validation gained the history check; and the error table covers `RECOMMENDATION_SUPPRESSED` and `VERSION_MISMATCH`. The eval asserts persistence instead of carving it out.

Worth settling in one place: if *"we don't use agentic-recommendations any more"* is the real direction, this file isn't where it should be decided — it would also apply to `recommend-ecommerce-strategy.md` Step 8 and to `api-recommendation-tracking.md` itself. I left the four tracking-related threads open rather than marking them resolved, since resolving them would misrepresent what the code now does. The other six threads are fixed and resolved.

Kept from round one: no `layer` key, short description, no region-specific holiday table, no hard-coded currency symbols, all doc URLs verified 200.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
itayhewix: Settled — see the full explanation on the review below. Short version: tracking stays 100% owned by the orchestrator's Step 2 (history) / Step 8 (BatchCreate); this file shouldn't restate any of the BatchCreate/Approve/MarkExecuting/MarkDone/SKIP_TRACKING/id+revision mechanics, matching how none of the pricing goal files (e.g. ecom-pricing-goal-increase-aov.md) mention tracking at all. My earlier review asking for it to come back here was wrong — sorry for the churn.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
itayhewix: Settled — see the full explanation on the review below. Short version: tracking stays 100% owned by the orchestrator's Step 2 (history) / Step 8 (BatchCreate); this file shouldn't restate any of the BatchCreate/Approve/MarkExecuting/MarkDone/SKIP_TRACKING/id+revision mechanics, matching how none of the pricing goal files (e.g. ecom-pricing-goal-increase-aov.md) mention tracking at all. My earlier review asking for it to come back here was wrong — sorry for the churn.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
itayhewix: Settled — see the full explanation on the review below. Short version: tracking stays 100% owned by the orchestrator's Step 2 (history) / Step 8 (BatchCreate); this file shouldn't restate any of the BatchCreate/Approve/MarkExecuting/MarkDone/SKIP_TRACKING/id+revision mechanics, matching how none of the pricing goal files (e.g. ecom-pricing-goal-increase-aov.md) mention tracking at all. My earlier review asking for it to come back here was wrong — sorry for the churn.

skills/wix-manage/references/ecommerce/recommend-gift-cards.md
itayhewix: Settled — see the full explanation on the review below. Short version: tracking stays 100% owned by the orchestrator's Step 2 (history) / Step 8 (BatchCreate); this file shouldn't restate any of the BatchCreate/Approve/MarkExecuting/MarkDone/SKIP_TRACKING/id+revision mechanics, matching how none of the pricing goal files (e.g. ecom-pricing-goal-increase-aov.md) mention tracking at all. My earlier review asking for it to come back here was wrong — sorry for the churn.

itayhewix (changes requested):
## Correcting my last review — the tracking ask was wrong

My previous review asked for `agentic-recommendations` persistence to come back into this file. That directly contradicted the inline comments already on this PR ("remove this section we don't use it any more, don't use agentic-recommendations API any more", "no process any more", "keep this output but no tracking or using agentic-recommendations") — I hadn't seen those threads when I wrote it. Sorry for the churn. Here's the settled direction:

**`recommend-ecommerce-strategy.md` and `api-recommendation-tracking.md` are correct as-is and don't need to change.** The orchestrator already owns tracking end-to-end for every domain it activates — Step 2 (recommendation-history load) and Step 8 (`BatchCreate` persistence) apply to `gift_cards` automatically now that it's a domain in Step 4, exactly like `discounts`/`shipping`. Nothing about that is deprecated.

**What's wrong is that this file re-describes tracking mechanics it doesn't own.** None of the pricing goal/flow files do this — grepped all of `pricing-promotions/` and `shipping/`: zero mentions of `BatchCreate`, `Approve`, `MarkExecuting`, `MarkDone`, `SKIP_TRACKING`, or a recommendation-level `id`/`revision`/`expiresAt` anywhere. `ecom-pricing-goal-increase-aov.md` never mentions tracking once. `ecom-pricing-flow-upsell-boost.md` goes straight from computing the discount to calling Create Discount Rule — no tracking-transition language at all. That's the model: the goal file owns the domain-specific recommendation content; the orchestrator silently wraps it in persistence.

Concretely, please strip from `gift-cards/ecom-gift-cards-goal-sell-gift-cards.md`:

- Frontmatter `references` — drop the `API: Recommendation Tracking` entry.
- Step 5 ("Hand the recommendation to the orchestrator for persistence") — drop the `BatchCreate`/`SKIP_TRACKING`/echoed-`id`-`revision` description. Just return the recommendation object: `title`, `reasoning`, `domain`, `urgency`, `action`, `params`, `successCriteria`. No `id`, `revision`, `expiresAt`, or `shortTitle`-nested-in-`params` — none of these appear anywhere else in the tree; they were only introduced by my incorrect review.
- Step 6 ("On approval — create the product") — drop "follow the tracking transitions... Approve → MarkExecuting → create → MarkDone/MarkFailed." Once the merchant approves, just call Create Gift Card Product directly, the same plain way `ecom-pricing-flow-upsell-boost.md` Step 7 does.
- Error table — drop `RECOMMENDATION_SUPPRESSED` and `VERSION_MISMATCH` (tracking-API errors, not Gift Card Products API errors).
- Constraints — drop "Persistence is the orchestrator's Step 8 and is mandatory unless `SKIP_TRACKING`."
- References section — drop the `API: Recommendation Tracking` link.
- Field-rules table — drop the `id` / `revision` row.

The orchestrator-wiring work from the second round (domain activation, routing gate, drop-the-domain behavior, moving into `gift-cards/`) is correct and doesn't need to change.

Once this is in, please go ahead and resolve the four open tracking threads — they were correctly left open pending this call.

itayhewix (dismissed):

ohadcawix:
#skipreview

itayhewix (approved):

PR #814, opened 2026-08-02
fix(ecommerce): re-add checkout conversion to strategy routing cues

itayhewix:
Two-line fix for an inconsistency I introduced in #789. Extracted from #805, which was closed — the rest of that PR was tuning against a non-deterministic gate, but this part is readable straight from the diff and doesn't depend on any eval score.

## The problem

`recommend-ecommerce-strategy.md:179` activates the SHIPPING domain on these cues:

> `| **SHIPPING** | Merchant mentions shipping, delivery, checkout conversion, cart abandonment. ...`

#789 replaced this skill's open-ended scope ("Use this for ANY business improvement request") with an explicit enumeration, in order to exclude traffic acquisition. But the enumeration omitted checkout conversion and cart abandonment:

| Location | On main today |
|---|---|
| body `:179` | shipping, delivery, **checkout conversion, cart abandonment** |
| frontmatter `description` | sales, promotions, discounts, coupons, clearance, holiday deals, AOV, shipping |
| `SKILL.md:167` | boost sales, promotions, help my business, holiday deals, improve revenue, discounts, shipping, coupons, clearance |

So the skill body handles a request type its own front door no longer advertises.

That was my error in #789. Excluding traffic acquisition was right; enumerating the in-scope list was a reasonable mechanism for it. What I got wrong was writing that list from the existing description instead of checking it against the domain-activation table in the body.

## The fix

Adds `checkout conversion` and `cart abandonment` to both cue sites. #789's traffic-acquisition exclusion is untouched — it has been green on both sides of every gate run I have seen (five consecutive), and this change does not widen scope back toward acquisition.

Description is 513 chars, within the 1024 limit.

Two files because `SKILL.md` is this skill's index entry, per CONTRIBUTING step 2.

## Coverage

Covered by the existing `yaml/wix-manage-evals/ecommerce/recommend-ecommerce-strategy.yml` and `recommend-strategy-not-traffic-growth.yml`.

Heads-up for whoever reviews: this gate has been unreliable on these scenarios. On #805, runs 4 and 5 evaluated byte-identical content and inverted on two of three scenarios ([detail](https://github.com/wix/skills/pull/805#issuecomment-5157950723)). Worth re-running before reading a single red result as a regression.

### DIFF
--- a/skills/wix-manage/SKILL.md
+++ b/skills/wix-manage/SKILL.md
@@ -164,7 +164,7 @@ These recipes do NOT cover frontend development or SDK usage for displaying data
 ## eCommerce
 
 **Routing — pick the right entry point:**
-- **Any sales/business improvement request** (boost sales, promotions, help my business, holiday deals, improve revenue, discounts, shipping, coupons, clearance) → use [Recommend: eCommerce Strategy](references/ecommerce/recommend-ecommerce-strategy.md). This is the **default entry point** — it analyzes ALL domains (discounts, shipping) and generates cross-domain recommendations. Do NOT ask clarifying questions.
+- **Any sales/business improvement request** (boost sales, promotions, help my business, holiday deals, improve revenue, discounts, shipping, coupons, clearance, checkout conversion, cart abandonment) → use [Recommend: eCommerce Strategy](references/ecommerce/recommend-ecommerce-strategy.md). This is the **default entry point** — it analyzes ALL domains (discounts, shipping) and generates cross-domain recommendations. Do NOT ask clarifying questions.
 - **Traffic acquisition is NOT an eCommerce-strategy request** ("grow my traffic", SEO, ads, social, content) → do NOT use Recommend: eCommerce Strategy; it only converts visitors a store already has. Route these to marketing.
 - **Pricing & promotions** (coupons, discount rules, ribbons, sales) → use the [Pricing & Promotions](references/ecommerce/ecom-pricing.md) dispatcher.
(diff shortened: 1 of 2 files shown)


itayhewix:
Closing — the premise does not hold up.

I claimed #789 dropped `checkout conversion` / `cart abandonment` from this skill's routing cues. Checking `19b49bf~1`, that is wrong for `SKILL.md:167`: its parenthetical read `(boost sales, promotions, help my business, holiday deals, improve revenue, discounts, shipping, coupons, clearance)` before #789 and is byte-identical today. #789 never touched that line's list, so those terms were never there to lose — including in whatever state the tracking scenario historically passed in.

#789 did narrow the frontmatter description, from "Single entry point for any 'help my business' request" to an enumeration. But that enumeration is prefaced with "requests about earning more from the visitors a store already has", which covers checkout conversion, and `SKILL.md:167` still leads with "**Any** sales/business improvement request" — so the parenthetical is illustrative, not a whitelist. The front door is not closed to checkout-conversion phrasing.

That leaves adding two example keywords to lists already introduced by language that covers the case. My evidence for actual harm was the `api-recommendation-tracking` scenario failing on an "improve my checkout conversion" trigger, and I subsequently demonstrated that scenario is non-deterministic (byte-identical runs inverting: https://github.com/wix/skills/pull/805#issuecomment-5157950723). With that gone, there is no measured benefit here — only a tidiness argument I overstated as a defect.

No change needed on main.

PR #818, opened 2026-08-03
fix(wix-headless): searchProducts returns products not items (GPV-8737)

gilgwix:
## Summary

- `searchProducts()` from the Wix Catalog V3 API returns `{ products, pagingMetadata }`, not `{ items }`
- The headless skill docs instructed agents to destructure `items`, which is always `undefined` → product gallery silently renders empty
- Fixed the comment (line 55) and the live code snippet (line 122) in `how-to-code-a-store.md`

## Changes

`skills/wix-headless/references/inline-recipes/how-to-code-a-store.md`
- Line 55: `result.items[]` → `result.products[]` (cheat-sheet comment)
- Line 122: `const { items }` → `const { products }` (category-filter code snippet)

## Test plan

- [ ] Confirm no `{ items }` destructuring from `searchProducts` remains in the file
- [ ] Confirm the fix matches the pattern already documented in `skills/wix-app/references/stores/QUERY.md` (`res.products ?? []`)

## Jira

[GPV-8737](https://wix.atlassian.net/browse/GPV-8737) — searchProducts() returns { products } not { items } — product gallery renders empty

### DIFF
--- a/skills/wix-headless/references/inline-recipes/how-to-code-a-store.md
+++ b/skills/wix-headless/references/inline-recipes/how-to-code-a-store.md
@@ -52,7 +52,7 @@ A concise contract for writing the **frontend code** of a storefront against a C
 The exact field paths the storefront reads, and the **plausible-wrong sibling** each is mistaken for — the sections below reference these instead of re-describing them. All `amount`s are **strings**. These are **read** shapes; the cart-add body (under *Adding to cart*) is a separate **write** shape, and the `_id` rule applies to read **entities**, not to method-return wrappers (note `checkoutId`).
 
 ```jsonc
-// productsV3.queryProducts() / .searchProducts()  →  result.items[]
+// productsV3.queryProducts() / .searchProducts()  →  result.products[]
 product = {
   _id,                                            // links · cart catalogItemId · variant filter   (NOT .id → empty → HTTP 500)
   slug, name, visible,                            // only visible:true is returned to a visitor token
@@ -119,7 +119,7 @@ const res = await categories.queryCategories({
 **⚠️ CRITICAL: category filtering MUST use `searchProducts`, NOT `queryProducts`.** `directCategoriesInfo.categories` is **not declared as filterable in `queryProducts`** — passing it there returns HTTP `400 "... is not declared as filterable"`, which the SDK **swallows silently**, leaving an empty category page that looks like "no products". This is the #1 way this breaks. Use Search Products:
 
 ```js
-const { items } = await productsV3.searchProducts({
+const { products } = await productsV3.searchProducts({
   filter: { 'directCategoriesInfo.categories': { $matchItems: [{ id: categoryId }] } },
 });
 ```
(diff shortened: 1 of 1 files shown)


ozsay:
The core claim checks out, but the line-55 edit overcorrects and introduces the same class of error in the opposite direction.

### The claim is correct, and it's not version-dependent

I unpacked the typings for `@wix/auto_sdk_stores_products-v-3` at `1.0.0`, `1.0.50`, `1.0.100` and `1.0.205` (latest). `SearchProductsResponse` has been the same in **every** published version:

```ts
interface SearchProductsResponse {
    /** Products which satisfy the provided query. */
    products?: V3Product[];
    pagingMetadata?: CursorPagingMetadata;
    aggregationData?: AggregationData;
}
```

No `items`, ever — so this isn't drift from a newer SDK, it was always wrong in the doc. (`V3SearchProductsResponse`, the other public alias, agrees.)

### But the two methods genuinely return different wrappers

| call | return type | array field |
|---|---|---|
| `searchProducts({...})` | `V3SearchProductsResponse` | `products` |
| `queryProducts().…find()` | `ProductsQueryResult` | `items` |

`ProductsQueryResult` is the standard query-builder result and legitimately uses `items`:

```ts
interface ProductsQueryResult extends QueryCursorResult {
    items: V3Product[];
    query: ProductsQueryBuilder;
    next: () => Promise<ProductsQueryResult>;
    prev: () => Promise<ProductsQueryResult>;
}
```

### So line 55 needs a split, not a flip

Line 55 is a **shared header for both methods**:

```jsonc
// productsV3.queryProducts() / .searchProducts()  →  result.items[]
```

Changing it to `result.products[]` fixes it for `searchProducts` and breaks it for `queryProducts` — which really does return `items`. Suggest:

```jsonc
// productsV3.queryProducts().…find()  →  result.items[]
// productsV3.searchProducts({...})    →  result.products[]
```

The **line 122 change is correct as written** — no objection there.

### Heads-up on the test plan

> - [ ] Confirm no `{ items }` destructuring from `searchProducts` remains in the file

Lines 66 and 142 also destructure `items`, but both are `readOnlyVariantsV3.queryVariants().…find()` — query-builder results, so `items` is correct there. They should **not** be changed; worth calling out so the checkbox doesn't get actioned on them.

### Field evidence for GPV-8737

Confirming the symptom is real and scoped to the category route. A generated storefront that followed this recipe:

```js
// category route — BUG: searchProducts returns `products`
const { items } = await client.productsV3.searchProducts({
  filter: { 'directCategoriesInfo.categories': { $matchItems: [{ id: categoryId }] } },
});
if (!cancelled) setProducts(items ?? []);   // <- `?? []` swallows the undefined

// "Shop All" route — correct, queryProducts really does return `items`
const { items } = await client.productsV3.queryProducts().limit(50).find();
```

The `?? []` is what makes it silent: the category page renders "No products found here yet." with no error, while the unfiltered listing page works fine. That matches the existing line-119 warning about `queryProducts` + `directCategoriesInfo` — two independent causes of the same empty-category-page symptom, which is probably why it went unnoticed.

gilgwix:
> Line 55 is a **shared header for both methods**:

addressed the changes @ozsay

ozsay (approved):

PR #658, opened 2026-07-20
🦞 igor-claw: fix wrong Wix Forms appDefId in Create Form recipe

ayal:
## What's broken

The **Create Form** recipe (`skills/wix-manage/references/forms/create-form.md`) tells agents to install appDefId `14ce1214-b278-a7e4-1373-00cebd1bef7c` as "the Wix Forms app" before calling `POST form-schema-service/v4/forms` with namespace `wix.form_app.form`.

That appDefId is actually **"Wix Forms (Old)"** (aka "Wix Forms & Payments") — a different, legacy app. Installing it succeeds (`enabled: true`, `status: "UNKNOWN"`) but it does **not** grant permissions on the `wix.form_app.form` namespace. Create Form then fails with:

```
400 "Permissions for given namespace not found. Please make sure you provided supported
namespace and/or have your app installed on site." (UNSUPPORTED_FORM_NAMESPACE)
```

...even after installing the (wrong) app and retrying multiple times — because it's the wrong app, not a propagation delay. The recipe's own troubleshooting table papered over this with "try creating a form through the UI first to activate the namespace," instead of naming the actual fix.

## Root cause (verified live)

The correct app is **"Wix Forms (New)"**, appDefId `225dd912-7dea-4738-8688-4b8c6955ffc2`:
- Already used correctly in `skills/wix-headless/references/inline-recipes/setup-forms.md` and `SETUP.md` — the two Forms skills disagree on the app ID.
- Listed as a distinct app from "Wix Forms (Old)" in the official ["Apps Created by Wix"](https://dev.wix.com/docs/api-reference/articles/work-with-wix-apis/platform/about-apps-created-by-wix) reference table.
- Repro: on a fresh blank-template site, installing `14ce1214-...` and calling Create Form fails 3/3 times (incl. after delay). Installing `225dd912-...` instead and immediately calling Create Form **succeeds on the first try**, no UI step needed.

## Fix

- `forms/create-form.md`: point every reference (prerequisites text + troubleshooting table, 3 spots) at `225dd912-7dea-4738-8688-4b8c6955ffc2`, and explicitly call out that the old ID looks-installed but doesn't activate the namespace.
- `app-installation/install-wix-apps.md`: add Wix Forms (New) to the common appDefId table, so the "missing installed app" fallback path also lands on the right ID.

## Context

[...]

PR #824, opened 2026-08-04
🦞 igor-claw: fix Create Form recipe's identifier guidance for choice/date fields

ayal:
#### Problem
The Create Form recipe's own recommended pattern for choice/date fields causes the exact bug it warns about. It instructs setting `identifier: "TEXT_INPUT"` for `RADIO_GROUP`/`DROPDOWN` fields (Field Types Reference table + "Choice fields" example), and doesn't cover `DATE_INPUT`/`DATE_PICKER` at all.

Live-verified on `wix.form_app.form` via Create Form: the Form Schemas API (form-schema-service) derives a field's rendered `componentType` on every subsequent read from its `identifier` string, matched against an internal fixed allow-list — completely independent of the `componentType` actually set in the request. Following the recipe's own guidance (`identifier: "TEXT_INPUT"` + `componentType: "RADIO_GROUP"`) makes the field's own create response immediately show `componentType: "TEXT_INPUT"`, discarding `radioGroupOptions` — no error, HTTP 200. Setting `identifier` to match the componentType instead (`"RADIO_GROUP"`, `"DROPDOWN"`, `"DATE_INPUT"`) reproducibly fixes it.

This is the root cause of an MCP feedback report where an agent built a lead-capture form with a dropdown "Preferred Campus" field and a "Date of Birth" field; both silently reverted to `TEXT_INPUT` after creation, and the DOB field's `validation.format` became immutably stuck at `DATE` even though its rendered type was lost.

#### Solution
- Corrected the "Field Types Reference" table and "Choice fields" example to use `identifier` matching `componentType` (`RADIO_GROUP`/`DROPDOWN`) instead of `"TEXT_INPUT"`.
- Added a new "Date fields (DATE_INPUT / DATE_PICKER)" section (previously undocumented), with the same `identifier`-must-match-`componentType` rule.
- Updated the "CRITICAL" bullet and troubleshooting table to describe the actual failure mode: a mismatched `identifier` doesn't just risk a fallback, it unconditionally coerces the field to `TEXT_INPUT` (or drops it if `identifier` is fully custom).

#### Considerations - important facts the reviewer should know
- The underlying service behavior (silently overriding `componentType` based on an undocumented `identifier` allow-list instead of erroring) is a real API bug being raised separately with the Forms team — this PR only fixes the recipe so agents following it stop reproducing that bug.
- Filed automatically by an AI agent acting on behalf of Ayal (ayalg@wix.com), researching Wix MCP user feedback. Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1785825864996019
- Did not touch the file's existing wrong `appDefId` (`14ce1214-...`) — that's already tracked by open PR #658.

### DIFF
--- a/skills/wix-manage/references/forms/create-form.md
+++ b/skills/wix-manage/references/forms/create-form.md
@@ -214,7 +214,7 @@ Verify the form in the dashboard: `https://manage.wix.com/dashboard/{siteId}/for
 
 - All `id` fields (for `formFields`, `steps`) and all `fieldId` references in the layout **must be valid UUIDs**. Generate fresh UUIDs for each form you create — do not reuse the example UUIDs above.
 - Each field needs a unique `id` and a unique `target` value. The `target` is used to map submissions to contact fields.
-- **CRITICAL: The `identifier` must be a recognized Wix value.** Custom identifiers like `"product_name"` or `"color_preference"` will cause the field to be silently dropped from the form — no error is thrown. For any generic/custom text field, use `"TEXT_INPUT"` as the identifier and set the display name via the `label` property in `textInputOptions`.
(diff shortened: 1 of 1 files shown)


ayal:
🦞 Update from an automated AI agent (acting on behalf of Ayal, ayalg@wix.com): this exact root cause (form-schema-service dropping fields whose `identifier` isn't on its internal allow-list) recurred in a fresh Wix MCP feedback report, 5 days after this PR was opened — this time hitting a `TEXT_INPUT` custom field, not just choice/date fields. Live-reproduced and pushed a correction to the proto doc in wix-private/app-market#60593 (identifier's drop behavior applies to TEXT_INPUT too, not just non-TEXT_INPUT/PHONE_INPUT types).

This PR has been open 6 days with 0 human reviews, and its `gate` check has been failing the whole time (EvalForge YAML Gate: Missing Coverage) — flagging in case that's a blocker worth someone's attention rather than a silent stall. Given this is a second live recurrence, could this get prioritized? cc @ozsay

Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1786336721924949

PR #834, opened 2026-08-04
feat(accessibility): add site scan management recipe [WIXSEO-4009]

omeb:
## Summary

- add a `wix-manage` recipe for full-site, single-page, and discovered page-collection accessibility scans
- orchestrate scan → polling → page summaries/findings with truthful partial/failure handling
- turn findings into actionable fix and verification guidance
- stop after the first non-retriable `403` and explain the missing Manage Accessibility Scans authorization
- cover the recipe with a Stores-backed eval

## Validation

- all six Accessibility Scans methods and the skill are published in Docs Portal
- docs and eval YAML parse successfully
- docs entry and `ReadFullDocsArticle` assertion use the same canonical URL
- `git diff --check`

The latest live eval exposed an environment issue: collection discovery succeeds, but its Wix identity lacks permission to start a scan. This change fixes the resulting retry loop; the eval identity still needs Manage Accessibility Scans to validate the successful end-to-end path.


omeb:
The manual eval exposed two concrete issues:

1. The recipe kept probing after the first non-retriable `403`. Fixed in `ff2643d`; it now stops immediately and explains the missing authorization.
2. The EvalForge Wix identity can read page collections but cannot start a scan because it lacks **Manage Accessibility Scans**. A successful end-to-end rerun still requires that permission on the evaluator identity/app; changing the recipe cannot grant it.

The eval assertion now explicitly guards against redundant permission-denied retries without weakening the required successful scan flow.

omeb:
@ayal could you please approve when you have a moment? All checks pass and auto-merge is enabled. Thanks!

PR #837, opened 2026-08-04
docs(stores): note image-attach read-back quirks (re-hosted URL + brief lag)

ayal:
# Feedback

Running the image-attach flow end-to-end on a live Wix Stores V3 site (bulk import → PATCH product → re-GET) surfaced two read-back quirks worth documenting, both of which can make a **successful** attach look failed:

1. **Wix re-hosts the image on attach.** The read-back `media.main.image.url` is a **new** `static.wixstatic.com` id — not the URL you imported/sent (that survives only in the `image.filename` field). String-equality against the imported URL fails even though the attach worked.
2. **The read-back lags the PATCH.** Immediately after the `200`, `media.main.image` can be briefly absent (observed: first re-GET missing it, present a few seconds later). One empty read is not failure.

Both reinforce the existing rule (a PATCH `200` is success). Add two sub-bullets to the "Attach images" verify note so an agent doesn't string-compare URLs or hard-fail on a single empty read-back.

Verified live (site-scoped CLI token), not inferred. Docs-only.


ayal:
Folded into #836 — same investigation (Wix media import). Kept as one PR.

PR #666, opened 2026-07-21
Fix Stores collection lookup in discount rule recipe

polinaka-wix:
# Feedback

Conversation(s): https://wix-bo.com/ai-assistant/conversations/3aa0eb73-9e90-4a06-881a-5704733e1d97
Feedback: Applying a 15% automatic discount to the "Summer" Stores collection failed because the agent tried `POST https://www.wixapis.com/categories/v1/categories/query`, which returned `428 APP_NOT_INSTALLED` for appId `f534c0d3-dd59-4047-a86a-be3234d4591f`.

## Conversation research

The reported issue is a true issue. The user asked: "Apply a 15% discount to all items in the Summer collection." The assistant activated `pricing-create-discount-rule`, tried to discover the collection ID, hit an `ExecuteWixAPI` `TOOL_FAILED`, then responded that it was blocked and asked the merchant to provide the collection ID instead of completing the requested discount.

Sample broken conversation(s):
- `https://wix-bo.com/ai-assistant/conversations/3aa0eb73-9e90-4a06-881a-5704733e1d97`: assistant message `468ae115-f243-4bb2-9897-6b61a041fc3d`; activated skill log `763bbf12-b6d7-492a-a4cf-18abe3bcadf1` (`pricing-create-discount-rule`); failing tool call log `78198519-f558-40c9-a122-39e2741e1c00`; failing result log `7be80abf-5044-419e-96f7-e2d8d5562198`.

## Code/content research

Root cause: the discount-rule recipe covered collection-scoped discount payloads, but did not document how to resolve a merchant-provided Wix Stores collection name to a collection GUID. The model filled that gap by reading Catalog V3 Categories docs and calling `categories/v1/categories/query` with `treeReference.appNamespace: "@wix/stores"`. That endpoint is not the right legacy Stores collection lookup for this flow and can require the Categories app even when Wix Stores is installed.

Why this is the owning surface:
- The conversation's `activateSkill` call loaded `pricing-create-discount-rule`.
- The recipe's error table said `categoryIds` empty -> "Call getCategoryIds to convert category names to GUIDs", but no concrete Wix MCP/API path was provided.
- Official docs for Wix Stores app collections document `POST https://www.wixapis.com/wix-data/v2/items/query` with `dataCollectionId: "Stores/Collections"` for querying Stores collections, and the Data Item FQDN identifies `QueryDataItems` as `POST /v2/items/query` returning `dataItems`.

Alternatives ruled out:
[...]

PR #842, opened 2026-08-04
fix(wix-manage): correct the Create Discount Rule discounts wire shape

danshw:
# Defect

The `pricing-create-discount-rule` recipe documented a Create Discount Rule body that the API
rejects, in two independent ways. Both were in the recipe's own `Critical: discounts structure`
section, so an agent following the page verbatim could not succeed on its first call.

**1. `discounts` was documented as a bare array.** It is an object wrapping a `values` array.

```
"discounts": [ { … } ]
→ 400 {"message":"Expected an object"}
```

**2. Discount entries nested the type and value in a `discount` object.** `discountType` and the
value field sit at the entry root.

```
"discounts": { "values": [ { …, "discount": { "discountType": "PERCENTAGE", "percentage": 10 } } ] }
→ 400 Validation failed: discount_rule.customer_gets.specific_items_discount
      Discount value type does not match discount value
      (violatedRule: OTHER, ruleName: DISCOUNT_TYPE_AND_DISCOUNT_VALUE_MISMATCH)
```

The shape the API accepts:

```json
"discounts": {
  "values": [
    {
      "targetType": "SPECIFIC_ITEMS",
      "specificItemsInfo": { "scopes": [ … ] },
      "discountType": "PERCENTAGE",
      "percentage": 10
    }
  ]
}
```

# Evidence

From a recorded agent run on the request *"Set up an automatic discount: 10% off any order over
$50."* The agent read this recipe, then spent four rejected `POST /ecom/v1/discount-rules`
attempts brute-forcing the payload before succeeding on the fifth:

| Attempt | Body | Result |
|---|---|---|
| 1 | `discounts` array, nested `discount` (exactly as documented) | `400 Expected an object` |
| 2 | `discounts.values`, nested `discount` | `400 DISCOUNT_TYPE_AND_DISCOUNT_VALUE_MISMATCH` |
| 3 | `discounts` array, `discountType` at root | `400 Expected an object` |
| 4 | `discounts.values`, `discountType` at root, `percentage` still nested | `400 DISCOUNT_TYPE_AND_DISCOUNT_VALUE_MISMATCH` |
| 5 | `discounts.values`, `discountType` **and** `percentage` at root | **2xx** |

Between attempts it issued six further schema-probing calls purely to work out the shape. The task
still completed — the outcome judge scored it 10/10 — so this defect is invisible to a
correctness-only assertion and shows up only as cost: the path judge scored **4/10** on two
consecutive daily runs, and the run took ~97 s against a ~50 s median for its batch.

Both corrections were verified against the Discount Rules API schema, which declares `discounts`
as a message wrapping `repeated Discount values`, and declares the value field as a `oneof`
(`percentage` / `fixed_amount` / `fixed_price`) with `discount_type` as its sibling — a `oneof` is
flattened in JSON, so its name is not a wire field. That is why the documented `discount` wrapper
could never have worked.

# Why this surface

The recipe is the file the failing agent actually loaded (its first action after the README was
`ReadFullDocsArticle` on this recipe's URL), and the recipe is where the wrong shape was written.
The API schema and reference are already correct — the agent's own schema lookups returned the
right structure, which is how it eventually recovered. So this is not a rendering or publishing
gap upstream; it is a recipe contradicting the API it documents, and correcting the recipe is the
whole fix.

# Change

Deliberately scoped to the wrong shape and nothing else — every changed line is one of the two
defects above. No reformatting, and the wrapper is added as `"discounts": { "values": [` on one line
so the entries keep their existing indentation and the diff stays reviewable.

- `skills/wix-manage/references/ecommerce/pricing-promotions/ecom-pricing-create-discount-rule.md`
  — corrects the six request/response examples, the `Critical: discounts structure` entry example,
  the key-field table, the find-by-name update pattern, and the recommendation mapping table. The
  two 400s go in the existing `Error Handling` table as two rows, mapping each message to its cause,
  rather than as new prose: the corrected examples already carry the shape, so the only thing worth
  adding is what an example cannot say. In the recorded run the agent *received* both 400s and still
  needed four attempts, because nothing connected the message to the cause.
- `yaml/wix-manage-evals/ecommerce/pricing-promotions/ecom-pricing-create-discount-rule-discounts-shape.yml`
  — new covering scenario. It keeps the original request wording, so nothing in the prompt reveals
  the payload shape; the discrimination is in the rubrics. Three assertions per
  `docs/eval-scenarios.md`: coverage on this recipe's URL, a correctness judge on the created rule,
  and a path judge that fails a run which takes either 400 on the way.

# On "Orchestration, not API shape"

`CONTRIBUTING.md` says request/response shapes belong in the Wix docs, because restating them
"creates a second source of truth that drifts as the API evolves." This recipe already restates seven
full request bodies, and the defect is exactly the drift that rule warns about — so this PR corrects
the restated shapes rather than adding new ones, and adds no new prose restating field names or enum
values. That rule is also why the two 400s became `Error Handling` rows instead of a shape
description at the top of the file.

Removing the restated bodies altogether, in favour of pointing at the Create Discount Rule reference,
would follow the rule more closely and would delete the drift risk at the source. That is a larger
change to a recipe several other recipes and scenarios lean on, so I have kept it out of this fix and
am flagging it as worth doing separately.

# Verification

- `ruby -e 'require "yaml"; YAML.load_file(…)'` on the new scenario — parses.
- Every `json` block that parsed before still parses. Two do not, and both are pre-existing and
  untouched: a `"<scopes unchanged from query response>"` placeholder, and a trigger example with an
  intentional `/* same scope */` comment.
- `git diff --check` — clean.

# The eval gate has not run on this PR

**No workflow has run on this PR, so nothing here is gate-verified.** GitHub Actions appears to be
disabled on the repository right now: the repo's Actions permissions report `enabled: false`, and no
workflow of any kind has started repo-wide since 16:09 UTC, while this PR was opened at 16:52 UTC.
Every PR opened earlier today has runs; this one has zero, for every workflow, not just the gate.

That is a repository-level condition rather than anything about this diff, and I have not changed the
setting. Once Actions is re-enabled, a push or a reopen will trigger `EvalForge YAML Gate` and the
new scenario will run. Please do not read the absence of a red check as a pass.

# Please note before merging

**This PR edits the same recipe as #666, which is open.** The two changes do not overlap in
content — #666 adds Stores collection-ID lookup guidance and touches a different section — and
they do not conflict textually. But both PRs edit the recipe, so both pull in the pre-existing
`ecommerce/pricing-promotions/create-discount-rule` scenario, and the eval gate treats a scenario
claimed by two open PRs as held. If this gate reports scenarios held by another PR, that is the
cause rather than anything in this diff, and landing or closing #666 first clears it. I have not
described this PR's gate as passing.


danshw:
## ❌ EvalForge YAML Gate: Scenario Locked by Another PR

These scenarios are draft-tagged for other PRs. Wait for those PRs to merge/close, or coordinate with their authors:

- `ecommerce/pricing-promotions/create-discount-rule` is held by another open PR: https://github.com/wix/skills/pull/666

PR #879, opened 2026-08-07
docs(seed): drop option-less impl detail from bulkCreateProducts' agent surface

ayal:
The stock-fix docs leaked the internal option-less mechanism ("stocked whether or not they have options — option-less ones stocked internally") into the agent-facing JSDoc + SEED.md table. That re-introduces the options-vs-not mental model we abstracted away. Simplified to what the agent actually needs: **pass `quantity`, products come out in stock.** The private `stockOptionlessProducts` helper keeps its own impl comment for maintainers. node --check clean; base44 copy synced.


ayal:
Folded into #880 (image-readiness fix carries the same option-less doc de-leak).

PR #926, opened 2026-08-08
fix(wix-manage): send each variant complete on a Catalog V3 update

danshw:
`variantsInfo.variants` replaces the array **and each variant object** — it does not merge them. The recipe already said to pass the entire existing array, and to include each existing variant `id`. An agent can satisfy both of those and still rebuild each variant from only its `id` plus the field it wants to change, which drops every field it did not carry. The update is then rejected on the first required one:

```
HTTP 400 {"message":"product is invalid:
`-- variantsInfo is invalid:
    `-- variants [at index 0] is invalid:
       `-- price must not be empty",
 "details":{"validationError":{"fieldViolations":[
   {"field":"product.variantsInfo.variants[0].price",
    "description":"must not be empty","violatedRule":"REQUIRED_FIELD"}]}}}
```

## How it shows up

Setting a SKU on an existing product. From a recorded run, the agent read the recipe, located the product and re-read it with Get Product — then assembled the PATCH as:

```js
const updatedVariants = currentVariants.map(v => ({ id: v.id, sku: "MUG-001" }));
```

The variant's price went with it, and the write was rejected. The agent recovered by retrying with the variant spread intact, so the user still got the right SKU — one rejected write and an extra round trip later.

The error table made this harder to diagnose than it needed to be: it attributed `price must not be empty` to a variant "created or replaced without a price" and prescribed the fix "on every **new** variant", so an agent updating an existing variant reads that row as not applying to it.

## Changes

Two lines in `update-product-with-options.md`:

- The paragraph on updating existing variants now says each variant object is replaced whole rather than merged, and to start from the variant as returned by Get Product.
- The `price must not be empty` row now names the existing-variant case and says to carry `price` on every variant sent, not just new ones.

No request or response shapes are restated — both edits are about how to assemble a call from what Get Product already returned.

## Coverage

New scenario `yaml/wix-manage-evals/stores/update-product-sku-full-variant.yml`, three assertions with the path judge gating at 7, per this repo's eval-scenario guide. Its 1-3 band is specifically a rejected-then-retried Update Product caused by a `REQUIRED_FIELD` violation on `variantsInfo.variants[...]` from fields the agent did not carry over, so the scenario fails on the behaviour this PR is meant to remove rather than on the outcome, which was already correct.

### DIFF
--- a/skills/wix-manage/references/stores/update-product-with-options.md
+++ b/skills/wix-manage/references/stores/update-product-with-options.md
@@ -332,7 +332,7 @@ curl -X PATCH "https://www.wixapis.com/stores/v3/products/{productId}" \
 - To update `variantsInfo.variants`, also pass `options`, and vice versa. Variants and options are mutually dependent and must stay aligned.
 - When converting a simple product to an optioned product, rebuild the variants list so every variant has `choices`; do not keep an existing choice-less default variant unchanged.
 - Always include `choicesSettings` with the complete list of choices when updating a product with options.
-- Use `optionChoiceNames` rather than `optionChoiceIds` in variants for more reliable updates.
+- Use `optionChoiceNames` rather than `optionChoiceIds` in variants for more reliable updates. Reading them back is not symmetric: Get Product returns each variant's `choices` with `optionChoiceIds` only, and fills in `optionChoiceNames` just when the request's `fields` array includes `"VARIANT_OPTION_CHOICE_NAMES"`. So to find the variant for a named choice such as `Large`, either pass that field and match on the name, or take the choice GUID from `options[].choicesSettings.choices[].choiceId` and match it against `variants[].choices[].optionChoiceIds.choiceId`. Matching on a name the response never carried raises nothing — it just selects no variant.
 - Include the `renderType` in `optionChoiceNames`.
 
 ## Error Message Reference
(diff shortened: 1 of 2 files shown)


danshw:
## Eval comparison — the fix works; the scenario still cannot certify it

Fourth run, and the first in which **both arms read the recipe**, so recipe content was genuinely the only variable.

| | PR | production |
|---|---|---|
| Recipe read | yes | yes |
| Correctness | 10/10 | 10/10 |
| Path judge | 5/10 | 5/10 |
| Steps | 5 | 7 |

**Production reproduced the defect this change documents**, verbatim from its path judge:

> First attempt matched choice by name but failed because variants only carry choice IDs, requiring subsequent attempts to resolve choiceId from product options (using `choice.id` first, then correcting to `choice.choiceId`). Suspected gap: lack of clear documentation on how variants map to option choices in Catalog V3.

The PR arm did not hit it at all — it went straight to a working update in two calls. That is the change doing its job.

It nonetheless scored 5, for an unrelated reason: it parsed the Search Products response with the Catalog V1/V2 shape (`productResults.results`) rather than V3 (`products`), and had to re-run. Both arms therefore fail the gating path judge, which yields `not-sure` and returns before any approval decision.

So the change is supported by this run — the documented friction appears without it and not with it — but the scenario cannot award `required` while a separate response-shape confusion keeps the PR arm below the threshold. That is a distinct defect and belongs in the Find Products recipe, not here.

danshw:
Folded into #927. The two changes are coupled: the variant fixes here could not reach a clean path while the Search Products response envelope was still undocumented, so a single gate run over both is the only way to test them together. All commits from this branch were cherry-picked onto #927 unchanged.

PR #927, opened 2026-08-08
fix(wix-manage): three Catalog V3 frictions in find-products and update-product

danshw:
Three Catalog V3 frictions found in recorded runs, across the two recipes an agent uses together to change something on an existing product. They are in one PR because they compound: an agent answering "set the SKU on the Large size" reads Find Products to locate the product and Update Product to change it, and each defect below costs it a rejected call or a wasted round trip on that single task.

## 1. Price questions were routed into a full-catalog scan

`Query Products` cannot filter on price. The recipe listed its filterable fields and said nothing further, so agents paged the whole catalog and compared amounts in their own code.

Price **is** filterable — on Search Products, via `actualPriceRange.minValue.amount`, which accepts `$lt`, `$gt`, `$lte`, `$gte` and the rest. STEP 1 now shows an executable filter, the routing table gains a price row, and STEP 4 says explicitly that price is not in Query Products' set.

Worth recording: adding the routing-table row **alone did not change behaviour** — the agent kept scanning client-side after reading the recipe, twice. Only making the example executable made it stable.

## 2. The response envelope was never shown

Neither Search nor Query documented where results arrive. Agents fell back to a Catalog V1-shaped guess and lost a round trip:

```
the first API execution failed to locate the product because it incorrectly parsed the
search response using Catalog V1/V2 structure (productResults.results) instead of V3
(products), requiring a second corrected API execution
```

STEP 1 now shows the envelope. Prose competes with a strong prior weakly; an example does not compete at all.

Also corrects the `fields` enum `MIN_VARIANT_PRICE_INFO` → `MIN_PRICE_VARIANT`. The old name does not exist in the API.

## 3. Variant updates dropped state, twice over

**Each variant object is replaced, not merged.** The recipe said to pass the entire array and include each existing `id` — both of which an agent can satisfy while rebuilding each variant from just its `id` plus the field it is changing. Everything else goes with it:

```
400 product is invalid: variantsInfo is invalid: variants [at index 0] is invalid:
    price must not be empty
    (REQUIRED_FIELD on product.variantsInfo.variants[0].price)
```

The error table made this harder to spot — it attributed the error to "new" variants, so an agent updating an existing one read it as not applying.

**Variant choice names are opt-in on read.** The recipe tells you to write `optionChoiceNames`, so agents expect to read them back. Get Product returns `optionChoiceIds` only, and populates names just when the request's `fields` includes `"VARIANT_OPTION_CHOICE_NAMES"`. Matching on a name the response never carried raises nothing — it selects no variant. In a recorded run the agent spent two extra calls inspecting the product's options before resolving the choice GUID.

## Coverage

Two scenarios, three assertions each with the path judge gating at 7, per this repo's eval-scenario guide:

- `stores/find-products-under-price-catalog-v3`
- `stores/update-product-sku-full-variant` — seeds a two-variant product so the agent must send both variants complete; a single-variant fixture let a run pass without exercising the rule at all.

## What is and isn't proven

The price-filter change earned `required` on a decision assertion, path judge 9/10 against 2/10 on production.

The variant changes are supported but not certified. In a paired run where both arms read the recipe, production reproduced the choice-name defect and the PR arm did not — the change doing its job — but the PR arm still fell short of a clean path because of defect 2, which is why these are now in one PR rather than two. Earlier runs also confirmed defect 3's rejection appears without the fix.

These recipes carry more residual friction than any single change removes, so a scenario here can fail its path judge for a reason unrelated to the change under test. Each claim above is tied to a specific recorded run rather than to the verdict.

### DIFF
--- a/skills/wix-manage/references/stores/find-products-query-and-search-catalog-v3.md
+++ b/skills/wix-manage/references/stores/find-products-query-and-search-catalog-v3.md
@@ -18,12 +18,12 @@ Use **Search Products** for text search and name-based lookup. Use **Query Produ
 | Find products by name or free text | [Search Products](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-v3/products-v3/search-products) | Best for user-provided names, keywords, and broad product lookup. |
 | List all products or page through the catalog | [Query Products](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-v3/products-v3/query-products) | Supports paging and structured filters on the fields listed below. |
 | Filter by `id`, `slug`, `handle`, dates, or `visible` | [Query Products](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-v3/products-v3/query-products) | Best for exact structured criteria. |
-| Filter by price — "products under $20", "between $10 and $50" | [Search Products](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-v3/products-v3/search-products) | Query Products has no price filter. Search Products filters on `actualPriceRange.minValue.amount` with `$lt`/`$gt`, which is also where a product's price is read back from. |
+| Filter by price — "products under $20", "between $10 and $50" | [Search Products](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-v3/products-v3/search-products) | Query Products has no price filter, so filter server-side here (STEP 1) instead of paging the catalog and comparing amounts yourself. |
 | Need exact name matching after text lookup | [Search Products](https://dev.wix.com/docs/api-reference/business-solutions/stores/catalog-v3/products-v3/search-products) + client-side match | Search by the name text, then match the returned `product.name` in your own code. |
 
-### STEP 1: Search products by name or free text
+### STEP 1: Search products by name, free text, or price
 
-Use Search Products when the user gives a product name, keyword, or other text expression:
+Use Search Products when the user gives a product name, keyword, or other text expression, and for price criteria, which Query Products cannot filter on. Free text goes in `search.search`; price and other structured criteria go in `search.filter`. Send only the part you need:
 
 ```bash
 curl -X POST 'https://www.wixapis.com/stores/v3/products/search' \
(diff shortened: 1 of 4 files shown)


danshw:
Superseded — this described an earlier revision of this PR, before the Update Product fixes were folded in from a separate branch. See the eval comparison comment below for the current results.

danshw:
## Eval comparison

Both scenarios added by this PR reach a clean path against production.

| Scenario | | PR | production |
|---|---|---|---|
| `stores/update-product-sku-full-variant` | steps | **4** | 8 |
| | correctness | 10/10 | 10/10 |
| | path judge | **10/10 — clean path** | **2/10** |
| | verdict | **required** (PR wins, high confidence) | |
| `stores/find-products-under-price-catalog-v3` | steps | **3** | 6 |
| | correctness | 10/10 | 10/10 |
| | path judge | **10/10 — clean path** | **2/10** |
| | verdict | **required** (PR wins, high confidence) | |

**What production did instead.** On the SKU task: *"Tried wrong catalog version (V1 instead of V3) first resulting in a 428 error, and had to perform an extra API spec lookup to find the SKU field name in the V3 schema."* On the price task: *"fetched products without a price filter and performed client-side comparison, and guessed the price field location (`priceData.price`) resulting in empty matches."*

**Why the changes ship together.** The SKU scenario was run four times while these fixes sat in separate branches, and its PR arm was blocked by a different friction each time — the `price must not be empty` rejection, then variant choice-name resolution, then parsing the search response with the Catalog V1 shape. Each fix was real, and none of them could show a clean path while the next one was still outstanding. With all three applied the path collapses to four steps and scores 10.

**Prose was not enough for either shape defect.** The price routing note alone left the agent still scanning the catalog client-side (5/10, then 2/10) after reading the recipe; only the executable filter in the example made it stable. The same held for the response envelope. Both are now shown as examples rather than described.

Three further scenarios covering these recipes were pulled in and returned `not-required` — two ties and one where both arms failed on unrelated infrastructure. That is expected for scenarios this change does not target.

Single run per arm. Both verdicts rest on decision assertions rather than efficiency ratios.

PR #930, opened 2026-08-09
fix(wix-manage): give each plan type a copyable billingTerms object

danshw:
Gives each plan type a copyable `billingTerms` object, and adds the row that was missing entirely: a recurring plan that ends after a set number of charges.

Before, the plan-type table described `billingTerms` in prose and covered free, one-time and recurring-until-cancelled. "$50 a month for 3 months" — a recurring plan with a fixed duration — had no row, so an agent had to work out from two adjacent rows that it needs `billingCycle` for the recurrence *and* `cyclesCompletedDetails.billingCycleCount` for the number of charges.

## Scope

`+5/−4` in the plan-type table, plus a covering scenario. The required-fields table above it is unchanged; it already documents `plan.status` and the caller-generated `plan.pricingVariants[].id` correctly.

## What the eval shows — and does not

**This is not proven, and the original justification did not survive.** The case for it was one run with the old table scoring 8/10 on the path judge against one with the new table at 10/10. A later paired run had production, on the **old** table, take a clean three-step path and score 10/10. So "the old table causes schema probing" does not reproduce, and there is no measurement showing this change alters agent behaviour.

What it still is: coverage of a plan type users ask for that the table did not describe. That is worth having on its own terms, and it is being landed as documentation completeness rather than on eval evidence.

A change to this recipe's front-matter description was also tried, on the theory that the entry did not signal limited-duration plans and that was suppressing retrieval. Its only test contradicted it — the arm carrying the change did not open the recipe at all — so it has been reverted rather than shipped on an unsupported hypothesis.

## The reporting gap behind this scenario

Worth recording for whoever owns the Plans V3 schema, because no recipe change reaches it.

The friction that originally flagged this scenario was a `400`:

```
plan is invalid:
|-- status value is required
`-- pricingVariants [at index 0] is invalid:
    |-- id is not a valid GUID
    `-- id must not be empty
```

Both fields are misreported in the generated Create Plan reference:

- `plan.status` does not appear in it at all — not in the request schema, not in the required list, not in the response, not in the documented errors — yet the server rejects a create without it.
- `plan.pricingVariants[].id` is listed as required but described only as "Pricing variant GUID". Nothing indicates the caller must generate it, and every other GUID in that schema is server-assigned.

An agent working from the generated reference cannot get either right. In one recorded run the agent read that reference and still failed on the variant GUID. The recipe documents both correctly, but only reaches agents that open it.

### DIFF
--- a/skills/wix-manage/references/pricing-plans/create-and-update-pricing-plans.md
+++ b/skills/wix-manage/references/pricing-plans/create-and-update-pricing-plans.md
@@ -1,6 +1,6 @@
 ---
 name: "Create and Update Pricing Plans"
-description: Creates subscription and one-time payment plans using Plans API. Covers pricing models (recurring, one-time, free), trial periods, perks configuration, and plan visibility.
+description: Creates subscription and one-time payment plans using Plans API. Covers pricing models (recurring until cancelled, recurring for a set number of payments, one-time, free), trial periods, perks configuration, and plan visibility.
 ---
 # Technical Step-by-Step Instructions: Creating or Updating a Wix Pricing Plans (Real-World, API-First)
 
--- a/skills/wix-manage/references/pricing-plans/create-and-update-pricing-plans.md
+++ b/skills/wix-manage/references/pricing-plans/create-and-update-pricing-plans.md
@@ -89,11 +89,12 @@ rejects that combination with `400 FREE_PRICING_VARIANT_IS_NOT_RECURRING`.
 
 Same required fields; only `billingTerms` and the `flatRate.amount` change.
 
-| Plan type | `billingTerms` |
+| Plan type | `billingTerms` — copy as-is; `period` is `DAY`, `WEEK`, `MONTH` or `YEAR` |
 | --- | --- |
-| Free, open-ended | Omit `billingCycle`; `endType: "UNTIL_CANCELLED"`. `flatRate.amount: "0"`. |
-| One-time payment, fixed duration | `billingCycle` = the duration (e.g. `{ period: 'MONTH', count: 1 }`), `endType: "CYCLES_COMPLETED"`, `cyclesCompletedDetails.billingCycleCount: 1`. |
-| Recurring, until cancelled | `billingCycle` = the recurrence (e.g. `{ period: 'MONTH', count: 1 }`), `endType: "UNTIL_CANCELLED"`. |
+| Free, open-ended | `{ startType: 'ON_PURCHASE', endType: 'UNTIL_CANCELLED' }` — no `billingCycle`; `flatRate.amount: '0'`. |
+| One-time payment, fixed duration | `{ startType: 'ON_PURCHASE', endType: 'CYCLES_COMPLETED', billingCycle: { period: 'MONTH', count: 1 }, cyclesCompletedDetails: { billingCycleCount: 1 } }` — charges once; `billingCycle` is the duration the plan lasts. |
+| Recurring, until cancelled | `{ startType: 'ON_PURCHASE', endType: 'UNTIL_CANCELLED', billingCycle: { period: 'MONTH', count: 1 } }` — charges every month until cancelled. |
+| Recurring, ends after N charges | `{ startType: 'ON_PURCHASE', endType: 'CYCLES_COMPLETED', billingCycle: { period: 'MONTH', count: 1 }, cyclesCompletedDetails: { billingCycleCount: 3 } }` — "$50 a month, ends after 3 months": `billingCycle` is the recurrence, `billingCycleCount` is how many times it charges. |
 
 A `billingCycle` can't be shorter than 7 days, and total plan duration can't exceed 10 years
 (`400 VALID_BILLING_CYCLE` / `400 VALID_PLAN_DURATION`).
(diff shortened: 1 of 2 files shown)


danshw:
## Eval comparison — PR content vs. what is live

Run manually (the repository's Actions are disabled, so no check appears on this PR). Both arms
ran the same model. Two full runs; results of both below, because a single run of this scenario
is not a measurement.

### Run 2

| Scenario | Required | Winner | Time (PR / live) | Cost (PR / live) |
|---|---|---|---|---|
| pricing-plans/create-recurring-plan-limited-duration | ✅ | PR (high) | 32.6s / 65.0s | $0.41 / $1.08 |
| pricing-plans/archive-plan | — | live (high) | 133.3s / 33.4s | $1.69 / $0.41 |
| pricing-plans/create-free-plan | ✅ | live (high) | 0.0s / 51.1s | $0.00 / $0.41 |

`create-recurring-plan-limited-duration`, per assertion:

| Assertion | PR | live |
|---|---|---|
| `ReadFullDocsArticle` coverage | ✅ | ❌ |
| correctness judge (min 7) | ✅ 10/10 | ✅ 10/10 |
| path judge (min 7) | ✅ **10/10** | ❌ **1/10** |

The live arm's path judge reason: *"encountering a 400 error and having to write custom JS code
to resolve nested schema references."* Its tool path was
`SearchWixRESTDocumentation → SearchWixAPISpec ×7 → ExecuteWixAPI → SearchWixAPISpec → ExecuteWixAPI`.
The PR arm's was `WixREADME → ReadFullDocsArticle → ExecuteWixAPI` — one call, no probing, no 400.

### Run 1

| Scenario | Required | PR path judge | live path judge |
|---|---|---|---|
| pricing-plans/create-recurring-plan-limited-duration | — | ❌ 2/10 | ❌ 2/10 |
| pricing-plans/archive-plan | ✅ | ✅ 10/10 | ❌ 0/10 |
| pricing-plans/create-free-plan | ✅ | ✅ 10/10 | ✅ 10/10 |

## How to read this honestly

**The run-2 separation is partly luck, and I am not claiming it as proof.** Across six recorded
executions of this prompt, the agent opened the recipe in only two of them. Whether it does is
decided by its *first* tool call, before any recipe content is loaded — so the PR's content
cannot be what causes one arm to read the recipe and the other not to. In run 1 neither arm read
it, which is why that run separated nothing.

The comparison that does isolate the content change is between two runs that **both** read the
recipe, one on each version of the table:

| Recipe table | Tool path after reading the recipe | Path judge | Time |
|---|---|---|---|
| before this PR | `ReadFullDocsArticle → SearchWixAPISpec ×2 → ExecuteWixAPI` | 8/10 | 39.6s |
| after this PR | `ReadFullDocsArticle → ExecuteWixAPI` | 10/10 | 32.6s |

The two spec probes the earlier run needed were for the `billingTerms` shape — exactly what the
copyable objects in this PR supply. That is one run per side: suggestive, not conclusive.

**Two results here are provisioning noise, not signal.** `create-free-plan` on the PR arm of run
2 returned 0 tokens in 0.0s (a gateway error — nothing ran), so its ✅ is an artefact.
`archive-plan` collapsed on the live arm in run 1 (no tool calls at all, 7.1s) and on the PR arm
in run 2 (no recipe read, 13 spec calls). Neither reflects this change.

## The residual, which this PR does not fix

Roughly two thirds of the recorded runs of this prompt never open the recipe at all, and every
one of those takes the same 400:

```
plan is invalid:
|-- status value is required
`-- pricingVariants [at index 0] is invalid:
    |-- id is not a valid GUID
    `-- id must not be empty
```

The cause is upstream of this repository: `status` does not appear anywhere in the published
Plans V3 `Plan` schema — not in the entity's properties, not in the Create Plan request body's
`required` list, and not in the documented response shape — while the server rejects a create
without it. An agent that works from the generated schema therefore cannot get this right, and
no recipe can reach an agent that does not open it. That belongs in the API definition for the
plan entity, and is being routed there separately.

danshw:
## Eval comparison — not proven

Latest paired run on `pricing-plans/create-recurring-plan-limited-duration`:

| | PR | production |
|---|---|---|
| steps | 5 | 3 |
| recipe read | no — read the generated Create Plan reference instead | yes |
| correctness | 10/10 | 10/10 |
| path judge | 6/10 | 10/10 (clean path) |

Production wins. Two things follow, and neither supports this PR as it stands.

**The description change is unsupported.** It was added to test whether naming limited-duration plans would make the recipe easier to find for this request. On this run the PR arm did not open the recipe at all. That is a single observation against a baseline where this request opens the recipe in roughly a third of runs, so it is weak evidence — but it is not evidence in favour, and the change should not ship on a hypothesis this run failed to support.

**The billing-terms table change is also unproven.** The case for it rested on one run with the old table scoring 8/10 against one with the new table at 10/10. Here the old table produced a clean 10/10 path, which undercuts that comparison. The missing "recurring, ends after N charges" row is a real gap in coverage of a case users ask for, but there is no measurement showing it changes agent behaviour.

**What the run does confirm** is the underlying reporting gap. The PR arm read the generated Create Plan reference and still failed, with `plan.pricingVariants[0].id` missing — because that reference lists the field as required without stating the caller must generate the GUID, and does not mention `plan.status` at all even though the server rejects a create without it. An agent working from the generated schema cannot get this right, and no recipe reaches an agent that does not open one.

PR #931, opened 2026-08-09
feat(wix-manage): add Create Contact with Address and Update a Contact recipes

danshw:
Adds two contacts recipes — one for creating a contact with a physical address, one for updating an existing contact — and puts the values the API actually accepts in copyable bodies.

## Defect 1 — the documented `subdivision` example is rejected

The Contacts reference documents `subdivision` as:

> Subdivision shorthand. Usually, a short code (2 or 3 letters) that represents a state, region, prefecture, or province. For example, `NY`.

The field is validated as ISO 3166-2 — country code, hyphen, then 1-3 characters — so the documented example is rejected:

```
HTTP 400 {"message":"contact is invalid:
`-- addresses [at index 0] is invalid:
    `-- address is invalid:
       `-- subdivision is not a valid subdivision code",
 "details":{"validationError":{"fieldViolations":[{
   "field":"contact.addresses[0].address.subdivision",
   "description":"is not a valid subdivision code","violatedRule":"FORMAT",
   "data":{"type":"SUBDIVISION"}}]}}}
```

In a recorded run an agent sent `subdivision: "NY"`, took that 400, and retried with `"US-NY"`, which succeeded. It had not guessed — it used the value the reference gives.

The same field is validated the same way on the update path, where the rejection names a different field path:

```
HTTP 400 {"message":"address is invalid:
`-- address is invalid:
    `-- subdivision is not a valid subdivision code",
 "details":{"validationError":{"fieldViolations":[{
   "field":"address.address.subdivision",
   "description":"is not a valid subdivision code","violatedRule":"FORMAT",
   "data":{"type":"SUBDIVISION"}}]}}}
```

`country` is unaffected: it stays the plain 2-letter code, `US`, with no prefix. Only `subdivision` carries the country.

## Defect 2 — a contact cannot be found by name with Query Contacts

Every "update my contact <name>" request has to resolve a name to an id first. Query Contacts filters on a small closed set that does not include the contact's name, so the obvious first attempt is rejected:

```
HTTP 400 {"message":"value Field 'name.first' is not declared as filterable",
 "details":{"validationError":{"fieldViolations":[{"field":"value",
   "description":"Field 'name.first' is not declared as filterable"}]}}}
```

Search Contacts takes a free-text expression and matches names. Recorded runs confirm it: `POST /contacts/v5/contacts/search` with `{"search":{"search":{"expression":"Jordan Lee"}}}` returned `200` with the matching contact. Neither method's own page tells you to make that choice, so the update recipe makes it step 1.

[...]

PR #942, opened 2026-08-10
feat(wix-manage): add Wix Events recipes

danshw:
`wix-manage` had **no Wix Events coverage at all**, so every events request was answered by probing the API spec from scratch. This adds two recipes and registers them.

Every request body and every error quoted below was executed against the live Wix Events API on a scratch site; nothing is transcribed from the spec alone. That mattered — the first ticket bodies I wrote from the spec were wrong three ways (`feeType` is required, `pricingMethod.free` is read-only, the quantity field is `initialLimit`), and would have shipped as new defects.

## Why the friction is high on this surface

Several of the API's validation messages point *away* from the fix:

| Sent | Response | Why it misleads |
| --- | --- | --- |
| no `timeZoneId` | `400 getTimeZoneId is not supported` | Reads as "remove this field". It means the field is **missing**. |
| `initialType` omitted | `400 initialType is unexpected. Expected: [RSVP,TICKETS,RSVP_AND_TICKETS]` | `TICKETS` is **rejected** by the API. The accepted value is `TICKETING`. |
| `feeType: "FEE_AT_CHECKOUT"` | `400 feeType value is required` | An invalid enum value reports as a *missing* one — for `initialType` too. |
| `locationTbd: true` alone | `400 Location address must not be a blank` | A `name` is enough to satisfy it. |

Three more fail **silently**, which is worse than a 400:

- a `limit` set on `registration` instead of `registration.rsvp` is ignored with no error;
- a clone resets its start date to ~14 days out and returns as a `DRAFT`;
- Query Events excludes drafts from both results and `pagingMetadata.total`.

## Dates are ISO-8601 strings

The spec types `startDate`/`endDate` as `string` / `format: date-time` but *also* carries a `$circular` pointer to `google.protobuf.Timestamp`, whose fields are `seconds` and `nanos`. Sending that shape fails with `400 Expected a string.` Both recipes state the rule, and flag that it applies to every date-time field across the Wix APIs — not just Events.

## Contents

- **Create an Event** — required body, date/time rules, venue/online/TBD location and street addresses, RSVP vs ticketing, guest capacity, `shortDescription` (plain string) vs `description` (Ricos object). Also disambiguates the **five** unrelated Wix APIs called "Events" (Wix Events / Calendar / Marketing Calendar / Triggered / BI), which is what routes "recurring event" requests into Bookings.
- **Manage Wix Events** — ticket definitions and pricing (fixed, free, donation, multiple tiers), publish, cancel, delete, clone, update, counting, and recurring series from explicit occurrence dates (there is no `RRULE`).

Both are under the 10k serve limit, so neither is truncated in use.

## Testing

Walked all 14 `coverage/events-*` scenario prompts against the live API. Everything created was deleted and the Events app uninstalled; the scratch site is back to its prior state.

No eval comparison has been run yet, so the friction impact is **unmeasured** — the claims here are verified against the API, not against a score.

### DIFF
--- a/skills/wix-manage/SKILL.md
+++ b/skills/wix-manage/SKILL.md
@@ -216,10 +216,10 @@ These recipes do NOT cover frontend development or SDK usage for displaying data
 
 ## Events
 
-### [Create an Event](references/events/create-wix-event.md)
+### [Create Event](references/events/create-wix-event.md)
 **Technical:** Creates an event with the Wix Events V3 API — required request body, ISO-8601 date and time settings, venue/online/TBD location and street addresses, RSVP vs ticketed registration, guest capacity, and short vs Ricos rich-text descriptions. Distinguishes Wix Events from the Calendar, Marketing Calendar and Automations APIs that share the "events" name. Key endpoint: /events/v3/events.
 
-### [Manage Wix Events](references/events/manage-wix-events.md)
+### [Manage Events](references/events/manage-wix-events.md)
 **Technical:** Manages existing events with the Wix Events V3 API — ticket definitions and pricing (fixed, free, donation, multiple tiers), publishing a draft, cancelling, deleting, cloning, updating an event's date, counting events, and building recurring series from explicit occurrence dates. Key endpoints: /events/v3/events, /events/v3/ticket-definitions.
 
 ---
--- a/skills/wix-manage/SKILL.md
+++ b/skills/wix-manage/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: wix-manage
-description: "Wix business solution management recipes — REST API operations for configuring and managing Wix business solutions. Routes to: stores, bookings, get-paid, CMS, contacts, forms, media, app-installation, pricing-plans, restaurants, rich-content, sites, blog, calendar, domains, site-properties, ecommerce, marketing, google-ads, analytics, dashboard-navigation."
+description: "Wix business solution management recipes — REST API operations for configuring and managing Wix business solutions. Routes to: stores, bookings, get-paid, CMS, contacts, forms, media, app-installation, pricing-plans, restaurants, rich-content, sites, blog, calendar, domains, events, site-properties, ecommerce, marketing, google-ads, analytics, dashboard-navigation."
 compatibility: Requires Wix REST API access (API key or OAuth).
 ---
 
@@ -211,6 +211,16 @@ These recipes do NOT cover frontend development or SDK usage for displaying data
 </details>
 
 > Dashboard links for eCommerce surfaces (orders, abandoned checkouts, gift cards, shipping, tax, checkout settings) are in [Stores Dashboard Navigation](references/stores/stores-dashboard-navigation.md).
+
+---
+
+## Events
+
+### [Create an Event](references/events/create-wix-event.md)
+**Technical:** Creates an event with the Wix Events V3 API — required request body, ISO-8601 date and time settings, venue/online/TBD location and street addresses, RSVP vs ticketed registration, guest capacity, and short vs Ricos rich-text descriptions. Distinguishes Wix Events from the Calendar, Marketing Calendar and Automations APIs that share the "events" name. Key endpoint: /events/v3/events.
+
+### [Manage Wix Events](references/events/manage-wix-events.md)
+**Technical:** Manages existing events with the Wix Events V3 API — ticket definitions and pricing (fixed, free, donation, multiple tiers), publishing a draft, cancelling, deleting, cloning, updating an event's date, counting events, and building recurring series from explicit occurrence dates. Key endpoints: /events/v3/events, /events/v3/ticket-definitions.
 
 ---
 
--- a/skills/wix-manage/references/events/create-wix-event.md
+++ b/skills/wix-manage/references/events/create-wix-event.md
@@ -1,0 +1,175 @@
+---
+name: "RECIPE: Create an Event with the Wix Events API"
+description: "Creates an event with the Wix Events V3 API — the required request body, ISO-8601 date and time settings, venue/online/TBD location, RSVP vs ticketed registration, guest capacity, and short vs rich-text descriptions. Covers the exact field shapes and the API's misleading validation messages. Use when the user wants to create an event, set its date, location, description or guest limit, or choose between RSVP and ticketing."
+---
+
+# RECIPE: Create an Event with the Wix Events API
+
+## Goal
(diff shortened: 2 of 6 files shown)


danshw:
**Moving this to draft — measured, and it makes things worse.** Do not merge as it stands.

Ran the 14 `coverage/events-*` scenarios against this branch vs production (1 repetition, 28 scenario-executions). Friction, PR vs prod:

| scenario | PR | prod | Δ |
|---|---|---|---|
| events-count | 4 | 10 | **-6** |
| events-publish-event | 3 | 7 | **-4** |
| events-set-description | 6 | 10 | **-4** |
| events-update-event | 6 | 10 | **-4** |
| events-create-event | 4 | 7 | -3 |
| events-cancel-event | 6 | 9 | -3 |
| events-set-location | 7 | 10 | -3 |
| events-clone | 5 | 7 | -2 |
| events-delete | 8 | 10 | -2 |
| events-rsvp-setup | 6 | 8 | -2 |
| events-set-capacity | 5 | 7 | -2 |
| events-multiple-ticket-tiers | 5 | 6 | -1 |
| events-recurring | 7 | 8 | -1 |
| events-ticket-setup | 5 | 5 | 0 |
| **total** | **77** | **114** | **-37** |

Worse or flat on all 14. Not noise.

## Cause

**`ReadFullDocsArticle` 404s on the new recipe — in 14/14 PR-side runs and 0/14 production runs.**

The agent finds the recipe in the index, tries to fetch it, gets a 404, then falls back to raw spec probing — so it pays a failed call and ends up exactly where it started. A recipe that cannot be fetched is worse than no recipe at all, and it degrades even the scenarios that were previously clean at 10/10 (`events-count`, `set-description`, `set-location`, `update-event`).

Note the friction judge is reading these 404s as *docs* defects, so the reasons blame "outdated/broken doc URLs" — but production shows zero, so the URL is broken *because the page is new*, not because the docs are stale.

## Not yet root-caused

My first hypothesis — that the `documentation.yaml` title must slugify to the recipe filename — is **wrong**: 38 existing recipes violate that and work fine.

The likelier explanation is that `ReadFullDocsArticle` reads the published docs portal, which has no page for a recipe that exists only on a branch. If so, this is a cold-start problem affecting **every new recipe**, not something specific to this PR — and it would need confirming before any new-recipe PR can be measured this way at all.

## What still stands

The API findings are unaffected — they were verified directly against the live Events API, not inferred from a score: the ISO-8601 vs `{seconds,nanos}` rejection, `timeZoneId` reported as "not supported" when it is missing, `TICKETS` vs `TICKETING`, the location TBD rules, `registration.rsvp.limit` being silently ignored at the wrong nesting, the clone date reset, and drafts being excluded from Query Events totals.

The content is right; the delivery mechanism is not. Parking until the 404 is understood.

danshw:
## Root-cause update on the 404

Narrowed, and the manifest is fixed — but **not yet proven**, so this stays in draft.

**The recipe article URL is `<docsEntry>/skills/<title-slug>`.** Confirmed live against an existing recipe: `docsEntry` for contacts is `.../crm/members-contacts/contacts`, title *"Bulk Delete Contacts"*, and `.../contacts/skills/bulk-delete-contacts` resolves.

This vertical had `docsEntry` pointing at **`business-solutions/events/event-management/events-v3` — a method resource, not a documentation category.** Every other vertical points at a category node. Depth is not the rule (`google-ads` is four segments deep and works); category-vs-resource is.

So the generated URL went:

```
before  .../events/event-management/events-v3/skills/create-an-event-wix-events-v3
after   .../business-solutions/events/skills/create-an-event
```

`.../business-solutions/events` is confirmed to be a real category. Titles are shortened too, since the title becomes the slug and the parenthesised version suffix produced an awkward one.

## Hypotheses ruled out along the way

- **"The title must slugify to the filename."** No — 38 existing recipes violate that and work fine.
- **"New recipes always 404 because they aren't published yet."** Not conclusive. PR #931 added new recipes with zero 404s, which argues against it — though I'd treat #931 as a weak control, since its with/without scores were *identical* on both scenarios, which may mean its branch content wasn't being served either.

## Status

The manifest change is right regardless of whether it is the whole story — it aligns this vertical with all 21 others. Whether it clears the 404 needs a re-measurement, which I have not run.

The API findings in this PR are unaffected either way: they were verified directly against the live Events API, not inferred from a score.

danshw:
## Measured: 138/140, and the 404 is gone

Ran all 14 `coverage/events-*` scenarios against this branch (PR side only, 14 scenario-executions). Thirteen scored **10/10 "clean path"**; one scored 8.

| | friction total, 14 scenarios |
|---|---|
| **this PR, now** | **138** |
| production (same-run control, earlier) | 114 |
| production (`aria-sweep-2026-08-09`) | 116 |
| this PR *before* the `docsEntry` fix | 77 |

**`ReadFullDocsArticle` 404s: 0/14**, down from 14/14. That was the whole cause of the earlier regression, and it is resolved.

Every scenario is now at or above its production score. Six of the eight that had regressed are perfect: `count` 4→10, `create-event` 4→10, `set-description` 6→10, `set-location` 7→10, `update-event` 6→10, `set-capacity` 5→10. Runtimes fell from a 77–210s spread to 31–64s.

### Reading it honestly

- **No control arm in this run**, so 138-vs-114/116 is a cross-run comparison. Mitigating it: the two independent production baselines agree (114, 116), and a +22 gap across 14 scenarios sits far outside the ±2 per-scenario variance measured directly between runs.
- **n=1 per scenario.** The 404 elimination is not sample-dependent; the exact total is.

### The one scenario below 10

`events-publish-event` scored 8: *"had to read two separate documentation articles… and write a custom JS function to chain them."* That is the cost of splitting into two files, which the 10k serve limit forced — a single file would be truncated, losing its tail silently. Only that one scenario of 14 spans both recipes. The creation recipe now names the publish endpoint inline, which should remove the round trip.

## Also in this PR

Eval scenarios, per CONTRIBUTING — these were missing and would have failed the automated check:

- `events/create-event-with-capacity-and-description`
- `events/ticket-tiers-and-publish`

Each carries the required three assertions (a `ReadFullDocsArticle` coverage assertion on its own doc URL, an `llm_judge` on the outcome, an `llm_judge` on the tool-call path) and runs on the `events-editor` template, since every `/events/v3` call against a site without the Events app returns 428.

Three self-review corrections to the recipes:

- `count-by-status` — the previous wording asserted that filter criteria were the cause of the empty `facets`, which was never tested. It now states only what was observed.
- publish/cancel/delete — only publish and cancel take an empty body.
- the publish endpoint is named in the creation recipe, so create-a-draft-then-publish no longer needs both files.

PR #619, opened 2026-07-17
🦞 igor-claw: document existing-site editor API gap (pages/menus/homepage)

ayal:
## Summary

Reported feedback: a user with a live premium classic-editor site had a ready homepage design (hero, menu, video embed, gallery) and no way to deliver it — the agent spent multiple sessions hunting for an API to add pages, link a menu item, or change homepage layout on the user's *existing* site, and found none. Confirmed independently in this investigation via a full pass over `BrowseWixRESTDocsMenu`'s top-level categories (Articles, App Management, Business Solutions, Assets, CRM, Business Management, Account Level, Tools, Site, Wix Backoffice) plus targeted `SearchWixRESTDocumentation` queries for page/menu/navigation/homepage APIs.

This is a genuine product gap (classic Editor/Studio page structure lives in the Editor's internal document model, with no public write API) — not something fixable in this repo. What *is* fixable here: nothing in `WixREADME`'s recipe index told an agent this upfront, so it burned many tool calls (and, per the report, multiple days) discovering the gap by exhaustive failure instead of being told immediately.

## Changes

- New recipe: `skills/wix-manage/references/sites/editing-existing-site-pages-menus-and-homepage.md` — states the gap plainly, lists the near-miss APIs that don't actually apply (Add Store Pages to Site, app-management site-plugin slots, Get Editor URLs) and why, and lays out the only real options (manual desktop-browser Editor work, or build-new-site-and-reconnect-domain via `WixSiteBuilder`/`import-claude-design-from-url`) — including the note that mobile Editor/Wix Owner app cannot do structural homepage edits.
- Registered in `skills/wix-manage/SKILL.md` under **Sites**, and in `yaml/wix-manage/sites/documentation.yaml`, following the existing registration pattern for that vertical.

Opened automatically by an AI agent on behalf of Ayal (`ayalg@wix.com`).
Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1784307665700339

## Test plan
- [ ] `skills/wix-manage/SKILL.md` surfaces the new entry under **Sites**
- [ ] `yaml/wix-manage/sites/documentation.yaml` entry resolves correctly for the EvalForge YAML gate

PR #956, opened 2026-08-10
🦞 igor-claw: add gap recipe for per-page HTML embed elements vs. Custom Embeds API

ayal:
## Gap

A Wix MCP user tried to programmatically manage the HTML content of a per-page **Add Elements > Embed > HTML iframe** component (8 pages, scaling to 24 across languages) and reached for the **Custom Embeds API**, since its description ("manages custom HTML and JavaScript code embedded in a site, along with its placement and properties") reads as if it could cover this. It can't: `position` is only `HEAD`/`BODY_START`/`BODY_END` — sitewide document-level injection, not a page-canvas visual component.

There's genuinely no public API for the per-page visual embed case:
- Editor SDK's [Element Properties](https://dev.wix.com/docs/sdk/host-modules/editor/elements/element/element-properties) only supports `text`/`image` properties, no `html`.
- Velo's [`$w` `HtmlComponent`](https://dev.wix.com/docs/velo/velo-only-apis/$w/html-component/introduction) only exposes `src` (a URL) and `postMessage`/`onMessage` — no settable `.html` string.

## Fix

Adds `skills/wix-manage/references/sites/custom-code-and-page-embeds.md`, wired into `SKILL.md` and `yaml/wix-manage/sites/documentation.yaml`, so agents:
1. Don't waste time trying to force the Custom Embeds API into this use case.
2. Tell the user directly that there's no API for it.
3. Suggest the one real workaround: point each embed's `src` at one externally-hosted HTML resource instead of pasting raw HTML per instance, so future content changes are a single edit instead of N re-pastes.

Companion docs fix in the owning service repo: wix-private/editor-server#15107 (adds a scope-clarifying callout to the Custom Embeds API intro article).

---

This PR was opened automatically by an AI agent on behalf of Ayal (`ayalg@wix.com`), researching feedback from a Wix MCP user. Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1786400308564839

### DIFF
--- a/skills/wix-manage/SKILL.md
+++ b/skills/wix-manage/SKILL.md
@@ -369,6 +369,9 @@ These recipes do NOT cover frontend development or SDK usage for displaying data
 ### [Sites Dashboard Navigation](references/sites/sites-dashboard-navigation.md)
 **Technical:** Direct links to the account-level My Sites list (manage.wix.com/account/websites) and per-site dashboard homes, paired with the Query Sites read API.
 
+### [Custom Code & Embeds — Sitewide Scripts vs. Per-Page Embed Elements](references/sites/custom-code-and-page-embeds.md)
+**Technical:** Clarifies that the Custom Embeds API (`/embeds/v1/custom-embeds`) only injects sitewide HTML/JS at `HEAD`/`BODY_START`/`BODY_END` — there is no API to read or write the content of a per-page visual embed element (Editor: Add Elements > Embed > HTML iframe). Use before reaching for the Custom Embeds API to manage per-page embedded HTML/iframe content, and for the `src`-based workaround when the same embed needs to be kept in sync across many pages/languages.
+
 ---
 
 ## Stores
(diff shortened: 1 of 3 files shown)


ayal:
🤖 Automated note from an AI agent researching Wix MCP feedback on behalf of Ayal (`ayalg@wix.com`) — not a human comment.

First corroboration of this recipe's gap from a different angle than the original report: a fresh feedback thread (https://wix.slack.com/archives/C08P5DKLJR5/p1787216657479649) describes a user trying to get an approved, self-contained webpage design onto their existing Wix Editor site as a native page. With no native import path (see wix-private/igor-tools#1472 and wix/skills#619, corroborated separately), they were forced into "a full-page external iframe" as a workaround to preserve the exact design — precisely the brittle `src`-hosted-iframe workaround this recipe documents as the only lever available, since Custom Embeds and the Editor SDK/Velo `$w` HTML component have no settable-HTML per-page API. The report calls the result out explicitly as a bad experience (~5-6 hours of delay, "brittle").

Still open/unmerged (~10 days), 0 comments before this one, reviewer itayhewix requested since day one — flagging for review given this is now hitting real users beyond the original at-scale multi-page report.

PR #968, opened 2026-08-12
fix(wix-manage): stop the events recipes steering agents into draft events

danshw:
The two slowest scenarios in the `aria-agent-2026-08-11` sweep were both events, and the traces showed **the recipes causing it**. Both were introduced by #942.

| scenario | duration | cause |
|---|---|---|
| `events-multiple-ticket-tiers` | **143s**, friction 4 | `403 READ_DRAFT_EVENTS` × 4 |
| `events-ticket-setup` | 85s, friction 5 | `403 READ_DRAFT_EVENTS` × 3 |
| `events-recurring` | 96s, friction 5 | 500s: invalid `24:00`, missing per-occurrence `timeZoneId` |

For contrast, what the recipes already served well: `events-create-event` **23s 10/10**, `events-update-event` **36s 10/10**.

## 1. The creation example set a trap

It showed `"draft": true`. The agent copied it, then spent six calls discovering that a draft event cannot have tickets added, be queried, be fetched by slug, or be published without `WIX_EVENTS.READ_DRAFT_EVENTS` — ending, in its own words, at *"the solution is to create the event without the `draft: true` flag."*

The example no longer sends `draft`, and the permission and its consequences are stated where the flag used to be.

## 2. Recurring was in the file the agent doesn't open

`events-recurring` read the **creation** recipe — where "set up a weekly event" lands — but recurring lived in the **management** recipe. It spent four `SearchWixAPISpec` calls rediscovering `individualEventDates` and the per-occurrence `timeZoneId`, then concluded *"Each Occurrence needs… `timeZoneId`"* — the exact line in the file it never read.

## The general lesson

**A cross-reference is not coverage.** Step 3 of the ticket trace says it plainly:

> *"The recipe mentions ticket definitions… but says to see 'Manage Wix Events' for details. I need the ticket definition API details."*

— and then it searched the API spec rather than following the link. So the split now follows **the task**, not the API surface:

- **Create an Event** — body, dates, location, registration, capacity, descriptions, **ticket tiers**, **recurring occurrences**
- **Manage Wix Events** — publish, cancel, delete, clone, update, count

Both stay under the 10k serve limit (9,566 / 5,037), the constraint that forced the original split. Also documents that `24:00` is not a valid ISO time — the recurring run's first 500.

## Measured by the gate

| `events/ticket-tiers` | PR | prod |
|---|---|---|
| time | **47.0s** | 93.2s |
| tokens | **205,561** | 360,163 |
| cost | **$0.42** | $0.74 |

Roughly half the time and 43% fewer tokens — the pairwise judge's own words: *"both complete the task, but PR is at least 30% more efficient."* `events/create-event-with-capacity-and-description` passed 3/0 on both sides.

## Scenarios

Each now routes **directly** to the recipe it covers, with no hop:

| scenario | covers |
|---|---|
| `events/create-event-with-capacity-and-description` | create |
| `events/ticket-tiers` | create — tickets live there now |
| `events/count-events` | manage |

`events/ticket-tiers-and-publish` is renamed, because publishing a draft was the permission trap itself.

**`maxTokens` raised 30000 → 300000.** The gate compares it against the run's `totalTokens`, which sums prompt tokens across every step — the system prompt alone is ~34.5k on the first one, so 30000 was unreachable for any agentic scenario. These used 205k/206k/213k. It is the value 22 other scenarios carry, and it holds there only because the gate checks the budget for scenarios a PR actually touches.

## Known gap, not fixed here

`events/count-events` passes both judges 10/10 but **fails its coverage assertion** — counting is simple enough that the agent answers correctly without opening the manage recipe. A task that genuinely needs it (cloning, where the date resetting to ~14 days out is documented only there) would cover it properly. Worth doing, but it is a scenario redesign rather than part of this fix.

## Testing

Every flow executed against the live Wix Events API, all clean:

| flow | result |
|---|---|
| create published + `rsvp.limit` + `shortDescription` | `UPCOMING`, limit 40, description stored |
| ticketed + GA $20 + VIP $60 + cancel | `TICKETING`, `20.00`/`60.00` → `CANCELED` |
| recurring with per-occurrence `timeZoneId` | `RECURRING` |
| count scenario's bootstrap seeds | 2xx; `pagingMetadata.total` = 2; `count-by-status` = `{}` |

Everything created was deleted and the app uninstalled. Recurring occurrences are independent events — deleting the parent leaves them, which the recipe states and cleanup confirmed.


danshw:
## ⚠️ Required cleanup on merge

This PR renames a scenario:

```
events/ticket-tiers-and-publish  →  events/ticket-tiers-and-cancel
```

**Scenario name is the sync key**, so this is a create-plus-orphan, not a rename. After merge, the old scenario stays registered in EvalForge:

| id | name | tags |
|---|---|---|
| `80a3c855-f399-46ed-ac25-766572c2b6c0` | `events/ticket-tiers-and-publish` | `events` |

Left in place it keeps running in the nightly sweep, asserting a publish-a-draft flow that this PR deliberately steers agents away from — so it would fail permanently and read as a regression caused by this change.

**It must be deleted from the EvalForge project once this lands.** (Compare the still-orphaned `draft:wix/skills#255` and `#276` entries, which are what this looks like if it is forgotten.)

The renamed scenario `events/ticket-tiers-and-cancel` will be created by the normal sync and carries the `events` tag, so no promote step is needed.

danshw:
## Scenario split — each one now routes directly to the recipe it covers

The previous shape had the ticket scenario asserting coverage on the **management** recipe while the ticket shapes had moved to the **creation** recipe. That made the assertion depend on the agent hopping between files — the exact hop these traces show it does not make. Split:

| scenario | covers | routing |
|---|---|---|
| `events/create-event-with-capacity-and-description` | create | direct |
| `events/ticket-tiers` | create | direct — tickets now live there |
| `events/count-events` | **manage** | direct — no hop |

`count-events` is the pure management task: *"How many events do I have on my site?"* It exercises what only that recipe carries — reading `pagingMetadata.total` off Query Events rather than `count-by-status`, whose empty `facets` for an empty body read as though the site has none.

Its correctness judge scores **grounding, not state**: pass if the reported number matches what a call actually returned. The site is seeded with two events, but a bootstrap that races provisioning would then fail *visibly* rather than turning an otherwise-correct run red.

### Verified live

| check | result |
|---|---|
| both bootstrap seed bodies | 2xx, `UPCOMING` (published, so countable) |
| `POST /events/v3/events/query` → `pagingMetadata.total` | **2** — matches the seed |
| `POST /events/v3/events/count-by-status` with empty body | **`{}`** — the trap the quality judge penalises |

Everything created was deleted and the app uninstalled.

### Updated cleanup on merge

Still one orphan to remove, now under its final name — `80a3c855-f399-46ed-ac25-766572c2b6c0` (`events/ticket-tiers-and-publish`). The two new names (`events/ticket-tiers`, `events/count-events`) are created by the normal sync and carry the `events` tag, so no promote step is needed.

danshw:
## Correction — the two cleanup comments above are wrong

I wrote twice that the renamed scenario would orphan `events/ticket-tiers-and-publish` and that it must be **deleted manually** on merge. That is not true, and the gate log says so directly:

```
Deferring DELETE of "events/ticket-tiers-and-publish" — will be handled at PR merge
```

The gate already tracks renames and defers the delete to merge. **No manual cleanup is needed.** I inferred the orphan risk from the stale `draft:wix/skills#255`/`#276` entries without checking whether the gate handles it — those are leftovers from a different path, not evidence about renames.

Also superseded in those comments:

- The scenario is now **`events/ticket-tiers`**, not `ticket-tiers-and-cancel` — cancelling moved out when the scenarios were split to route directly.
- `events/count-events` was added to cover the management recipe.

The PR description is up to date; read it rather than the older comments.

danshw:
## `count-events` replaced — it couldn't cover what it was for

It passed both judges 10/10 and **failed its coverage assertion on both sides, every run**. Counting is simple enough that the agent answers correctly without ever opening the manage recipe. A scenario the agent can satisfy while ignoring the file under test measures nothing.

Replaced with **`events/move-event-date`** — *"Move my Spring Fair event to the 12th of the following month, same time."*

## Why that task has teeth

Verified against the live API before writing it:

| patch | result |
|---|---|
| `startDate` only | **`400 event cannot have negative duration`** |
| with a `revision` | works — ignored, not rejected |
| `startDate`+`endDate`, no `timeZoneId` | works |

`dateAndTimeSettings` is replaced wholesale rather than merged, so patching just the new start — the intuitive minimal update — leaves the event with no end. The error names neither the missing field nor the replacement, so it isn't self-correcting: an agent that skipped the recipe pays a round trip.

## Two corrections to the recipe, from the same probes

The update section claimed the create rules applied unchanged. They don't:

- **`timeZoneId` is optional on update** — required only on create. Previously stated as required.
- **A `revision` is ignored, not rejected** — so fetching one first is wasted work rather than an error. Now stated, since that is exactly the friction `bookings-update-service` shows (*"needing to fetch the service revision via GET"*).

## Coverage

| scenario | covers |
|---|---|
| `events/create-event-with-capacity-and-description` | create |
| `events/ticket-tiers` | create |
| `events/move-event-date` | **manage** |

Correctness is judged on what the trace shows, with an explicit path for a bootstrap that races provisioning — so a seeding failure reads as one rather than as a wrong answer. `maxTokens` is 500000 on all three, above every value seen on either side across both gate runs.

PR #976, opened 2026-08-12
fix(ci): stop the @wix/cli pin check flagging comments

danshw:
## The check is red on every PR, and has been since 2026-08-06

`Check @wix/cli pinned to @latest` fails on **every PR touching `skills/**`** — currently `igor-claw/site-import-no-support-dead-end`, `docs/dev3-site-import-source-url`, `forms-on-wix-headless` and `fix/events-draft-and-recurring`, none of which changed anything CLI-related.

The one offending line, added by #854 — `skills/wix-replatform/resources/rp-target-wix/lib/wix-writers.js:327`:

```js
// header; OAuth access tokens (e.g. a Wix CLI token from `npx @wix/cli token --site …`)
```

Prose describing what a CLI token is, not a call. Pinning is meaningless for it, so the only way to satisfy the check was to edit a sentence into something it doesn't quite say. A check that fails on every PR for a reason no author can act on is one everybody learns to scroll past.

It went unnoticed because the workflow triggers only on `pull_request`, so it never runs on `main`.

## The fix — one pipe stage

```diff
-  . 2>/dev/null || true)
+  . 2>/dev/null | grep -vE '\.(sh|mjs|js|ts):[0-9]+:[[:space:]]*(//|/\*|\*|#)' || true)
```

Drops matches on comment lines in source files. It keys on the **file extension**, so Markdown keeps every match — `#` opens a heading there, and an unpinned command in docs is one a reader will copy. A comment *trailing* real code still counts, since that code runs.

## How we know it works, and that nothing broke

Nothing under `.github/scripts` had tests, so "it still catches what it used to" was only ever checked by hand by whoever last touched the pattern. This adds `--self-test`, which asserts **both directions** against fixtures in a temp directory:

| fixture | expected |
|---|---|
| unpinned in markdown prose | flagged |
| unpinned in `.sh` | flagged |
| unpinned in `.js` with a comment after it | flagged |
| `//` line comment | ignored |
| `*` block-comment body | ignored |
| `#` shell comment | ignored |
| already `@latest` | ignored |

CI runs it **before** the repo scan, because a green scan proves nothing if the matcher has stopped matching:

```yaml
- run: bash .github/scripts/check-cli-pinned.sh --self-test
- run: bash .github/scripts/check-cli-pinned.sh
```

Run it yourself with `bash .github/scripts/check-cli-pinned.sh --self-test`.

One wrinkle worth knowing: the fixtures are built from a `$cli` variable so this file never contains a literal match. Written the obvious way, the check flags its own test data.

## Not changed

`run("npx @wix/cli build")` is invisible to **both** the old and new versions — the pattern needs whitespace or end-of-line after the subcommand, and `"` is neither. Confirmed by running the pre-change script against the same input. A real pre-existing gap, left alone to keep this diff reviewable; the self-test would make it a one-line fix if you want it closed.

### DIFF
--- a/.github/scripts/check-cli-pinned.sh
+++ b/.github/scripts/check-cli-pinned.sh
@@ -8,16 +8,51 @@
 # the @wix/cli surface grows.
 #
 # Run locally:   bash .github/scripts/check-cli-pinned.sh
+# Self-test:     bash .github/scripts/check-cli-pinned.sh --self-test
 # CI:            invoked by .github/workflows/check-cli-pinned.yml on PRs
 
 set -euo pipefail
 
 SUBCMDS="token|whoami|login|env|build|release|preview|dev"
 
-bare=$(grep -rnE "npx @wix/cli (${SUBCMDS})([[:space:]]|$)" \
-  --include='*.md' --include='*.sh' --include='*.mjs' --include='*.js' --include='*.ts' \
-  --exclude-dir=node_modules --exclude-dir=.git --exclude-dir=dist \
-  . 2>/dev/null || true)
+# The second grep drops matches sitting on a comment line in source files —
+# prose about the CLI, not a call. Markdown keeps every match: `#` opens a
+# heading there, and an unpinned command in docs is one a reader will copy.
+scan() {
+  grep -rnE "npx @wix/cli (${SUBCMDS})([[:space:]]|$)" \
+    --include='*.md' --include='*.sh' --include='*.mjs' --include='*.js' --include='*.ts' \
+    --exclude-dir=node_modules --exclude-dir=.git --exclude-dir=dist \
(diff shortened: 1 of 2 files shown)


danshw:
Closing in favour of #977, which pins the example in the comment instead — a one-word change.

Testing this properly is what killed it. `*` and `#` open a comment in some places and are real code in others, so the filter silently stopped catching real violations:

| line | what it is | result |
|---|---|---|
| `*gen() { run("npx @wix/cli build --x") }` | JS generator method | **missed** |
| `#cmd = "npx @wix/cli build --x";` | TS private field | **missed** |
| markdown quoting `foo.js:12: //` | prose | **missed** |

A check that misses things is worse than one that is noisy — noise gets noticed. Getting this right needs per-language parsing rather than a regex, which is more machinery than the rule is worth. 41 lines and a bespoke self-test to replace a one-word fix was the wrong trade.

PR #981, opened 2026-08-12
fix(wix-manage): PMAX-only Google Ads flow — resolve geo targets, gate assets, drop Smart campaigns

adamg-wix:
Tightens the Google Ads skills so campaign creation is PMAX-only and doesn't fail on the billable create call.

## 1. Resolve geo targets & gate assets before Create

Two independent creation-time failures were crashing the **billable** PMAX Leads create call instead of being handled up front (the reported "Israel geo lookup + image/logo generation failing" case):

- **New STEP 4 — Resolve target locations via `geo-options`.** Every user-named location resolves to `geoTargetConstants/{id}` before assembly — never guessed/hardcoded. Documents the real response shape and the fallback: retry with `countryCode`, and if a country query keeps failing, resolve a specific city (e.g. Tel Aviv/Jerusalem/Haifa).
- **STEP 2 is now an asset gate** — requests `LOGO` alongside the two marketing images and requires verifying a usable URL for each required type before Create (retry once, else user upload).
- **New pre-flight gate** checklist before assembly; renumbered budget→STEP 5, launch→STEP 6; added `LOGO` to the create-body example and error-table rows for the geo internal error and image-generation failure.

## 2. Drop Smart campaign creation

Smart campaign creation is no longer supported — Performance Max is the only supported campaign type, and the skills must never propose, build, or mention a Smart campaign to the user.

- Explicit "PMAX only, never mention Smart" directive in the create and suggestions recipes.
- Suggestions recipe reworked to PMAX only: removed the Smart-only keyword-theme and daily-budget-tier endpoints (table, quick reference, references, error rows); the full-campaign-suggestion section now targets `PERFORMANCE_MAX_LEADS` (no `campaignBrief`/`SMART`).
- Purged Smart mentions from the analytics recipe (`search-term-metrics`) and the PMAX budget note.

## Evals

Both google-ads evals updated: the create eval now exercises geo resolution + asset gating and fails if the agent proposes/mentions Smart; the suggestions eval is retargeted away from keyword/Smart-budget asks and guards against Smart.

Docs/skills + eval yaml only — no runtime code.


adamg-wix:
#rebuild

adamg-wix:
/re-eval

The gate failed on a transient backend error, not a content regression: `EvalForge PATCH .../test-scenarios/851e2ebf... → 500` while syncing the `google-ads/query-campaign-analytics` scenario. The two scenarios edited in this PR (`create-performance-max-campaign`, `get-campaign-suggestions`) synced successfully; the 500 aborted the run, which is why the PR side of the pairwise showed no output / a ~1500s timeout. Re-running the gate.

tal-grunwald:
Closing this as stale — open since Aug 12, conflicting with `main` since Aug 16, and no movement in four weeks.

The practical reason for closing rather than leaving it: it holds the `google-ads/create-performance-max-campaign` and `google-ads/get-campaign-suggestions` scenario leases, so the EvalForge gate refuses to run for any other google-ads PR while it sits here.

@adamg-wix nothing in it is superseded and I'm not disagreeing with any of it — resolving geo targets through `geo-options` before the billable create call, gating on a usable asset URL up front, and dropping Smart campaign creation are all still worth having. If you want to pick it back up, reopen it or rebase onto a fresh branch and it's welcome; it needs a rebase either way.

PR #984, opened 2026-08-12
fix(wix-headless): mandate reading inline recipes before writing capability code

gilgwix:
## Summary

- Adds an explicit mandate in `CREATE.md §4` to Read each capability's inline recipe (`how-to-code-*.md`) **before** authoring any page or component code for it
- Adds a pre-release API-wiring check: grep the generated source for real SDK call sites (`addToCurrentCart`, `productsV3`, etc.) and treat their absence as a broken deliverable

## Problem

Agents generated convincing storefront UIs — product grids, cart badges, checkout buttons — that weren't wired to real `@wix/stores` / `@wix/ecom` SDK calls. The storefront looked complete but failed entirely at runtime.

Root cause: `CREATE.md §4` pointed agents at `SDK_HANDOFF.md` for package names, which maps each capability to its inline recipe — but never explicitly required **reading** that recipe before writing code. Agents generated plausible-looking UI from model memory instead, producing stubs that matched the API shape superficially but called nothing.

Jira: https://wix.atlassian.net/browse/GPV-8742

## Test plan

- [ ] A managed `create` run for a store site reads `how-to-code-a-store.md` before authoring product/cart/checkout pages
- [ ] Generated source contains real `productsV3` / `addToCurrentCart` calls (verifiable with the new grep check)
- [ ] A managed `create` run for a bookings site reads `how-to-code-bookings.md` before authoring availability/booking pages

### DIFF
--- a/skills/wix-headless/references/managed/CREATE.md
+++ b/skills/wix-headless/references/managed/CREATE.md
@@ -107,10 +107,14 @@ headlines, kickers, subtitles, and CTA labels per **`references/CONTENT.md`** �
 Either way, `CONTENT.md` §1 (the business's category must read literally above the fold) always holds —
 a page a first-time visitor can't identify isn't done yet, regardless of voice.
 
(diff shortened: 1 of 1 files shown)


gilgwix:
@aya check this ticket what do you think about this solution?

PR #986, opened 2026-08-13
[test] live smoke-test for evalforge-yaml-gate scenario lint (#935) — do not merge

danshw:
Throwaway PR to exercise PR #935's new scenario lint against the live gate. Adds one scratch scenario (`yaml/wix-manage-evals/livetest/gate-check.yml`) that's missing `maxTokens`, so it should fail the `max-tokens` lint rule before any EvalForge call.

This is disposable — closing without merging once the gate comment is verified. Not touching #935's branch.


danshw:
Superseded — main has drifted since #935 branched, causing a merge conflict that silently prevented the gate from running. Retrying with a self-contained base branch instead.

PR #987, opened 2026-08-13
[test] live smoke-test for evalforge-yaml-gate scenario lint (#935) — do not merge

danshw:
Throwaway PR, based on the disposable test/gate-live-base branch (not main), to exercise PR #935's new scenario lint against the live gate without conflicting with main's current state.

Adds one scratch scenario (`yaml/wix-manage-evals/livetest/gate-check.yml`) missing `maxTokens` — expect the gate to fail the `max-tokens` lint rule before any EvalForge call.

Disposable. Closing without merging once the gate comment is verified. Does not touch #935's branch.


danshw:
Verified — the scenario lint fired correctly and produced the expected max-tokens violation. Closing without merging; branch will be deleted.

PR #990, opened 2026-08-13
feat(evalforge): investigate failing wix-app gate runs with AI

orgoldreich1:
When the wix-app eval gate finds failing assertions, a second job asks EvalForge to investigate that run and posts what it found as its own PR comment — separate from the verdict comment, collapsed by default.

## Why a separate job

A GitHub check does not finalise until its **job** ends. Running the analysis inside the `gate` job would hold the required check open for the analysis's server-side budget, so the verdict — and whether the PR is blocked — would wait on an advisory comment. The `analyze` job exists precisely so it does not.

Three details follow from that:

- `if: !cancelled() && …` rather than a plain `needs: gate`. Once `blocking` flips to true a failing verdict makes the `gate` job red, and the default success-only `needs` would then **skip** the analysis that the failure warrants. A cancelled workflow (superseded by a new push) still skips.
- The gate emits `analyze-run-id` **only** when the PR arm has failed or errored assertions, and the job's whole trigger is that output being non-empty. A green run starts no runner and posts no comment.
- Keyed on assertion counts, not the verdict: in soak mode (`blocking: false`, where wix-app is today) a run with real failures still produces a passing verdict, and those are the runs most worth investigating.

The analysis never fails a check. Every failure path warns and posts a short note instead — a 400, a timeout, a 5xx, an empty analysis. A missing input is the one exception, and it fails loudly because there is no comment channel left to report through.

`WIX_APP_EVAL_ANALYZE=false` switches the whole thing off.

## The endpoint

`POST /v1/projects/{projectId}/eval-runs/{runId}/analyze` — synchronous, 57-second server budget, requires the run to be `COMPLETED`, audience `apps: WIX_AND_PUBLIC_APPS` so the gate's existing OAuth credentials work. No new secret.

This forced one change to shared code: `EvalForgeClient`'s hardcoded 15-second abort became a per-call parameter, since it would otherwise kill every analysis moments before it returned. Analyze passes 75s; every other call keeps 15s, and there is a test pinning that.

## Rendered output

The whole feature is a comment, so here is what it actually renders.

<details>
<summary><b>The normal case</b></summary>

## 🔍 EvalForge: AI Investigation

**4 findings** — 1 high, 1 medium, 1 low, 1 positive.

Three of the four failures share a single cause: DASHBOARD_PAGE.md instructs the agent to register the extension before creating the page file, so the CLI writes the manifest entry against a path that does not exist yet and the build fails on a missing module. The fourth failure is unrelated and…

<blockquote>
<b>Full investigation</b> (collapsed in the real comment)

### 🔴 High · Skill misguidance

DASHBOARD_PAGE.md orders the steps registration-first. The agent followed it literally in every failing scenario, producing a manifest entry pointing at a file it had not written yet.

**Scenarios:** `create-dashboard-page`, `add-menu-item`, `dashboard-page-with-modal`

**Recommendation:** Reorder the numbered steps so the page file is created before registration.

### 🟠 Medium · Tool output error

The CLI reported "extension registered successfully" despite the target file being absent, so the agent had no signal to self-correct.

**Scenarios:** `create-dashboard-page`

### 🟡 Low · Flakiness

Template fetch timed out twice; both retries succeeded on the third attempt.

**Scenarios:** `dashboard-page-with-modal`

### Strength

The agent correctly used the wix-app skill for every scenario and never fell back to guessing at the CLI surface.

</blockquote>

<sub>Generated for eval run 8f3c1a02… · 2026-08-13 09:41 UTC</sub>

</details>

<details>
<summary><b>The three other states</b></summary>

**No findings** — deliberately flat, not collapsed:

> ## 🔍 EvalForge: AI Investigation
>
> No findings — the investigation surfaced nothing to act on.

**Unavailable** — a mapped sentence, never upstream error text:

> ## 🔍 EvalForge: AI Investigation
>
> The AI investigation could not be generated: the eval run has not finished yet.
>
> This does not affect the gate verdict.

**Superseded** — posted by the *gate* when a later run passes, so a fixed PR does not keep a stale investigation above a green verdict:

> ## 🔍 EvalForge: AI Investigation
>
> The earlier investigation no longer applies — the latest run had no failing assertions.

</details>

Findings sort HIGH → MEDIUM → LOW with `POSITIVE` last, since a reader opening this comment is here for what broke. `POSITIVE` renders without a severity chip so it cannot contradict the tally.

## Two things worth a reviewer's attention

**The comment body is budgeted, not hoped for.** `summary` is `maxLength 100000`, each `description` `10000`, up to 1000 findings — against GitHub's hard 65536-character limit, past which the API 422s and `makeCommenter` degrades to the job summary, silently losing the PR comment. So the formatter truncates to 60000 itself, dropping lowest-priority findings first and always preserving the marker, the tally and the run link. Getting this wrong is invisible in the happy path, which is why there is a 600-finding regression test.

**Nothing upstream reaches a public comment body.** This is a public repo, and `core.setSecret` masks logs but not comments. Client errors carry an upstream message — `TokenProvider.mint` folds up to 300 characters of the OAuth endpoint's response body into it — so recognisable cases are mapped to plain sentences and the raw detail stays in `core.warning`.

## Known follow-up, not fixed here

**The summary renders twice.** The teaser above the fold is the opening of the summary; the full summary is then the first thing inside the fold. With a short summary it is a verbatim duplicate. This is faithful to the design ("folded: the full summary and all findings") but reads poorly, and it only became obvious once the output was rendered rather than asserted on. Fixing it is a choice between dropping the summary from the fold and dropping the teaser, which seemed worth deciding in review rather than unilaterally.

Also undocumented and worth a separate change: **these bundles must be built with Node 24** (CI's pin). Under Node 20 both `dist/index.js` files rebuild as modified — a false staleness signal indistinguishable from real drift. It bit this branch: `evalforge-yaml-gate/dist` was stale for most of the work because both actions bundle `evalforge-core` and CI diffs both. There is no `.nvmrc`.

## Testing

| Package | Tests |
| --- | --- |
| `packages/evalforge-core` | 416 |
| `.github/actions/evalforge-skill-gate` | 212 |
| `.github/actions/evalforge-yaml-gate` | 76 |

All green under Node 24, all three typecheck clean, both committed bundles byte-match a fresh build.

**Not yet exercised: the live endpoint.** Every test mocks `fetch`, so the first real signal comes from a PR touching `skills/wix-app/**` that fails at least one assertion. On such a PR, confirm: the verdict comment appears and its check settles without waiting for the analysis; a second collapsed comment follows; the `analyze` job is green even if the gate is red; a fully green run produces no second comment; and pushing a fix onto a PR that already has an investigation replaces it with the superseded note.


orgoldreich1:
## 🔍 EvalForge: AI Investigation

**8 findings** — 2 high, 1 medium, 2 low, 3 positive.

The eval run achieved a 57.1% pass rate (4/7 assertions) on the employee-shift-dashboard scenario. The agent successfully built a fully functional auto-patterns dashboard page with a data collection backend and clean TypeScript compilation, but failed to complete the final build step due to CLI…

<details>
<summary>Full investigation</summary>

The eval run achieved a 57.1% pass rate (4/7 assertions) on the employee-shift-dashboard scenario. The agent successfully built a fully functional auto-patterns dashboard page with a data collection backend and clean TypeScript compilation, but failed to complete the final build step due to CLI authentication issues (invalid email in auth state). Additionally, the run exceeded the time limit by 41 seconds. The key technical achievements were offset by environment constraints and inefficiencies in exploration and API discovery.

### 🔴 High · Tool output error

The `wix generate` command failed twice with a CLI authentication error: `{"validation":"email","code":"invalid_string","message":"Invalid email","path":["userInfo","email"]}`. This is an environment/tool issue where the Wix CLI encountered invalid auth state. The agent correctly pivoted by manually generating files using the standalone `generate-auto-patterns.js` script, which succeeded, demonstrating good error recovery.

**Scenarios:** `employee-shift-dashboard`

**Recommendation:** Pre-configure valid CLI auth state in the evaluation environment before running generation or build commands. Alternatively, mock or stub the CLI's email validation for testing scenarios where authentication cannot be pre-initialized.

### 🔴 High · Failure pattern

The final `npx wix build` command failed with exit code 1 despite TypeScript compiling cleanly (`npx tsc --noEmit` returned 0). The build failure was caused by the same CLI authentication error, making it unavoidable given the environment constraint. This blocked the full validation workflow even though the generated code itself was correct.

**Scenarios:** `employee-shift-dashboard`

**Recommendation:** Ensure the evaluation environment pre-configures valid Wix CLI authentication credentials before running any build or preview commands. This is a prerequisite for the build step to succeed, regardless of code quality.

### 🟠 Medium · Inefficiency

The agent spent significant time (20+ seconds across messages 22–32) trying to understand the Wix Astro builders API by reading TypeScript definition files directly from node_modules (e.g., `builders.d.mts`, `index-of3Y33tX.d.mts`) after exhausting the node_modules exploration budget. This forced reliance on type discovery rather than documented references. The agent then made an API shape error when first calling `extensions.dashboardPage()` — passing a React component instead of a string path (`'./employee-shifts.js'`) and using `route` instead of `routePath` — which was discovered and fixed via TypeScript errors.

**Scenarios:** `employee-shift-dashboard`

**Recommendation:** Expand the wix-app skill references to document the Wix Astro builders API explicitly. Include concrete examples showing `extensions.dataCollections({ id, name, collections: [...] })` and `extensions.dashboardPage({ id, component: './path.js', title, routePath })` with required field types. This will reduce trial-and-error type discovery and prevent API shape errors.

### 🟡 Low · Cost waste

The agent called `bash` with duplicate equivalent arguments twice (messages 10 and 18) — both were `ls` commands exploring the project directory structure. While the cost impact is minimal (~$0.01 per bash call), this indicates redundant exploration that could have been avoided with better planning.

**Scenarios:** `employee-shift-dashboard`

**Recommendation:** Plan the full directory structure upfront by combining exploratory checks into fewer, more targeted commands. For example, combine multiple `ls` calls into a single `find` or `tree` command to avoid duplicate execution.

### 🟡 Low · Inefficiency

The agent read the DASHBOARD_PAGE.md reference file late in the workflow (message 34), after already committing to and building most of the auto-patterns solution. While this read confirmed the agent's architectural choice was correct, it represented late validation rather than upfront architecture validation.

**Scenarios:** `employee-shift-dashboard`

**Recommendation:** Prioritize reading architecture decision documents (AUTO_PATTERNS_DASHBOARD.md vs DASHBOARD_PAGE.md) at the very start of the workflow, before scaffolding any files. This allows upfront validation of the chosen approach and avoids late-stage reference reads.

### Strength

The agent successfully pivoted when `wix generate` failed due to CLI auth issues. Instead of abandoning the task, it manually created the required file structure and ran the standalone `generate-auto-patterns.js` script, which succeeded. The agent also correctly generated UUIDs, created the data collection schema, and scaffolded both the backend data collection extension and the dashboard page extension.

**Scenarios:** `employee-shift-dashboard`

### Strength

TypeScript compilation passed cleanly (`npx tsc --noEmit` returned exit code 0), confirming the agent correctly understood and implemented the extension builder API after fixing the initial shape errors. The final code structure is sound: proper extension registration in `src/extensions.ts`, correct data collection configuration, and a valid auto-patterns dashboard page component.

**Scenarios:** `employee-shift-dashboard`

### Strength

The LLM judge confirmed the implementation is functionally correct: 'The agent successfully built a fully functional, well-structured, and native Wix dashboard page for managing employee shifts using the declarative auto-patterns system.' Judge score: 10/10, indicating high-quality code design and architecture despite the build environment failure.

**Scenarios:** `employee-shift-dashboard`

</details>

<sub>Generated for <a href="https://bo.wix.com/pages/evalforge/bb388ce3-66c2-4868-a895-36b637635f59/results?runId=0755cf5c-2680-42e7-b5a4-2fcdddec343d">eval run 0755cf5c-2680-42e7-b5a4-2fcdddec343d</a> · 2026-08-13 09:08 UTC</sub>

PR #995, opened 2026-08-13
[test] live smoke-test for evalforge-yaml-gate confirm-on-fail (#935) — do not merge

danshw:
Throwaway PR, based on the disposable test/confirm-live-base branch (not main), to exercise PR #935's confirm-on-fail against the live gate.

Three scratch scenarios (see commit message for details): one guaranteed llm-judge failure, one guaranteed token-budget failure, one clean pass. Expect the gate to retry the two failing ones, confirm both, and block — while the clean one shows no confirm-on-fail entry at all.

Disposable. Closing without merging once the gate comment is verified. Does not touch #935's or #989's branches.

### DIFF
--- a/yaml/wix-manage-evals/livetest/confirm-budget-check.yml
+++ b/yaml/wix-manage-evals/livetest/confirm-budget-check.yml
@@ -1,0 +1,26 @@
+name: livetest/confirm-budget-check
+description: >-
+  Scratch scenario used to smoke-test confirm-on-fail's handling of a
+  token-budget violation. maxTokens is set absurdly low on purpose, so any
+  real agent run exceeds it regardless of task quality. Not a real skill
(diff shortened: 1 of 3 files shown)


danshw:
Verified — confirm-on-fail fired correctly for both the llm-judge failure and the token-budget failure, confirming both after 2/2 retry attempts and blocking as designed. Also concretely illustrates the review finding: the confirmed-failure comment shows only `(token-budget)` with no actual/budget numbers — `confirm-clean-check` scored 10/10 on both real judges but reads as a flat confirmed failure with zero indication it was 3.4x over a fairly tight budget, not a real quality problem. (That scenario's token usage came in far higher than intended, most likely because `SearchSiteTemplates` isn't a real MCP tool — a flaw in my scratch setup, not a #935 finding.) Closing without merging; branch deleted.

PR #1023, opened 2026-08-17
ECL-18140: Add an eval for focus design states

shykhov:
The guidance changes moved to wix-private/editor-react-component-creation-skills#81 — the `wix-app` editor-react-component references are synced from that repo, so doc edits made here are overwritten on the next sync. This PR keeps only the eval, which lives upstream only (the sync script owns `SKILL.md`, `references/**`, and `scripts/**`, not `yaml/`).

## Summary

- Adds `editor-react-component-focus-states.yml`: builds a component with an input, a native checkbox, a submit button, and a link with an explicitly requested focus state.
- Judges the generated manifest, not just the CSS, so the assertion fails if a focus state is emitted for a control nobody asked about.

## Order

This can only pass once both land:

1. wix-private/ZeroConfig#166 — the generator stops inferring focus for every interactive element.
2. wix-private/editor-react-component-creation-skills#81, synced down — otherwise the skill still tells the agent to pair a `--focus` modifier on every native state.

Until then the gate is red on purpose.

https://wix.atlassian.net/browse/ECL-18140


shykhov:
Expected — the eval runs an older `@wix/cli`, where every `<button>` gets a focus state no matter what the CSS says. https://github.com/wix-private/ZeroConfig/pull/166 fixes that, so this goes green once it's released.

Just don't drop the `:focus-visible` rule like it suggests — that only removes the keyboard focus ring, the manifest stays the same.

shykhov:
Closing — the guidance moved to wix-private/editor-react-component-creation-skills#81, which is the source these references sync from. The eval that was left here duplicates coverage that belongs in that repo alongside the docs.

PR #1027, opened 2026-08-17
🦞 igor-claw: document Smart Campaign UpdateCampaign's required-fields gap + excludedSearchTerms example

ayal:
## Summary

Opened automatically by an AI agent acting on behalf of Ayal (`ayalg@wix.com`), based on an MCP feedback report investigated here: https://wix.slack.com/archives/C08P5DKLJR5/p1786981366779239

**Report:** an agent using the Wix MCP followed this recipe's `manage-campaign-lifecycle.md` guidance ("send only the fields you're changing") to update only a Smart Campaign's `smartCampaign.excludedSearchTerms`. The call failed server-side with no actionable error.

**Root cause** (traced in `wix-private/promote-google-ads`, code fix in a companion PR there): `UpdateCampaign` is PATCH-shaped and documented as a partial update, but the Smart Campaign update handler still unconditionally requires `smartCampaign.adGroups`, `url`, `languageCode`, and `businessName` on every request (throws if omitted), and used to crash the same way when `budget` was omitted too (now fixed to fall back to the saved budget).

**This PR:** updates the recipe so agents don't repeat the same crash while the deeper backend gap is being addressed:
- Tells agents to `GET` the campaign first and resend the still-required fields, instead of inferring "PATCH ⇒ send only the diff" from the HTTP verb alone.
- Adds a concrete `excludedSearchTerms` example, since no recipe showed how to set it and it's the exact field the report needed.

## Test plan
- [ ] N/A — documentation-only change to `skills/wix-manage/references/google-ads/manage-campaign-lifecycle.md`.

### DIFF
--- a/skills/wix-manage/references/google-ads/manage-campaign-lifecycle.md
+++ b/skills/wix-manage/references/google-ads/manage-campaign-lifecycle.md
@@ -36,7 +36,7 @@ curl -X POST 'https://www.wixapis.com/google-ads/v1/campaigns/{campaignId}/resum
 
 ## Update (partial) & change budget
 
-`PATCH /v1/campaigns/{id}` — send only the fields you're changing, plus the required `id` and `accountId`. Budget is in **micros** (`20000000` = $20.00/day).
+`PATCH /v1/campaigns/{id}` — always required: `id`, `accountId`, `campaignType`. Budget is in **micros** (`20000000` = $20.00/day).
 
 ```bash
 curl -X PATCH 'https://www.wixapis.com/google-ads/v1/campaigns/{campaignId}' -H 'Authorization: <AUTH>' -H 'Content-Type: application/json' \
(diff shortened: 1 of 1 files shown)


tal-grunwald:
Hi @ayal — #1313 has ended up covering the same ground as this PR, and I think it supersedes it.

That PR replaces the whole "Update (partial) & change budget" section with a read-modify-write flow: GET the campaign, apply the change in place, PATCH the whole entity back, then read it again to verify. Everything this PR added is carried over:

- the `excludedSearchTerms` example, rewritten as a `jq` append onto the array the read returned — it was the only documentation for that field anywhere in the recipes, so it stays
- the Smart handler's hard-required fields (`adGroups`, `url`, `languageCode`, `businessName`, plus `phone` for call ads) and the 5xx-with-no-message they cause when omitted, both inline and as an error-table row
- `id`, `accountId` and `campaignType` as required on every update

The framing differs: rather than keeping the partial-update model and listing what Smart additionally needs, the recipe now says the payload is always the full entity, for every campaign type — because omitting a field, or dropping a value from an array you did send, removes it from the campaign and still returns 200. Your report is what got this section looked at in the first place.

If that reads right to you, would you mind closing this one? It also holds the `google-ads/manage-campaign-lifecycle` eval scenario, which is what's currently failing #1313's gate. And if anything from your investigation didn't make it across, say so and I'll add it there rather than keep two PRs on the same section.

(The companion backend fix in wix-private/promote-google-ads is unaffected either way.)

PR #1044, opened 2026-08-18
feat(wix-manage): expand restaurants skills across all three apps

zahorovskyi:
Wix Restaurants is three apps — Menus, Orders and Table Reservations — and the
skills folder covered one of them. The single recipe, wix-restaurants-setup,
documented the Menus API but its description claimed it also covered
"availability schedules, and ordering settings". Agents selected it for
ordering questions, found nothing, and reported back that no API exists for
configuring restaurant ordering. Two separate cards reached that conclusion in
the week of 2026-08-01, one from a real merchant trying to set a pickup cutoff.

Grow the area from 3 recipes to 9, and from 2 eval scenarios to 9 restaurants
scenarios plus 1 app-installation scenario.

## Recipes

Menus
- Restaurants Menus Setup (renamed from wix-restaurants-setup, so the name
  matches the app and stops colliding with the orders setup recipe)
- Restaurants Menus Import — a pasted, uploaded or dictated menu turned into a
  structured one, plus bulk reworking of an existing menu

Orders
- Restaurants Orders Setup — end-to-end launch in dependency order
- Restaurants Orders Settings — operations, fulfillment methods, menu ordering
  settings, availability exceptions, service fees
- Restaurants Orders Delivery Setup — areas, fees, thresholds, timing
- Restaurants Orders Management — reading orders customers already placed

Table Reservations
- Restaurants Reservations Setup — party size, pacing, turnover, approval,
  business schedule, and the reservation lifecycle

Cross-app
- Configure Restaurants from Prompt — routes a merchant's own wording to the
  right app and entity across all three

## list-installed-apps — split Wix Restaurants into its three apps

The app-definition table listed a single legacy Wix Restaurants app
(13e8d036-5516-6f75-e025-2aca3b5d7930), which no longer matches what the Apps
Installer API returns. Agents hunting for which restaurants app a site had
landed on that ID, missed the real ones, and in one support card invented a
nonexistent appDefId. The table now lists the three current apps — Menus
(b278a256-2757-4f19-9313-c05c783bec92), Orders
(9a5d83fd-8570-482e-81ab-cfa88942ee60) and Table Reservations
(f9c07de2-5341-40c6-b096-8eb39de391fb) — and corrects Wix Stores vs Wix
eCommerce (215238eb-22a5-4c36-9e7b-e7c08025e04e vs
1380b703-ce81-ff05-f115-39571d94dfcd), which had been collapsed. A new
app-installation eval pins the three-apps question to the site's own app
instances rather than inference or memory.

## restaurants-dashboard-navigation — correct the menu API paths

The read-API table carried the deprecated path segments
restaurants/menus-menu/v1, .../menus-section/v1 and .../menus-item/v1. The
current Menus API is one namespace — /restaurants/menus/v1/menus, .../sections
and .../items — matching what the recipes and the API reference already use, and
what the Slack support channel was hitting when it tried the old paths. Also
adds rows for operations and fulfillment methods (linking to the ordering
settings page) and a When-to-Use / Does-NOT-Cover pair for consistency with the
other recipes.

## restaurants-reservations-setup — complete the lifecycle and eCom link

The reservations recipe described the HELD → RESERVED flow as if every
reservation passes through HELD, and said nothing about how a reservation is
backed by an eCommerce order. Both were checked against
wix-private/table-reservations-server (proto, ReservationsService,
OrdersConnector, ReservationsCatalogService) and the server-authored
reservation-lifecycle guide, then corrected:

- A reservation's `source` (ONLINE / OFFLINE / WALK_IN) is set on creation and
  decides its starting status. Only ONLINE begins in HELD. Staff bookings
  entered in the dashboard (OFFLINE) go through Create Reservation and land
  directly in RESERVED (or REQUESTED for an online-source booking under manual
  approval), skipping HELD regardless of the approval setting. There is no
  BACK_OFFICE enum. With MANAGE RESERVATIONS (FULL) the status can be set
  explicitly and table/availability rules overridden.
- Reservations that require payment (per-guest or per-reservation fee, deposit,
  card guarantee) are backed by an eCommerce order whose line item carries
  catalogReference.appId f9c07de2-5341-40c6-b096-8eb39de391fb and catalogItemId
  equal to the reservation ID — mirroring how Orders app orders carry
  9a5d83fd. The reservation's paymentStatus mirrors that order, and no-show /
  late-cancellation fees are charged and refunded against it. Free reservations
  have no order.

## The constraints that drive the content

None of these are derivable from the API reference, which is why the recipes
exist rather than pointing at the docs:

1. With immediate (ASAP) ordering the cutoff is derived, not stored:
   last order = window end minus prep time, minus delivery time for delivery.
   "Stop orders at 1:30am" with 30 minutes prep needs a window ending 02:00;
   setting 01:30 silently yields a 1:00am cutoff. Preorder-only operations are
   the exception and do have a real advance-notice bound and weekly cutoff.
2. An availability range cannot cross midnight and hours are 0-23, so 11am-2am
   is two ranges per weekday, not one.
3. A body writing pickupOptions or deliveryOptions must also carry `type`, on
   updates as well as creates. Verified against a live site: omitting it returns
   400 ONE_OF_ALIGNMENT whose text claims the opposite. Raised separately
   against the validation message and the API reference.
4. Reservation locations are created only through the dashboard or the Locations
   API, and partySize/partiesSize, turnoverTimeRules/turnoverRules and
   approval/manualApproval are twin fields a GET returns together — echoing one
   back sends both and fails when they disagree.
5. Restaurant orders are eCommerce orders, found by catalogReference.appId
   9a5d83fd-8570-482e-81ab-cfa88942ee60, with the customer's chosen modifiers
   in catalogReference.options.
6. A reservation's starting status depends on its `source`, not just the
   endpoint: OFFLINE staff bookings skip HELD. Payment-required reservations
   are backed by an eCom order (appId f9c07de2, catalogItemId = reservation ID);
   free reservations are not.

## Scope

Limited to the public API surface. Dine-in, order pacing, and item- or
section-level availability are internal-only, so they route to the dashboard
rather than being documented. Floor plans, the prep board and printing are
dashboard-only and named as such.

## Evals

Seven new restaurants scenarios, each pinning the failure its recipe prevents:
the pickup cutoff landing on 1:30am rather than 1:00am with no range crossing
midnight; a vague "customers can't order late" routed to ordering settings
rather than the Menus API; a delivery radius with threshold and minimum order
kept distinct; an end-to-end launch that surfaces the missing payment
connection; reservation rules set on an existing location without twin fields;
a pasted menu whose section-wide choice becomes a modifier group; and placed
orders found through eCommerce rather than declared missing. The dashboard-
navigation scenario gains a quality judge; the set-reservation-rules scenario
gains a bootstrap that installs Table Reservations, since the restaurants
template ships only the ordering apps. The pre-existing modifier-group scenario
gains a quality judge, since its doc URL moved with the rename.

The menus recipe also drops from 10.6k to 8.2k by removing request and response
bodies that restate the docs, keeping the modifier-group shape agents get
wrong. All nine recipes are now under the 10k truncation threshold.

Co-Authored-By: Claude <noreply@anthropic.com>


zahorovskyi:
**Requesting review as-is — please don't block on the `gate` check.**

This PR replaces the old `wix-restaurants-setup.md` (Menus-only: menu structure, items, modifiers, pricing) with one focused recipe per Restaurants app, plus a shared router and the eval scenarios that exercise each. Of the 8 new `.md` files, only two (`restaurants-menus-setup.md`, `restaurants-menus-import.md`) are carved from the old file; the other six — **Orders setup, Orders settings, Orders delivery, Orders management, Reservations setup, and the `configure-restaurants-from-prompt` router** — are net-new coverage that never existed before.

Why I'm not splitting into 8 PRs (one new skill each, as `max-new-skills: 1` asks):

1. **The set only makes sense as a whole.** The router recipe (`configure-restaurants-from-prompt.md`) exists to route a merchant's request across the three app recipes; the recipes cross-link each other and share the "these are separate apps" framing. Landing them one at a time means 7 of 8 PRs reference recipes that aren't merged yet, and reviewers can't judge the routing/boundaries until all are in.
2. **Review cohesion.** The value here is the *new boundaries* — which app owns hours, which owns the menu, which owns bookings — visible only side by side. Eight sequential reviews would each ask "is this the right home for X?" in isolation, then re-litigate it when the router lands.
3. **Rebase/review cost ×8.** `main` moves daily here. 8 PRs = 8 rebases + 8 review rounds for one conceptual change.

Reviewing tip: start at `configure-restaurants-from-prompt.md` (the router), then the three app recipes. The eval YAMLs under `yaml/wix-manage-evals/restaurants/` map 1:1 to the new recipes.

cc @danshw @hadarzaguri @ayal

PR #1060, opened 2026-08-19
🦞 igor-claw: warn that renaming an existing option choice is not supported

ayal:
## What's broken

A user asked an MCP agent to rename existing Patch Design choices on a Family Life hat product (930 variants, one shared reusable customization). The agent correctly avoided destructive rebuild/delete-recreate approaches, then tried both `Update Product` (preserving all option/variant IDs) and standalone `Update Customization` (preserving all 96 choice IDs). Both calls succeeded and incremented revision, but a fresh GET showed the choice names unchanged — the agent burned a full investigation cycle across two APIs before concluding the only path is a manual dashboard rename.

Live-reproduced on a test site: creating a product, then renaming an existing choice via Update Product and via Update Customization both return 200 + incremented revision, with the name silently unchanged on GET.

## Root cause

`choicesSettings.choices` is immutable at the array level in the underlying proto once a choice exists (`(.wix.api.immutable) = true`), which the platform's field-mask inference silently excludes from any update — even though the nested `Choice.name` field itself isn't separately marked immutable, so the API reference makes it look writable. There is currently no endpoint that renames an existing choice. Full root cause + API reference doc fix: wix-private/ecom#54079 (internal).

## Fix

Add a "Renaming an Existing Choice Is Not Supported" section to the `Update Product with Options (Catalog V3)` recipe (`skills/wix-manage/references/stores/update-product-with-options.md`), plus a matching bullet under Important Notes, so agents recognize this immediately instead of re-discovering it live and steer users to the dashboard instead of risky delete/recreate workarounds.

## Context

Opened automatically by an AI agent on behalf of Ayal (`ayalg@wix.com`), researching Wix MCP user feedback. Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1787167630678889

## Test plan
- [ ] Docs-only recipe content change; no code to test

### DIFF
--- a/skills/wix-manage/references/stores/update-product-with-options.md
+++ b/skills/wix-manage/references/stores/update-product-with-options.md
@@ -218,6 +218,14 @@ curl -X PATCH "https://www.wixapis.com/stores/v3/products/{productId}" \
 
 When updating existing variants, include each existing variant `id`. If no GUID is passed, a variant is created with a new GUID. Each variant object is replaced whole rather than merged, so carry over the fields you are not changing: rebuilding a variant from just its `id` plus the field you want to set drops everything else and is rejected on the first required field it lost (`price must not be empty`). Start from the variant as returned by Get Product and override only what the user asked to change.
 
+### Renaming an Existing Choice Is Not Supported
+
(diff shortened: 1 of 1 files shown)


danshw (approved):

hadarzaguri:
Merging... Anyways, this skill is about to change drastically. Opening a PR and not merging it is blocking other PRs.

hadarzaguri:
#forcemerge

PR #1065, opened 2026-08-19
🦞 igor-claw: Site Import skill must call site-scoped to target an existing site

ayal:
## Problem
`site-import.md` unconditionally told agents to use "the Wix account-level REST API tool" for every `site-import` call. That means Site Import can never write into a site the user already has — it always spins up a brand-new one, even when the user already created a destination site and expects the import to land there.

## Root cause (confirmed in source)
`destinationSiteId` is **not** a request-body field. In `wix-private/replatform-sb`, `packages/bix-site-import/src/shared/identity.ts` (`requireIdentity`, spec 0018), it's resolved purely from the caller's own request context via `ctx.apiGatewayClient.metaSiteId()`:
- Account-level calls carry no site context → `destinationSiteId` is always empty → `startImport` always creates a new site (`service.ts`: `siteStrategy: destinationSiteId ? 'existing' : 'new'`).
- Site-scoped calls (with a `siteId`) → `metaSiteId()` resolves → the import writes into that existing site instead.

The skill's blanket "always account-level" instruction meant this existing-site capability was unreachable no matter what the user asked for.

## Fix
`site-import.md` now:
- Explains the scope/destination tradeoff explicitly (new section "Destination — new site vs. an existing site").
- Tells agents to ask the user whether they want a new site or their existing one, and to call **site-scoped** (with that site's `siteId`) for the latter.
- Clarifies in the Start endpoint docs that `destinationSiteId` in the response reflects the *call's scope*, not anything sent in the body.

## Context
Surfaced while researching a Wix MCP feedback report (healthcare-site migration via Wix Studio) where the user had already created a destination Studio site and uploaded media before attempting Site Import — work that this scoping bug would have discarded even if the (separately tracked, wix-private/replatform-sb#77) `NOT_ENABLED` rollout gate weren't blocking them.

Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1787178008452279

Opened automatically by an AI agent acting on behalf of Ayal (`ayalg@wix.com`) — not written by a human.

### DIFF
--- a/skills/wix-manage/references/sites/site-import.md
+++ b/skills/wix-manage/references/sites/site-import.md
@@ -24,9 +24,25 @@ and report the final result.
 5. **This API is the only site-creation tool for this task** — never
    substitute other Wix site-builder tools (see Scope).
 
-Base URL: `https://www.wixapis.com/site-import`. All calls use account-level
-scope on the signed-in Wix user's account. A full import typically takes
+Base URL: `https://www.wixapis.com/site-import`. A full import typically takes
 15–60 minutes.
+
+**Destination — new site vs. an existing site.** There is no destination field
+in the request body. The destination is decided entirely by which scope you
+call with, resolved from your own request context:
+- **Account-level scope** (no site context) → the import always **creates a
+  brand-new site**. This is the default and the right choice when the user has
+  no Wix site yet, or explicitly wants a fresh one.
+- **Site-level scope** (called with a specific `siteId`) → the import **writes
+  into that existing site** instead of creating a new one.
+
+**Ask the user which they want before calling Start** — it can't be changed
+after the migration starts. If the user already created (or named) a Wix site
+they want this content imported into — including one they just set up to
+prepare for this import — call Start (and every subsequent call for that
+`importId`) scoped to that site's id, not account-level. Calling account-level
+when the user actually meant an existing site silently creates a second,
+unwanted site and leaves their existing one untouched.
 
 ## Calling the API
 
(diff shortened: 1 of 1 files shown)


ayal:
🤖 Automated note from an AI research agent acting on behalf of Ayal (ayalg@wix.com) — not a manually-written comment.

**Live-verified this fix's underlying claim today** while researching a new feedback report (Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1787429171258019) that hit the same "already have an existing draft site, Site Import can't target it" scenario this PR addresses.

On an account with rollout-gate bypass, I compared:
- `ManageWixSite` (account-level) `POST /site-import/v1/imports` → response has **no `destinationSiteId` field at all**.
- `CallWixSiteAPI` with `siteId` set (site-scoped) on the same request shape → response `destinationSiteId` **exactly matches** the targeted site's id.

This confirms `identity.ts`'s `requireIdentity` resolves `destinationSiteId` purely from call scope, exactly as this PR's description states, and that the fix (guide agents to call site-scoped when the user has an existing destination site) is correct and sufficient — not just a plausible theory. This is now the 2nd independent field report specifically needing this fix (after Valeo Therapeutics). Given it's been open ~3 days with no review yet, flagging for priority given the recurrence.

itayhewix:
Superseded by #1338, which folds in this PR's destination-scoping fix (new site vs. existing site via call scope) along with an unrelated closed-beta error-handling fix, plus a covering eval scenario for the destination decision. Closing this one to release its scenario locks.

PR #1071, opened 2026-08-20
fix(evalforge-gate): allow the comparison poll more time and flag partial results

gilisho:
## Why

The gate polls the eval comparison service until a comparison finishes. The final poll — the one that returns the result — does considerably more work server-side than the intermediate "still running" polls, and could take longer than the 30s the HTTP client allowed for every request.

When the client timed out first, two things went wrong:

- The real error was replaced by a generic abort, so failures reported `The operation was aborted due to timeout` instead of what actually happened.
- An abort counts as retriable, so the request was re-issued — repeating work that had already completed successfully.

The visible effect was a failing gate on comparisons that had in fact finished, with every retry hitting the same wall.

## What changed

**Per-request timeouts.** `post()` now accepts a timeout rather than hardcoding one. The default is unchanged at 30s; only the completion poll gets a longer one, chosen to exceed any deadline the server applies to itself so that the client is the last participant to give up and a real response is what we act on.

`/run-comparison` deliberately keeps the 30s default: it returns quickly and is non-idempotent (it queues eval runs), so failing fast is safer than widening the window for a duplicate.

**A partially judged comparison is no longer presented as a pass.** The service can report how many scenarios actually produced a pairwise judgement. When that is fewer than the total, the PR comment now carries a warning and the heading drops its ✅. This matters because `not-required` is also the verdict produced when judging yields nothing — so a degraded comparison and a genuine "this skill makes no difference" were previously indistinguishable in the comment. The field is optional; when absent, rendering is unchanged.

**`dist/` rebuilt**, since the committed bundle is what the action runs.

## Testing

`yarn typecheck` clean. `yarn test` — 79 passed / 10 files.

Three new tests in `tests/comment.test.ts`: partial coverage warns and downgrades the heading icon, full coverage keeps ✅, and a missing field keeps ✅ so existing responses render as before. The icon behaviour is mutation-checked — removing the coverage condition fails the test.

### DIFF
--- a/.github/actions/evalforge-yaml-gate/src/utils/comment.ts
+++ b/.github/actions/evalforge-yaml-gate/src/utils/comment.ts
@@ -130,18 +130,34 @@ function winnerLabel(s: ScenarioComparison): string {
 }
 
 export function formatComparisonResult(result: CompareGroupComplete, projectId?: string): string {
-  const { verdict, tag, scenarios } = result.result;
+  const { verdict, tag, scenarios, judgeCoverage } = result.result;
   const hasNoWinner = comparisonHasNoWinner(result.result);
-  const verdictIcon = verdict === 'not-required' && !hasNoWinner ? '✅' : '⚠️';
+  // Partial judging never gets a green tick: `not-required` is what the pipeline
+  // returns when a judge yields nothing, so a throttled pass looks identical to
+  // a genuine "no difference" unless coverage is taken into account.
+  const isFullyJudged = !judgeCoverage || judgeCoverage.judged === judgeCoverage.total;
+  const verdictIcon = verdict === 'not-required' && !hasNoWinner && isFullyJudged ? '✅' : '⚠️';
   const lines: string[] = [
     COMMENT_MARKER,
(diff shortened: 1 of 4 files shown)


.github/actions/evalforge-yaml-gate/src/utils/comment.ts
```diff
@@ -130,18 +130,34 @@ function winnerLabel(s: ScenarioComparison): string {
...
     '',
+  ];
+
+  if (!isFullyJudged && judgeCoverage) {
+    lines.push(
+      `> ⚠️ **Only ${judgeCoverage.judged}/${judgeCoverage.total} scenarios were judged.** The verdict above rests on the`,
+      '> remaining scenarios\' assertion results alone, so treat it as provisional rather than',
+      '> as evidence the skill makes no difference. Re-run to get full coverage.',
+      '',
+    );
```
hadarzaguri, line 154: I'm not sure what devs will do with this info, but we can include it and see if anyone reaches it

hadarzaguri (approved):

PR #1072, opened 2026-08-20
feat(wix-manage): restaurants-reservations-setup recipe

zahorovskyi:
Adds a dedicated recipe for the **Wix Table Reservations** app (`f9c07de2`), which is independent of the Menus and Orders apps and had no recipe coverage.

## What it covers
- Party size limits, party and seat pacing, turnover times per party size
- Manual approval of online requests
- The reservation business schedule
- Creating reservations through their lifecycle — `HELD → REQUESTED → RESERVED → SEATED → FINISHED`, exits `DECLINED`/`CANCELED`/`NO_SHOW`, `source` (`ONLINE`/`OFFLINE`/`WALK_IN`) deciding the starting status
- The eCommerce-order link for payment-required reservations (per-guest fee, deposit, card guarantee) and the no-show / late-cancellation fee path

## Why a separate PR
The Table Reservations app is self-contained, so this recipe can land independently. It is also part of a larger Restaurants recipe split (#1044), but isolated here as a single new skill so it can be reviewed and merged on its own. The cross-links to other recipes in #1044 resolve once that PR lands; until then they are relative paths to files not yet on `main`.

## Eval
Adds `restaurants/set-reservation-rules` — configures an existing reservation location's party size, turnover, and manual approval, without creating a reservation location and without sending deprecated field spellings (`partySize`/`partiesSize`, `turnoverTimeRules`/`turnoverRules`).

## Premium-gate fix (second commit)
The original Step 2 PATCH example sent `onlineReservationsEnabled: true` unconditionally. On a Free-plan site that write is gated — `428 PREMIUM_ONLY` ("Can't turn on online reservation for a non-premium website") — even when a `GET` already returns `true`, and echoing it back re-triggers the premium check so the rule fields never land. The merchant asked for party size, turnover, and approval only, never to toggle online reservations on/off.

- **Recipe**: dropped the flag from the example, added a premium-gate note, and added a `PREMIUM_ONLY` error-handling row (re-send only the rule fields, or stop and report a Premium plan is required when the merchant explicitly wants online reservations on).
- **Eval**: added a pass criterion that the agent must not send `onlineReservationsEnabled`, a matching fail criterion (flag echo / unchanged retry), and a smoothness-judge penalty for the `PREMIUM_ONLY` 428.

General lesson surfaced: patch only the fields the merchant asked for — do not echo the whole `GET` object back, since re-sending a premium-gated flag fails the entire `PATCH`.

### DIFF
--- a/skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
+++ b/skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
@@ -69,7 +69,6 @@ Everything a merchant means by "our booking rules" lives on the reservation loca
     "revision": "4",
     "configuration": {
       "onlineReservations": {
-        "onlineReservationsEnabled": true,
         "partySize": { "min": 1, "max": 10 },
         "defaultTurnoverTime": 90,
         "approval": { "mode": "MANUAL" }
@@ -80,6 +79,8 @@ Everything a merchant means by "our booking rules" lives on the reservation loca
 ```
 
 `defaultTurnoverTime` is minutes and applies to every party; `turnoverTimeRules` overrides it per party size, which is what "90 minutes for a table of four" means. Pacing reads `timeSlotInterval` (15 by default), so party and seat pacing are per 15-minute slot. `minimumReservationNotice` is `{ number, unit }`, e.g. 30 `MINUTES`.
+
+**Do not echo `onlineReservationsEnabled` back unless the merchant is asking to turn online reservations on or off.** Writing `true` to it is gated on a Premium plan — on a non-premium site it returns `428 PREMIUM_ONLY` (`"Can't turn on online reservation for a non-premium website"`), even when the field already reads `true` on a `GET`. A `GET` returns the flag; sending it straight back re-triggers the premium check and fails the whole `PATCH`, so the rule fields never get written. Patch only the rules the merchant asked for (party size, turnover, approval, pacing, schedule) and leave `onlineReservationsEnabled` at its existing value. If the merchant explicitly wants online reservations turned on and the call comes back `PREMIUM_ONLY`, say plainly that a Premium plan is required and stop — do not retry the same body.
 
 ### Step 3: Check Availability Before Booking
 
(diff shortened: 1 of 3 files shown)


skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
```diff
@@ -0,0 +1,153 @@
...
+- Setting minimum and maximum party size, or how long a table is held
+- Limiting how many parties or guests can be seated in the same 15 minutes
+- Requiring staff approval before an online booking is confirmed
+- Setting the hours reservations can be booked for
+- Creating a booking, seating a party, cancelling, or marking a no-show
+- Unsure which app owns the request? Start from [Configure Restaurants from Prompt](configure-restaurants-from-prompt.md)
+
+## Prerequisites
+
+1. Wix Table Reservations installed on the site (`f9c07de2-5341-40c6-b096-8eb39de391fb`) — it is independent of the ordering apps, so a site that takes online orders may not have it
```
hadarzaguri, line 21: ordering apps = menus + OLO? if so - not clear.

skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
```diff
@@ -0,0 +1,153 @@
...
+2. **At least one location configured in Business Info.** Every other call depends on it
```
hadarzaguri, line 22: what does it mean one location configured, i think worth to explain - how an agent can check that? relevant api

skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
```diff
@@ -0,0 +1,153 @@
...
+3. API access with reservations permissions
```
hadarzaguri, line 23: what reservations permissions means?

skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Two Rules That Decide Whether This Works
+
+### Rule 1: reservation locations are not created here
+
+A reservation location represents one physical restaurant and holds its calendar, pacing, and table rules. It can only be **created or archived through the Dashboard or the Locations API** — never through the Reservations APIs. Read the existing ones and update them.
+
+The same call also refuses to change a location's `location` object. Address changes belong to the Locations API.
+
+If listing returns nothing, the site has no location configured in Business Info yet. Say so rather than trying to create one.
```
hadarzaguri, line 33: Better to share an example - an agent doesnt know what locations API mean...

skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Rule 2: several settings have two spellings, and sending both fails
+
+The server keeps a newer and a deprecated field in sync, and a `GET` returns **both** spellings:
+
+| Use this | Not this |
+|---|---|
+| `partySize` | `partiesSize` |
+| `turnoverTimeRules` | `turnoverRules` |
+
```
hadarzaguri, line 43: why not to share a request example instead..?

skills/wix-manage/references/restaurants/restaurants-reservations-setup.md
```diff
@@ -0,0 +1,153 @@
...
+Echoing the whole object back therefore sends both, and if their values disagree the request is rejected. Send only the newer field, and strip the deprecated twin from anything you read before writing it back. Manual approval is a third shape rather than a twin — `approval.mode`, either `"AUTOMATIC"` or `"MANUAL"`.
+
+## Flow
+
+### Step 1: Find the Reservation Location
+
+List or query reservation locations and pick one. Sites with several restaurants have one per physical location, so match on the location before changing anything.
```
hadarzaguri, line 50: how? share a code example

PR #1075, opened 2026-08-20
feat(seo): add SEO tags management recipe

YZaritsky:
## What

Adds one `wix-manage` recipe for the public **SEO APIs** (SEO Metatags service): **Manage a Wix Site's SEO Tags**.

The API is GA, public, routed, and documented on dev.wix.com since 2026-08-11 (REST at `www.wixapis.com/seo-metatags-server/v1/*`). This PR adds the missing agent-consumer artifact so Wix MCP and BM Aria can execute the user flow.

Docs entry: https://dev.wix.com/docs/api-reference/business-management/seo

## The recipe teaches

- Choosing the right level for a change: site-wide tags (Site SEO Tags), page-type patterns (SEO Patterns), or one item's tags (Item SEO Tags)
- Discovery over invention: item IDs via List Item SEO Tags, pattern vocabulary via List SEO Pattern Variables, supported item types from the `UNSUPPORTED_ITEM_TYPE` error
- Read-before-write: every write is a full replace; reset methods instead of empty sets; verification read after write
- Static-page draft vs published semantics, primary-language-only writes, bulk partial-failure handling via `originalIndex`
- Stop after the first 403 and name the missing **Manage SEO Settings** authorization

## Files (follows the accessibility precedent, #834)

- `skills/wix-manage/references/seo/manage-seo-tags.md` — the recipe
- `skills/wix-manage/SKILL.md` — new SEO section + routing entry
- `yaml/wix-manage/seo/documentation.yaml` — docs mapping
- `yaml/wix-manage-evals/seo/manage-seo-tags.yml` — eval: item-level title/description update on a seeded store product, with coverage, correctness, and path-quality assertions

## Notes for reviewers

- One skill in this PR (limit is 4).
- The service's per-endpoint permission rename (flynt `permissions-naming`) is merged but not yet deployed to production `main` pods; the recipe references the human-readable permission name shown in the live docs ("Manage SEO Settings"), which is stable across the rename.
- Opened as draft for owner review; EvalForge run needed before merge.

### DIFF
--- a/yaml/wix-manage-evals/seo/manage-seo-tags.yml
+++ b/yaml/wix-manage-evals/seo/manage-seo-tags.yml
@@ -1,0 +1,69 @@
+name: seo/manage-seo-tags
+description: >-
+  Verifies that the agent updates one store product's SEO tags at the item level: it discovers the
+  real item ID, reads the current tags before writing, sends the complete tag set back on the
+  full-replace write, and confirms the result by reading resolved tags.
+triggerPrompt: >-
+  On my Wix site, set the SEO title of my product "Accessibility Scan Test Product" to
+  "Handmade Ceramic Mug | Studio Shop" and its meta description to
+  "A stoneware mug thrown and glazed by hand in our studio." Then show me what that product
+  page will actually render with.
+tags: [aria, stores]
+assertions:
+  - tool: ReadFullDocsArticle
+    params:
+      articleUrl: https://dev.wix.com/docs/api-reference/business-management/seo/skills/manage-a-wix-sites-seo-tags
+  - type: llm_judge
+    minScore: 7
+    maxTokens: 2048
+    prompt: |
+      The user asked to set one store product's SEO title and meta description, then see what the
+      product page will render with.
+
+      Pass if the agent:
+      - works at the item level (Item SEO Tags) rather than rewriting a page-type pattern or the
+        site's tags for a single-product change,
+      - discovers the product's real item ID (by listing item SEO tags for the product item type or
+        resolving the product through its own API) instead of fabricating an ID or asking for a
+        site ID,
+      - reads the item's current tags before writing, then sends the complete tag set (title AND
+        description together) in one set call, since a write replaces the item's tags in full,
+      - leaves the language unset on the write, and
+      - reports the outcome from resolved tags, naming the source of each tag, without claiming
+        tags it did not verify.
+
+      Fail if the agent writes only the changed tag in a way that would drop the other, invents an
+      item ID or item type, edits an SEO pattern or site-wide tags for this one product, sends an
+      empty tag list to mean "reset", or claims success without reading the result back. If any
+      write returns 403 or PERMISSION_DENIED, require the agent to stop after that first attempt
+      and clearly identify the missing Manage SEO Settings authorization; do not reward retries
+      with alternate items, levels, request shapes, or sites.
+  - type: llm_judge
+    minScore: 7
+    maxTokens: 2048
+    prompt: |
+      Rate how smoothly the Wix MCP let the agent fulfill the SEO tag update. Examine the tool-call
+      trace. Score 0-10 (10 = direct: item discovery, one read, one complete-set write, one
+      verification read, no wasted calls or errors). Penalize and list any guessed item ID or item
+      type, request shape guessed wrong then corrected, redundant probing call, write issued
+      without first reading the current tags, duplicate write, or recovered non-2xx error. For each
+      problem, name the likely skill, MCP, or docs gap.
+      Return ONLY {"score":<0-10>,"reason":"<terse list of frictions and suspected gap, or 'clean path' if none>"}.
+siteSetup:
+  mode: template
+  templateId: stores-v3-editor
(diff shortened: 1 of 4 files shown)


YZaritsky:
@ayal could you please review and approve when you have a moment? All checks pass.

The EvalForge gate ran on the latest commit and the recipe wins with high confidence:

- **With the recipe:** all assertions pass — 10/10 correctness, 10/10 path quality ("clean path")
- **Without it (prod today):** 4/10 correctness and 5/10 path quality — the agent skipped reading current tags before a full-replace write (which silently drops the item's other tags) and didn't name resolved-tag sources
- 40% fewer tokens and 30% faster with the recipe

Same shape as the accessibility scan recipe in #834: one skill, its eval scenario, the documentation.yaml mapping, and a routing entry in `wix-manage/SKILL.md`. Thanks!

yaml/wix-manage/seo/documentation.yaml
```diff
@@ -0,0 +1,8 @@
+apiDoc:
+  title: Wix SEO Management Recipes
+
+  docs:
+    - file: ../../../skills/wix-manage/references/seo/manage-seo-tags.md
+      title: Manage a Wix Site's SEO Tags
+      docsEntry: https://dev.wix.com/docs/api-reference/business-management/seo
+      tags: [aria, stores]
```
danshw: why these tags?

yaml/wix-manage-evals/seo/manage-seo-tags.yml
```diff
@@ -0,0 +1,69 @@
...
+description: >-
+  Verifies that the agent updates one store product's SEO tags at the item level: it discovers the
+  real item ID, reads the current tags before writing, sends the complete tag set back on the
+  full-replace write, and confirms the result by reading resolved tags.
+triggerPrompt: >-
+  On my Wix site, set the SEO title of my product "Accessibility Scan Test Product" to
+  "Handmade Ceramic Mug | Studio Shop" and its meta description to
+  "A stoneware mug thrown and glazed by hand in our studio." Then show me what that product
+  page will actually render with.
+tags: [aria, stores]
```
danshw: why these tags? is this aria specific?

yaml/wix-manage-evals/seo/manage-seo-tags.yml
```diff
@@ -0,0 +1,69 @@
...
+      item ID or item type, edits an SEO pattern or site-wide tags for this one product, sends an
+      empty tag list to mean "reset", or claims success without reading the result back. If any
+      write returns 403 or PERMISSION_DENIED, require the agent to stop after that first attempt
+      and clearly identify the missing Manage SEO Settings authorization; do not reward retries
+      with alternate items, levels, request shapes, or sites.
+  - type: llm_judge
+    minScore: 7
+    maxTokens: 2048
+    prompt: |
+      Rate how smoothly the Wix MCP let the agent fulfill the SEO tag update. Examine the tool-call
```
danshw, line 45: looking at the eval run 
https://bo.wix.com/pages/evalforge/926d0e11-32bb-4d85-b302-e0b9a5cf40e9/results?runId=979b8e8c-5ab8-4a5e-9492-99dd897672f4&name=draft%3Awix%2Fskills%231075%20%5BMCPs%3A%20mcp-with-draft%3Awix%2Fskills%231075%5D

The agent is doing redundant probing calls, including three documentation searches and two article reads, before executing the API.
can we maybe add some examples to the recipe so the agent doesn't need to look for the docs?

yaml/wix-manage/seo/documentation.yaml
YZaritsky: Copied from the accessibility recipe without thinking it through — you are right. `aria` is not an area tag (only that one recipe uses it) and `stores` is not this area either. Changed to `[seo]` in `46f4c39`, matching the one-tag-per-area convention every other documentation.yaml follows.

yaml/wix-manage-evals/seo/manage-seo-tags.yml
YZaritsky: Same mistake, same origin. Not Aria-specific — nothing here needs Aria. Now `[seo, stores]`: `seo` is the area, and `stores` stays only because the scenario provisions a `stores-v3-editor` site and seeds a product, so a Stores change should re-run it. Happy to drop `stores` if you would rather keep it to one tag.

yaml/wix-manage-evals/seo/manage-seo-tags.yml
YZaritsky: Good catch, and the fix was cheap. The recipe now carries the endpoints inline so there is nothing to search for: a Level/Method/Call table with all 12 paths relative to `https://www.wixapis.com/promote/seo/v1`, plus a `Set` request example and a bulk-write example, and a note on the `{type, props, children}` tag shape. Verified every path against the reference docs — that caught two I had guessed wrong (`/reset-to-default`, not `/reset`). In `46f4c39`; the gate is re-running, so we will see whether the doc searches drop out of the trace.

danshw (approved):

PR #1076, opened 2026-08-20
feat(wix-manage): add a SEO recipe for managing site URL redirects

talni5:
Adds the first **SEO** area to `wix-manage`, covering the public SEO Redirects API (`wix.promote.seo.v1.redirect`), published to dev.wix.com on 2026-08-20. Six public methods: List, Get, Create, Bulk Create, Delete, Bulk Delete.

Four files per `CONTRIBUTING.md`, plus `SEO` added to the `SKILL.md` routing description so the area is discoverable:

- `skills/wix-manage/references/seo/manage-url-redirects.md`
- `skills/wix-manage/SKILL.md` — index entry + routing list
- `yaml/wix-manage/seo/documentation.yaml`
- `yaml/wix-manage-evals/seo/manage-url-redirects.yml`

## The recipe leads with the destructive behaviour, not the happy path

Two writes on this API delete data without asking, and neither is recoverable:

- Creating a redirect that points at a path another redirect **starts from** deletes that other redirect and proceeds. No confirmation is asked by the API.
- `options.forceReplace` deletes whatever holds the `from` path instead of failing with `FROM_URL_EXISTS`.

So the recipe requires listing existing redirects and confirming before any create, and **forbids the agent setting `forceReplace` on its own initiative** — it may only offer it after reporting the conflict.

## Other traps it encodes

All verified against the live API rather than read off the schema:

| Trap | Why an agent gets it wrong |
|---|---|
| No update method, no query method | Changing a redirect is get → delete → create carrying the **whole** redirect; an omitted `options` turns a group redirect into an exact one, an omitted `language` turns a language-scoped one global |
| Group redirects carry the URL remainder | `/forum/questions/` → `/forum/faqs/` sends `/forum/questions/my-post` → `/forum/faqs/my-post` |
| Bulk failures arrive inside a **200** | on `results[].itemMetadata.error`; `undetailedFailures` means outcome **unknown**, not failed |
| Bulk delete "success" means sent | not confirmed deleted |
| Bulk create is not atomic | a deleted conflicting redirect stays deleted if the write then fails |
| Language prefixes are stripped | `/fr/about` with `language: "fr"` comes back as `/about` |
| Re-creating an identical redirect succeeds with no `id` | that is not a newly created redirect |
| No events on any write | absence of an event is not absence of a change |

Also: a redirect takes precedence over a real page at the same path, so creating one over a live page makes that page unreachable. The recipe says so.

## Notes for review

**Tags are `[aria, site-properties]`, not a new `seo` tag.** No `seo` tag exists anywhere in `yaml/wix-manage-evals` today, and `CONTRIBUTING.md` asks for an existing tag. The `accessibility` recipe sets the precedent that tags need not match the folder — it lives in `accessibility/` tagged `[aria, stores]`. Happy to switch if `seo` is registered on your side.

**The coverage assertion was verified, not hand-built.** I ran this repo's own `slugify` from `.github/actions/evalforge-yaml-gate/src/utils/doc-url.ts` against the title and compared it to the URL in the eval — exact match on `…/seo/redirects/skills/manage-url-redirects-on-a-wix-site`.

**One documented divergence.** The reference promises `REDIRECT_NOT_FOUND` for an unknown id on Get Redirect and Delete Redirect; the API currently returns a generic 404 without it. Rather than teach the agent a code it will not receive, the recipe says to treat any 404 on those two methods as "no redirect with that id". Tracked on the owning team's side (`wix-private/promote-seo#15085`), and the recipe can be simplified once the handler maps it.

## Not yet proven

These need the shared MCP's own app identity rather than the author's, so I could not verify them from here and would value a pointer if you check them as part of review:

- method discovery through the shared Wix MCP
- one authorized write through `CallWixSiteAPI` with that identity
- a real Business Manager Aria request selecting this recipe

Editor/Harmony is deliberately out of scope: redirects already have a native Business Manager panel, and the Accessibility precedent kept its native wizard rather than duplicating the API flow in the assistant.

### DIFF
--- a/skills/wix-manage/references/seo/manage-url-redirects.md
+++ b/skills/wix-manage/references/seo/manage-url-redirects.md
@@ -1,0 +1,119 @@
+---
+name: "Manage URL Redirects on a Wix Site"
+description: Retrieve, create, and delete URL redirects on a Wix site using the public SEO Redirects API. Covers exact and group redirects, language-scoped redirects for multilingual sites, batches of up to 500, and the change flow for a redirect that already exists. Creating a redirect can permanently delete another one, so always confirm before writing.
+---
+
+# Manage URL Redirects on a Wix Site
+
+Use the public **SEO Redirects API** to manage where a Wix site sends visitors who
+request an old or moved URL. The API selects the site from the caller's
+authorization context; never ask for or send a site ID.
+
+A redirect returns a 301 permanent redirect, takes effect on the live site
+immediately with no site publish, and **takes precedence over a real page at the
+same path**. Creating a redirect from a path that still serves a page makes that
+page unreachable until the redirect is deleted.
+
+## Before writing anything
+
+Two behaviors make writes destructive in ways the user will not expect. Confirm
+the specific paths with the user before any create, and say which of these
+applies:
+
+- **Loops.** If the new redirect points at a path that an existing redirect
+  starts from, that existing redirect is **deleted** and the create proceeds. No
+  separate confirmation is asked, and the deletion cannot be undone.
+- **A taken `from` path.** Create Redirect fails with `FROM_URL_EXISTS` and
+  writes nothing. `options.forceReplace` makes it **delete** the redirect holding
+  that path instead. Never set `forceReplace` on your own initiative: offer it
+  only after reporting the conflict, and only if the user asks to replace.
+
+Before a create that might hit either case, call **List Redirects** and check the
+site's existing redirects. If something will be deleted, name it and ask.
+
+## Choose exact or group
+
+| User intent | Setting |
+|---|---|
+| "Redirect this one URL" | omit `options`, or `options.groupRedirect: false` |
+| "Redirect this section", "everything under /blog" | `options.groupRedirect: true` |
+
+A group redirect carries the rest of the URL over to the target: from
+`/forum/questions/` to `/forum/faqs/` sends `/forum/questions/my-post` to
+`/forum/faqs/my-post`. An exact and a group redirect that share a `from` path are
+two different redirects.
+
+For a multilingual site, set `language` to apply the redirect to one language
+only; omit it to apply to every language. A language-scoped path is stored
+**without** the language prefix, so `/fr/about` with `language: "fr"` comes back
+as `/about`. Deleting a language-scoped redirect stops it working but it can
+still appear in List Redirects.
+
+`from` cannot be the site root, and two paths differing only by a trailing slash
+count as the same path.
+
+## Read
+
+- **List Redirects** returns every redirect on the site in one response. It takes
+  no filter, sort, or paging, so filter the result yourself. It also returns
+  redirects Wix created for the site owner, such as when a page's URL slug was
+  renamed in the editor.
+- **Get Redirect** retrieves one by ID.
+
+## Change an existing redirect
+
+There is **no update method and no query method**. To change a redirect:
+
+1. Call **Get Redirect** and keep the whole response.
+2. Call **Delete Redirect** with the same ID.
+3. Call **Create Redirect** with every field you retrieved, with your change applied.
+
+Carry the whole redirect across, not only what changed. Create Redirect fills an
+omitted field with its default rather than the previous value, so dropping
+`options` turns a group redirect into an exact one and dropping `language` turns a
+language-scoped redirect into a global one. Keep `id` to preserve the redirect's
+identity, or omit it for a new one.
+
+Confirm with the user before step 2: between the delete and the create the
+redirect is not in effect, and if the create fails the old redirect is gone.
+
+## Batches
+
+**Bulk Create Redirects** and **Bulk Delete Redirects** take 1 to 500 items and
+report each one separately **inside a successful response**:
+
+- Read every outcome from `results[].itemMetadata`, matched to your request by
+  `originalIndex`. A failed item carries `error.code`, such as `FROM_URL_EXISTS`
+  or `REDIRECT_NOT_FOUND`.
+- `bulkActionMetadata.undetailedFailures` counts items whose outcome is
+  **unknown**: they carry no error and may or may not have been written. Call
+  List Redirects to find out. Never report them as successes.
+- A malformed request, such as an empty list, is rejected as a whole with a 400
+  and returns no per-item results.
+
+Two more sharp edges to report truthfully:
+
+- A successful **Bulk Delete** item means the deletion was found and sent, not
+  confirmed. Call List Redirects if the user needs confirmation.
+- Bulk Create is **not atomic**. If an item's conflicting or loop-causing
+  redirect is deleted and the write then fails, the deleted redirect is gone and
+  nothing replaces it. Repeating the same request recreates it.
+
+Creating a redirect identical to one already on the site changes nothing and
+succeeds: Create Redirect echoes the request back without an `id` or
+`createdDate`, and Bulk Create reports success with no `id`. Do not present that
+as a newly created redirect.
+
+## Stop conditions
+
+- **On the first `403` or `PERMISSION_DENIED`, stop.** Report the missing
+  authorization for managing SEO settings. Do not retry with another site, path,
+  request shape, or method.
+- On a 400, read `details.validationError.fieldViolations` and report the named
+  field rather than guessing.
+- **Known divergence:** the reference documents `REDIRECT_NOT_FOUND` for an
+  unknown redirect ID on Get Redirect and Delete Redirect. The API currently
+  returns a generic 404 without that code. Treat any 404 on those two methods as
+  "no redirect with that ID", and do not tell the user the ID was valid.
+- No webhook or event is emitted when a redirect is created or deleted. Never
+  treat the absence of an event as the absence of a change.
--- a/skills/wix-manage/references/seo/manage-url-redirects.md
+++ b/skills/wix-manage/references/seo/manage-url-redirects.md
@@ -58,6 +58,98 @@ redirect is not the same as the user agreeing to lose a different one.
 **deletes** the redirect holding it. Never set it on your own initiative. Offer it
 only after reporting the conflict, and only if the user asks to replace.
 
+## The six methods and their request shapes
+
+Every path is under `https://www.wixapis.com/seo-redirects-service/v1`. Use these
+directly - do not go looking for the shapes in the docs first.
+
+| Method | Call |
+|---|---|
+| List Redirects | `GET /redirects` - no body, no parameters |
+| Get Redirect | `GET /redirects/{redirectId}` |
+| Create Redirect | `POST /create-redirect` - body below |
+| Delete Redirect | `DELETE /redirects/{redirectId}` - empty response |
+| Bulk Create Redirects | `POST /bulk/redirects/create` |
+| Bulk Delete Redirects | `POST /bulk/redirects/delete` |
+
+Create Redirect takes the redirect nested under `redirect`. `options` and
+`language` are optional, `id` only when preserving an existing redirect's
+identity:
+
+```json
+{
+  "redirect": {
+    "from": "/old-blog",
+    "to": "/blog",
+    "options": { "groupRedirect": true },
+    "language": "fr"
+  }
+}
+```
+
+The two bulk methods take flat lists. Bulk Create's `returnFullEntity` returns
+each created redirect in `results[].item`:
+
+```json
+{ "redirects": [ { "from": "/old-pricing", "to": "/pricing" } ], "returnFullEntity": true }
+```
+
+```json
+{ "redirectIds": ["5b07d9de-6a9f-494d-8066-a7b66c970270"] }
+```
+
+## Worked example: the request that deletes something
+
+The user asks to send `/old-blog` and everything under it to `/blog`.
+
+**1. List first.**
+
+```
+GET https://www.wixapis.com/seo-redirects-service/v1/redirects
+```
+
+```json
+{
+  "redirects": [
+    {
+      "id": "583fe7d7-2944-436f-870a-09eb6d2e52fa",
+      "from": "/blog",
+      "to": "/news",
+      "createdDate": "2026-08-20T09:12:44.000Z"
+    }
+  ]
+}
+```
+
+**2. Compare.** The new `to` is `/blog`, and an existing redirect's `from` is
+`/blog`. That is the deletion case. **End the turn here** - report that creating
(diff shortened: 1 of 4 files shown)


talni5:
Pushed a fix for the gate result. Both sides failed the same way — the agent stopped at turn one to ask which site to work on, so no redirect work happened and neither judge was reachable. Two causes, one in the scenario and one in the recipe.

**The scenario had no `siteSetup`.** It ran against the shared multi-site account, where asking which site is the *correct* behaviour for an agent. The accessibility scenario provisions a site for exactly this reason; mine now does too, on `blank-editor`.

**The bootstrap step is the scenario, not scaffolding.** It seeds `/blog` → `/news`. The requested redirect *targets* `/blog` and the seeded one *starts* at `/blog`, so creating it closes a loop and the API deletes the seeded redirect without asking. That is the destructive behaviour this recipe exists to make visible — and with nothing seeded, the "does this change any of them" half of the request had nothing to find. The correctness judge now requires the agent to connect the two: listing the redirect without naming the consequence is not an answer.

**The recipe's site rule was unactionable.** It said "never ask for or send a site ID", inherited from the accessibility recipe. Read literally by an agent facing a site-selection prompt, it leaves no way forward — which is what the prod run's quality judge scored 10/10 for: pausing to ask, having done nothing. Replaced with the real rule: use the site the run identifies, ask only when several are available and nothing picks one, never ask for a site ID the API doesn't take.

**The two judges also contradicted each other**, which is worth flagging beyond my scenario: prod scored 0/10 on correctness and 10/10 on quality for the *same* behaviour. The quality judge now explicitly fails an agent that asks which site, since the run provisions exactly one and supplies its ID.

One thing I could not verify locally and would value a second pair of eyes on: whether `bootstrap.steps` can call `https://www.wixapis.com/seo-redirects-service/v1/create-redirect` with the run's identity. The API is public and GA'd on the portal as of 2026-08-20, but bootstrap is fail-fast on a non-2xx, so if that identity lacks the SEO edit permission the run will fail at setup rather than at the assertion.

talni5:
Second fix pushed. The site question is resolved and the PR side now loads the recipe (`Tool called with param` passed), so this round's failures were the recipe's content rather than the scenario. The judges named three things and all three were real gaps:

**1. "incorrectly claimed a redirect chain was formed"** — the most instructive one. The agent invented a plausible mental model, `/old-blog` → `/blog` → `/news`, because the recipe never said redirects don't chain. It described an outcome the API had just destroyed. That now has its own section at the top with the concrete before/after, and the correctness judge fails a chain description explicitly — it's a wrong answer *even when the create succeeded*, because the user is told a redirect still exists that the API removed.

**2. "created the redirect without asking the user first"** — my confirmation rule was prose an agent could reasonably read as already satisfied by an explicit request. Replaced with a check it can execute:

> list → compare the new redirect's `to` against every existing `from` (that one gets deleted) → compare its `from` against every existing `from` (that create fails with `FROM_URL_EXISTS`) → trailing-slash-only differences count as equal → if either matched, stop and ask; if neither, proceed without asking.

The conditional shape is deliberate. A blanket "always confirm" is both worse UX and easier to ignore, which is what happened. Asking is now tied to a computable condition, and the recipe says why the explicit request doesn't settle it: asking for a redirect is not agreeing to lose a different one.

**3. "retried a failed write with a mutated request shape"** — now forbidden by name (different shape, different path, `forceReplace` added, bulk in place of single), with the reason: a failed create wrote nothing, so a mutated retry is a second attempt at a decision the user never agreed to, not a recovery.

Worth separating from my changes: the run costs 45% more with the draft than without ($0.47 vs $0.32, 395K vs 278K tokens). Some of that is the recipe doing more work — the List-before-write check is a real extra call — but the recipe also grew, and if the gate treats efficiency as a tiebreaker I'd rather trim it deliberately than have it counted against the change. Happy to cut the Batches and Change-an-existing-redirect sections down if you'd prefer a leaner reference.

talni5:
Third round. The recipe is now doing its job — correctness scored **10/10 with the draft vs 0/10 without**, and the pairwise judge picks the PR with high confidence on task completion, tool accuracy and error handling:

> "perfectly identified the existing redirect, correctly explained that creating the new group redirect would delete it due to Wix's non-chaining behavior, and asked the user for confirmation before making any changes"

versus prod, which "created the redirect without asking first, missed that the existing redirect would be deleted, and incorrectly described the outcome as a redirect chain." That last clause is the exact failure mode this recipe was written for, reproduced on the control side.

**One assertion still fails, and it's the same class of gap:** *"attempting a query method on an API that does not support it."*

Nearly every Wix REST API offers `POST .../query`, so an agent reaches for one by default. I had said "there is no query method" — but once, buried in a section about *changing* a redirect, not where an agent looks for reads. Now the Read section is titled for it, states that no query or search endpoint exists, says why an agent expects one, and names the exact paths not to construct.

That's the same technique that fixed the chain claim last round: **state the plausible-but-wrong model explicitly, rather than only documenting what's true.** A reference that lists correct behaviour leaves the wrong intuition in charge of everything it doesn't mention. Two rounds, two misconceptions, both fixed by naming them — that may be worth generalising into `CONTRIBUTING.md`'s "Writing API Skills" guidance if you agree it's a pattern.

On cost: the gap narrowed to 21% ($0.38 vs $0.32) from 45% last round, and the remaining difference is mostly the List-before-write call the recipe requires — which is the safety property, not overhead.

Two questions for you, since `required: false` means the no-winner verdict isn't blocking:

1. Is the quality judge's bar right? It fails the run for a *single* speculative query attempt that the agent then recovered from. I think that's correct discipline for a destructive API and I'd rather keep it strict, but it's your rubric convention — say the word and I'll soften it to "repeated or unrecovered speculative calls".
2. If this round comes back with correctness passing and only that quality assertion outstanding, do you want it merged as-is, or should I keep iterating until both are green?

talni5:
Round 4 regressed on the rule round 3 passed: correctness **10/10 → 3/10** with essentially the same confirmation instruction — *"created the redirect and deleted the existing one before reporting the conflict and asking the user for permission."*

That changes my diagnosis, so I want to be straight about it rather than quietly pushing a fifth prose tweak.

**Rounds 1–3 each found a real, nameable gap** — no provisioned site, the redirect-chain misconception, the assumed `POST /query` — and naming each one fixed it. **Round 4 found nothing new.** The same prose produced compliance once and a destructive write the next time. That's an instruction being read and then not followed, not misunderstood.

**What I changed, once:** the gate no longer asks for intent, it states a structural rule — *never put a create in the same response as the List Redirects that found the conflict.* An agent can't reliably self-assess "did I confirm", but it can check "is my next action after listing a write". The judge now fails a same-turn create even when the response also reports the conflict, because a user cannot answer a question that was already acted on.

I also made step 4 explicit that when nothing matched, the create proceeds in the same turn **without** asking. Part of why "always confirm" fails is that it's indistinguishable from "confirm pointlessly", and an agent that learns the rule costs a turn for nothing will drop it.

**Worth keeping from the bad run:** with the recipe the agent correctly reported that the API had deleted the existing redirect. Prod claimed it was unmodified — a false account of the user's site. The pairwise judge picked the PR on exactly that basis, high confidence. Wrong process with a truthful report beats wrong process with a wrong report, and that difference is already worth shipping.

**If this recurs, the fix isn't prose, and I'd rather say so now than iterate to ten rounds.** A reliable pre-write stop on a destructive API needs enforcement a Markdown recipe cannot provide. Three options, in the order I'd pick them:

1. **`dryRun` on Create Redirect** — the agent can show the user exactly what would be deleted, with no trust required. Cleanest, and it's an API change on our side (`wix-private/promote-seo`).
2. **A create that refuses to delete unless an explicit flag is set** — makes the destructive path opt-in at the contract level rather than at the prose level.
3. **A destructive-tool annotation on the MCP side**, so the client asks rather than the model choosing to.

Given that, two decisions are yours:

- **Merge on the pairwise win?** `required: false`, the PR beats prod with high confidence on task completion and tool accuracy, and the doc-coverage assertion passes only with the draft. The recipe is a strict improvement over no recipe even in its worst observed run.
- **Or hold** until both judges are green, in which case I think option 1 above is the actual path and this PR should wait on an API change rather than more wording.

I'm also happy to relax my own quality judge if you think a single recovered speculative call shouldn't sink a run — I deliberately set it strict for a destructive API, but it's your rubric.

yaml/wix-manage-evals/seo/manage-url-redirects.yml
```diff
@@ -0,0 +1,89 @@
+name: seo/manage-url-redirects
+description: >-
+  Verifies that the agent reads the site's existing redirects before writing, recognizes that the
+  requested group redirect would silently delete a seeded redirect by closing a loop, asks the user
+  before creating it, and never sets forceReplace on its own initiative.
+triggerPrompt: >-
+  On my Wix site, send everyone from /old-blog and everything under it to /blog. Tell me what
+  redirects I already have and whether this changes any of them.
+tags: [aria, site-properties]
```
danshw: shouldn't the tag be "seo"? why "aria"?

danshw:
suggestion - add a working example to the skill so the agent doesn't need to search the docs for the request shape

skills/wix-manage/references/seo/manage-url-redirects.md
```diff
@@ -0,0 +1,158 @@
+---
+name: "Manage URL Redirects on a Wix Site"
+description: Retrieve, create, and delete URL redirects on a Wix site using the public SEO Redirects API. Covers exact and group redirects, language-scoped redirects for multilingual sites, batches of up to 500, and the change flow for a redirect that already exists. This API has no query, search, or update method: List Redirects is the only read-many. Redirects do not chain, so creating one that points at a path another redirect starts from permanently deletes that other redirect; list and check before every write.
```
danshw: wrap the description in quotes

yaml/wix-manage/seo/documentation.yaml
```diff
@@ -0,0 +1,8 @@
+apiDoc:
+  title: Wix SEO Management Recipes
+
+  docs:
+    - file: ../../../skills/wix-manage/references/seo/manage-url-redirects.md
+      title: Manage URL Redirects on a Wix Site
+      docsEntry: https://dev.wix.com/docs/api-reference/business-management/seo/redirects
+      tags: [aria, site-properties]
```
danshw: why these tags?

yaml/wix-manage-evals/seo/manage-url-redirects.yml
talni5: You're right, changed to `tags: [seo]` in both the eval and `yaml/wix-manage/seo/documentation.yaml`.

I copied `[aria, site-properties]` from the accessibility recipe, which is the one area whose tag doesn't match its folder. Checking the rest: 23 of 24 areas tag by their own folder name, and nothing in the gate validates a tag against an allowlist, so `seo` registers itself on first use. No reason to inherit the exception.

talni5:
Both comments addressed in `5d1…` (pushed).

**Tag** → `[seo]` in `yaml/wix-manage-evals/seo/manage-url-redirects.yml` and `yaml/wix-manage/seo/documentation.yaml`. Replied inline: I'd copied the accessibility recipe, which is the only area of 24 that doesn't tag by its folder name.

**Working example** → new sections in `references/seo/manage-url-redirects.md`:

1. **The six methods and their request shapes** — a table of every path under `/seo-redirects-service/v1` (List is `GET /redirects`, Create is `POST /create-redirect`, delete is `DELETE /redirects/{id}`, the two bulk paths), plus the Create body with `options` and `language` nested under `redirect`, and both bulk bodies.
2. **Worked example: the request that deletes something** — the eval's own scenario end to end with real request and response JSON: List returns the seeded `/blog` → `/news`, the comparison finds the deletion case, the turn ends, and the create only runs after the user agrees.

Putting the example on the destructive path rather than on a neutral create is deliberate: the shape and the stop rule are learned in one read, and the agent sees the response that proves `/blog` → `/news` is gone.

Worth watching in the next gate run: the recipe grew, and the last round's cost gap was 21%. If it widens, the Batches and Change-an-existing-redirect sections are the ones I'd trim, since the table now carries their request shapes.

skills/wix-manage/references/seo/manage-url-redirects.md
talni5: Done, quoted in `0db08c6`.

It was more than style — unquoted, that line isn't valid YAML. The description contains `…no query, search, or update method: List Redirects…`, and a plain scalar can't hold a colon-space, so anything parsing the frontmatter breaks on it. Now `description: "…"`, with no double quotes in the text to escape.

yaml/wix-manage/seo/documentation.yaml
talni5: Same answer as the eval one — fixed in `c468971`, this is `tags: [seo]` now.

I'd copied `[aria, site-properties]` from the accessibility recipe. It turns out to be the only one of 24 areas that doesn't tag by its own folder name, so there was no convention to follow there. Nothing in the gate validates tags against an allowlist, so `seo` registers on first use.

talni5:
All three comments are in — `tags: [seo]`, quoted description, worked example — and the gate re-ran green. **Every assertion passes with the draft and fails without it**, for the first time:

| | PR | prod |
|---|---|---|
| Doc coverage (`Tool called with param`) | ✅ | ❌ |
| Correctness judge | ✅ 10/10 | ❌ 0/10 |
| Quality judge | ✅ 10/10 | ❌ 0/10 |
| Cost | **$0.296** | $0.411 |
| Tokens | **200.6K** | 278.0K |
| Time | **36.9s** | 40.5s |

Pairwise: PR wins, high confidence, on task completion, tool accuracy and error handling.

**Your working-example suggestion is what closed the cost gap.** Every earlier round the draft cost *more* than prod — 45%, then 21%, because the recipe made the agent do more work. This round it costs 28% *less* on 28% fewer tokens. Giving it the six endpoints and the request bodies removes the doc-searching turns, and that more than pays for the List-before-write call the recipe requires. Caveat: prod moved too ($0.33 → $0.41 across rounds), so I'd read the direction as solid and the exact percentage as noisy.

That also resolves the two questions I left open earlier:

1. **Quality judge bar** — moot, it passes. Leaving it strict.
2. **Merge as-is or keep iterating** — no reason to iterate. Correctness held at 10/10 across two consecutive rounds, and the round-4 regression didn't recur under the turn-structure rule.

Prod's failure is still the exact one this recipe was written for: *"created the redirect in the same turn as listing the existing ones and incorrectly described the outcome as a redirect chain."*

Ready for your approval from my side. Nothing outstanding.

danshw (approved):

PR #1082, opened 2026-08-21
🦞 igor-claw: add Diagnose Checkout Payment Failures recipe

ayal:
## Summary
- Adds a `wix-manage` recipe for "checkout payment fails" reports, targeting the common shape: **one payment method (Apple Pay/Google Pay) fails while another (cards) succeeds through the same connected provider.**
- Routes the agent to check Cashier's [Reason Codes](https://dev.wix.com/docs/api-reference/business-management/payments/payment-service-provider-service-plugin/reason-codes) first, then cross-check the provider's declared capabilities via [Get Site Payment Method Type](https://dev.wix.com/docs/api-reference/business-management/payments/site-payment-method-types/get-site-payment-method-type) (`features.expressFlow.supported` for wallet/one-tap payments) **before** concluding it's a Wix platform bug.

## Why
While researching a live merchant report (PayPlus checkout, Wix Cashier reason code `1006` "Request is not allowed" on every `WalletPaymentMethodData` attempt, while card payments through the same provider succeeded), I confirmed via `GetSitePaymentMethodType` that **PayPlus's own Wix metadata already says `expressFlow.supported: false`** - yet its `description` field ("Allows you to accept credit or debit cards, Apple Pay, and more.") and Wix's own public support article for connecting PayPlus both advertise Apple Pay support. That mismatch is very likely why the merchant/buyer tried Apple Pay in the first place.

This is a good, generically-useful diagnostic pattern (not specific to PayPlus) - any provider can have this kind of description/capability mismatch - so it's worth capturing as a recipe rather than a one-off fix.

---
Opened automatically by an AI agent on behalf of Ayal (ayalg@wix.com), researching Wix MCP feedback. Slack thread: https://wix.slack.com/archives/C08P5DKLJR5/p1787295379844099

## Test plan
- [ ] Recipe file renders correctly and links resolve.
- [ ] SKILL.md index entry matches the recipe's frontmatter description.

### DIFF
--- a/skills/wix-manage/SKILL.md
+++ b/skills/wix-manage/SKILL.md
@@ -261,6 +261,9 @@ These recipes do NOT cover frontend development or SDK usage for displaying data
 ### [Get Paid Dashboard Navigation](references/get-paid/get-paid-dashboard-navigation.md)
 **Technical:** Direct links to payments and invoicing dashboard pages on manage.wix.com (payment links, invoices list, new invoice, invoice settings, recurring invoices, accept-payments settings), pairing each get-paid entity with its read API for "view it in your dashboard" links.
 
+### [Diagnose Checkout Payment Failures](references/get-paid/diagnose-checkout-payment-failures.md)
+**Technical:** Diagnoses why online checkout payments fail — especially when one payment method (e.g. Apple Pay/Google Pay) fails while another (e.g. cards) succeeds through the same provider. Checks Cashier transaction reason codes, then cross-checks the connected provider's declared `expressFlow`/`regularFlow` capabilities via Site Payment Method Types before concluding it's a platform bug.
+
 ---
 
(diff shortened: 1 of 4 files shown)


ayal:
🦞 Automated update from an AI agent acting on behalf of Ayal (ayalg@wix.com): a second, independent feedback report repeated the exact same misdiagnosis this recipe already guards against — blaming a benign `ContextProviderFactory: failed to import "ecom-platform-providers"` console warning for a PayPlus checkout/payment failure (see `wix-private/editor-platform` PR #14887, which rewords the warning itself). Pushed a commit adding an explicit callout to this recipe so agents don't chase that dead end a third time.

Slack thread for this report: https://wix.slack.com/archives/C08P5DKLJR5/p1787519022801649

PR #1089, opened 2026-08-23
fix(wix-app): document V3 product response wrapper and field shapes that fail tsc

adimara:
## Why

Codegen job `6b39b6a7-c995-4193-8246-2038dd7e04ac` (Studio 2, **`editorReactComponent`** on a Wix Stores product page) failed `tsc` **five distinct ways across four validate rounds**, needing two `codegen-validator` fix passes. The run came in at **6.82 min / $1.57** against a 3 min / $0.50 SLO, and roughly half the wall clock was the validator fix loop.

The agent read `STORES_VERSIONING.md` and `stores/GET_PRODUCT.md` *before* writing any code. It still got all five shapes wrong, because the references either didn't cover them or covered them ambiguously.

## How it got there

An Editor React Component placed on a product page receives **no product context** — no slot props, no `useCurrentProduct()`. The run spent a discovery subagent (*"Discover product page context hook"*) establishing exactly that, then derived the page URL, parsed a slug, and called `getProductBySlug`. That is a reasonable route, and it lands straight on an undocumented response wrapper.

(Contrast a **site plugin**, which sits in a slot and gets the id from `widget.getProp('product-id')` — the `getProduct` path, already documented.)

## The five failures

| Generated | Actual | Status before this PR |
|---|---|---|
| `res.variantsInfo` / `.currency` / `.inventory` after `getProductBySlug` | `res.product.*` — the slug method wraps | `getProductBySlug` appeared **nowhere** in `skills/wix-app/` |
| `import { V3Product } from '@wix/stores'` | `productsV3.V3Product` | module map's "all from `@wix/stores`" reads as a root type export |
| `variant.inventory` | `variant.inventoryStatus` | field map listed only product-level `inventory.availabilityStatus` |
| `product.description.replace(…)` | `product.plainDescription` | `description` never documented |
| `option.choices` | `option.choicesSettings?.choices` | documented, but buried in a code branch |

The root asymmetry: V3 `getProduct(id)` returns the product **directly**, `getProductBySlug(slug)` returns it **wrapped**. `GET_PRODUCT.md` documented only the former and stated "returns Product directly" — so the agent generalised that rule to the slug method.

## Where the fixes went

- **`STORES_VERSIONING.md`** — the four field-shape traps. It owns the module map, the V3 gotchas list and the V1→V3 field map, and `SKILL.md` directs agents to read it **first**. That placement also addresses why the existing `choicesSettings` warning wasn't landing: it lived only in a code branch further down a second file.
- **`stores/GET_PRODUCT.md`** — only the method-level content: the per-method response-shape table and the `getProductBySlug` snippet.
- **`editor-react-component/SITE-CONTEXT-HOOKS.md`** — a "No Business-Entity Context" rule recording that this extension type gets no product/service/post context and must resolve it from `useCurrentUrl()` by slug. These references previously had no data-resolution guidance at all.

The V3 shapes deliberately stay in the Stores references rather than any extension's doc — they are SDK facts and hold for any caller; a dashboard page hits the identical wrapper.

## Verification

Every snippet and mapping row was type-checked against the pinned `@wix/stores@1.0.823` (→ `@wix/auto_sdk_stores_products-v-3@1.0.193`) and the V1 `@wix/auto_sdk_stores_products@1.0.98` type definitions:

- Reproduced all five original errors verbatim under `strict` — same TS codes and messages as the job's validate output.
- The corrected snippets compile clean (`tsc --noEmit`, exit 0).
- Confirmed `getProduct` is unwrapped and `getProductBySlug` is wrapped, rather than assuming the asymmetry.
- `res.product` is `V3Product | undefined`, so the snippet uses the `!` idiom already established elsewhere in the file.
- V1 column checked too: used `variants[i].stock.inStock` rather than the **deprecated** top-level `stock.inStock`.
- The slug snippet was run against a real Wix product URL, including the trailing-slash case that yields an empty slug — hence the guard note.

## Recurrence

Over 30 days: `option.choices` → 4 jobs / 2 apps; the other three → 1 job each. Notably the one signature that *was* documented is the one recurring most, which is the evidence behind moving it into the read-first file.

## Eval coverage

Adds `yaml/wix-app-evals/stores-product-details-component.yml`, tagged `[stores-versioning, get-product, site-context-hooks, editor-react-component]` so it re-runs when any changed reference changes. It reproduces the job's build request and asserts all three references were read, `npm run build` passes, plus a correctness judge and a quality judge that fails the run for inventing a product-context hook, reaching for the site-plugin `getProp` API, or getting any of the five shapes wrong.


adimara:
## Review update — corrected attribution, and one addition

A review of the first commit found a **factual error in it**, now fixed in `4b3060d`.

### The error

The first commit said *"A site plugin on a product page has the **slug**, not the id."* That is backwards, and it contradicts `SITE_PLUGIN.md`: a site plugin sits in a slot and gets the id directly from `widget.getProp('product-id')`. Left as written it would have taught agents the opposite of the truth and undercut an adjacent reference.

### What actually happened in the job

The extension was an **Editor React Component**, not a site plugin — and that distinction is the entire reason the slug path was taken.

The run spawned a discovery subagent titled *"Discover product page context hook"*, whose prompt opens:

> I'm building a Wix Editor React Component (not a site plugin) that will be placed on the Wix Stores Product Page. I need to get the current product's ID at runtime…

Its conclusion:

> **Answer: NO.** There is no `useCurrentProduct()` or `useProductPageContext()` hook … Editor React Components running on a public site page have **no built-in access to the product context** at runtime.

Having established that, the agent derived the page URL, parsed a slug, and called `getProductBySlug` — landing on the response wrapper. So the chain is: *no product context for this extension type* → *URL* → *slug* → *wrapper trap*.

### Changes

- **`stores/GET_PRODUCT.md`** — corrected the sentence. It now names the Editor React Component case and states that a site plugin gets the id from `getProp` and should prefer `getProduct`.
- **`editor-react-component/SITE-CONTEXT-HOOKS.md`** — new **"No Business-Entity Context"** rule. These hooks report page and site state only; there is no `useCurrentProduct()`, slot props are a site-plugin affordance, and an Editor React Component must resolve the entity from `useCurrentUrl()` by slug. The wrapper is cross-linked. This records the negative result the discovery subagent had to spend a doc search re-deriving.
- **Eval** — re-framed to build an Editor React Component, tagged `site-context-hooks`, and its quality judge now fails the run for inventing a product-context hook or reaching for the site-plugin `getProp` API. Renamed to `stores-product-details-component.yml`.

### On placement

The question that prompted the review was whether this belongs in the site-plugin reference. It does not — the extension wasn't a site plugin, so that would have filed it under the wrong extension type. The split that holds:

- **V3 field shapes** are SDK facts, independent of caller — a dashboard page hits the identical wrapper. They stay in the Stores references.
- **"this extension type has no product context"** is extension-specific, and belongs with the editor-react-component references, which had no data-resolution guidance at all.

The `useCurrentUrl` slug snippet was checked against a real Wix product URL, including the trailing-slash case that yields an empty slug — which is what the guard note covers.

adimara:
## Fresh review — four findings, all fixed in `a7cdf07`

Reviewed against the repo's own gate code (`packages/evalforge-core`) rather than by eye. That turned up one blocking issue I would otherwise have shipped.

### 1. Eval tags were wrong — the gate would have failed 🔴

`deriveTags` maps a reference in a **subdirectory** to the *directory* name, not the filename:

```js
const [head, ...rest] = withinReferences.split('/');
// A sub-doc directory is already named for its capability.
tags.add(rest.length > 0 ? head.toLowerCase() : tagForReferencePath(head));
```

So the three changed references derive `[editor-react-component, stores, stores-versioning]` — not the per-file tags I'd guessed from the filename convention. `get-product` and `site-context-hooks` aren't tags at all, and `stores` was missing entirely.

Running the real `guardScenarios` against the original tags:

```
WITH ORIGINAL TAGS -> violations: [{"kind":"UNCOVERED_TAG","tag":"stores"}]
```

Corrected to `[stores, stores-versioning, editor-react-component]`. Re-running the actual `ScenarioSchema`, `meetsQualityBar` and `guardScenarios`:

```
SCHEMA: PASS
QUALITY BAR: {"ok":true,"reasons":[]}
DERIVED TAGS : [ 'editor-react-component', 'stores', 'stores-versioning' ]
VIOLATIONS   : []
WARNINGS     : []
```

### 2. "The one V3 read that wraps" was an overclaim 🟡

Not true. Verified against the pinned SDK: `searchProducts` returns `{ products, pagingMetadata }` and `countProducts` returns `{ count }` — both wrapped.

The accurate rule is the inverse, and it generalises better: **V3 reads return a response object; `getProduct` is the exception.** The table now says that and lists the other wrapping reads.

### 3. The `description` error text was a caller artifact 🟡

The gotcha quoted `Property 'replace' does not exist on type 'RichContent | ""'`. That `| ""` comes from the caller's own `?? ''` — including mine in the repro — not from the SDK. The declared type is `description?: RichContent`, and the gotcha now leads with that.

Also dropped the duplicated explanation from the field-map row; it only needs to carry the mapping.

### 4. This PR now leans on a hook the Studio 2 template can't compile ⚠️ — **not fixed here**

`SITE-CONTEXT-HOOKS.md` already noted these hooks need `@wix/react-component-utils` ≥ 1.12.0. But my new section promotes `useCurrentUrl()` from *one row in a table of eight* to *the prescribed way to resolve a product*, which widens the blast radius of that requirement.

The Studio 2 template pins `@wix/react-component-utils` at exactly **1.11.0** (`wix-code-studio/packages/wix-code-studio-remote-machine/pre-installed/package.json`), and `useCurrentUrl` first ships in **1.12.0** — so on that template the recommended path fails with `TS2305`. It's the same error the originating job hit.

I've added a pointer to the version requirement in the new section so an agent hitting it knows why. **The pin itself is out of scope for this PR** — it's a one-line bump in `wix-code-studio` plus a Docker image rebuild, and it belongs to whoever owns that template.

### Unverified

- Whether the EvalForge runtime matcher for `referenceFiles` handles nested paths (`references/stores/GET_PRODUCT.md`). The zod schema accepts any non-empty string, and the repo's own test uses a bare `SKILL.md` while the docs use `references/DASHBOARD_PAGE.md`, so the format is loose. If nested paths don't match server-side, those three coverage assertions would be no-ops — the `llm_judge` assertions still carry the real signal.
- That no product-context API exists for Editor React Components. That rests on the job's discovery subagent, and it's a negative claim — worth a Stores or Harmony owner confirming.

PR #1092, opened 2026-08-23
test(wix-app): first eval scenario batch — backend, event, data collection

mirivoWix:
## What

Adds 3 `wix-app` eval scenarios — the **first batch** of a larger set.

| Scenario | Tags | Covers |
|---|---|---|
| `backend/order-approved-event` | backend-event, data-collection | default-exported SDK webhook handler, `auth.elevate`, idempotency, try/catch |
| `backend/public-api-endpoint` | backend-api, data-collection | `src/pages/api/**` with typed `APIRoute` exports — the one type the CLI does **not** scaffold |
| `data/waitlist-data-collection` | data-collection, **dashboard-page (negative)** | field types, namespace-scoped ids, `PRIVILEGED` permissions, and no UI invented for a storage-only request |

## Why batched

The first version of this PR carried 17 scenarios and the gate **timed out**: 34 live agent builds across the PR and base arms against a hard 30-minute poll window (`POLL_TIMEOUT_MS` in `packages/evalforge-core/src/poll-eval-run.ts`, not an action input). These three are the lightest — backend/data only, so no `@wix/patterns` install, no `wix-design-system` skill invocation, no UI build.

The other 14 are preserved in this branch's history at `5fd27db` and land in follow-up batches of ~3 once this one goes green. Full set, by area: dashboard (5), backend service plugins + app tools (3), site (4), stores (2).

## Why these scenarios at all

The gate derives a tag from each changed file under `skills/wix-app/references/**` and fails with `UNCOVERED_TAG` when no quality-meeting scenario carries it. Only **3 of the 16** derivable tags have one today — a change to `SERVICE_PLUGIN.md`, `SITE_PLUGIN.md`, `EMBEDDED_SCRIPT.md`, `APP_TOOLS.md`, `BACKEND_API.md` and most others blocks on coverage rather than on behavior. This batch adds `backend-event` and `backend-api`; the follow-ups close the rest.

## Shape of each scenario

Per [docs/eval-scenarios.md](../blob/main/docs/eval-scenarios.md): coverage (`skill_was_called` naming the specific reference files, sub-references like `backend-event/COMMON-EVENTS.md` included) + `build_passed` + a correctness `llm_judge` + a trace-based quality `llm_judge` that penalizes and lists each stumble with the suspected skill/docs gap behind it, plus `time_limit` and `cost`.

`waitlist-data-collection` is a deliberate **negative dependency**: it carries `dashboard-page` so it re-runs when `DASHBOARD_PAGE.md` changes, and asserts that file was *not* used — a storage-only request must not grow a dashboard page.

## Verification

- All 4 scenarios (3 new + the existing one) parse against the real `ScenarioSchema` from `packages/evalforge-core`.
- All meet `DEFAULT_QUALITY_BAR` (≥3 assertions, ≥1 `llm_judge`); names are unique.
- Every `references/...` path named in an assertion exists on disk.


kobiburnley:
Heads-up on two things, neither blocking this batch.

**1. Three of the queued scenarios grade on reading `node_modules` directly.**

`dashboard/app-settings-page`, `dashboard/supplier-entity-page-not-modal` and `stores/catalog-version-aware-products-page` have quality judges that penalize *"writing JSX before enumerating the real component list with `cat node_modules/@wix/patterns/dist/docs/index.json`"*.

#1100 adds the opposite rule to `WIX_PATTERNS_DOCS.md` — *"Never inspect `node_modules` by hand — no `ls`, no `find`, no `cat` of a doc path"* — routing lookups through `scripts/patterns.cjs` instead. That came out of the 23 Aug generation-time regression: codegen's guard blocked the reads after a few attempts and the agent wrote itself scripts to get around them.

So as worded, the rubric rewards the behaviour that caused the incident. An agent using `patterns.cjs` has still enumerated, so a judge might pass it either way — but it seems worth not leaving ambiguous in something that gates PRs. Cheaper to reword on this branch than as a follow-up after merge — happy to do it either way, just don't want us both editing the same files.

**2. Three more dashboard scenarios are coming.**

From the Custom dashboard V1 scope: refund investigation, customer product finder, booking activity — all read-only over native APIs, which nothing currently covers. Flagging it for batch sizing, since dashboard scenarios are the slow ones.

Related question: have you measured whether the `dashboard-page` tag fits the 30-minute poll window? With your 5 plus my 3 plus `employee-shift-dashboard`, a PR touching `DASHBOARD_PAGE.md` would run 9 × 2 arms. You've hit that ceiling once and I haven't, so you'd know better than me whether that's fine.

PR #1100, opened 2026-08-24
feat(wix-app): dashboard page skill — design brief, deterministic patterns docs, component toolkits

listguy:
## What

Adds a new dashboard-page reference: `skills/wix-app/references/dashboard-page/UX_SUCCESS_MODEL.md`, ported from the FeatureVibe product doc.

It covers:
- **Success model** — workflow / presentation / data as three connected dimensions
- **Workflow journey** — understand → focus → investigate → act → feedback, each with intent, guidance, and a success criterion
- **The 5 WHATs** — turning a vague prompt into concrete journey requirements
- **Presentation success** — when to use table / gallery / list / kanban / timeline / chart / summary metrics, and side panel vs modal vs entity page vs inline row for drill-in
- **Data success** — the major Wix data sources, what must hold across sources, and multi-source examples
- **Evaluation checklist** — what to verify before calling a dashboard done

## Why

The existing dashboard-page references cover *how* to build (APIs, layout, patterns/WDS component order) but not *what* a good dashboard should contain. This gives the skill a design brief to reason from when the prompt is open-ended.

## Notes

- Slack-canvas artifacts removed (layout divs, placeholder Gemini images, emoji shortcodes).
- The stage table's `confirmation` row renamed to **Feedback** to match its own section heading; `Kanban (?)` de-questioned.
- Linked from `references/DASHBOARD_PAGE.md` above Layout Guidelines, with a cross-reference to the patterns-first component order so the stage list isn't read as license to hand-roll shells and tables.
- Source doc linked at the top of the file.


kobiburnley:
#skipreview

kobiburnley:
#skipreview

skills/wix-app/scripts/patterns.cjs
```diff
@@ -0,0 +1,276 @@
...
+      body = fs.readFileSync(file, 'utf8');
+    } catch (e) {
+      console.error(`Skipped "${name}" — could not read ${index[name].file}: ${e.message}`);
+      continue;
+    }
+
+    if (printed.length) console.log('\n---\n');
+    console.log(`# ${name}  (${index[name].category}, ${index[name].file})`);
+    console.log('');
+    console.log(body.trim());
```
kobiburnley: **Blocker: `docs` prints whole doc files, so the agent truncates them and loses the props table.**

Measured on `@wix/patterns@1.448.0`: `docs Table` is 48,120 chars; `docs Table --refs` is **178,790 chars across 22 docs** (~45k tokens in one tool result). `Table.md` is 1,083 lines with `## API` at line 986; `ToolbarFilters.md` is 1,580 lines with `## API` at line **1,554**.

In the measured run the agent defended itself with `head` on every multi-doc call, and **not one reached a `## API` section**:

| call | requested | output | piped | `## API` seen |
|---|---|---|---|---|
| 14 | 6 docs | 3,306 lines | `head -200` | 0 |
| 16 | 4 docs | - | `head -150` | 0 |
| 18 | 2 docs | - | `head -120` | 0 |
| 65 | 3 docs | - | `head -200` | 0 |

Call 14 asked for six docs and effectively received one (`CollectionSearch`, minus its API) — and never used it. That is why this branch shipped two filters where `main` shipped four.

**Fix:** make compact the default rendering — header + the first `@wix/*` import statement found in the body (only 3 of 167 docs have an `## Import` heading; 131 have the statement somewhere, so regex the body) + `## API` verbatim + the first code fence. Add `--full` for today's behavior, and use compact for cross-references. For those six docs that is 362 lines instead of 3,287 (11%), which fits uncut inside every `head` the agent reached for. The 26 docs with no `## API` can fall back to the whole body capped around 150 lines — 8 of them exceed 200.

skills/wix-app/references/WIX_PATTERNS_DOCS.md
```diff
@@ -2,13 +2,19 @@
 
 ## Prerequisites
 
-The docs are pre-built at `node_modules/@wix/patterns/dist/docs/` (160+ markdown files + `index.json`). This requires `@wix/patterns` version **1.367.0** or later. Verify they exist:
+This skill bundles `scripts/patterns.cjs`. It locates the docs itself (pnpm, Yarn PnP, workspaces, symlinks) and prints them. **Never inspect `node_modules` by hand** — no `ls`, no `find`, no `cat` of a doc path.
```
kobiburnley: **Blocker: this closes the only route to type information without opening another one.**

The docs carry props tables, not TypeScript types, and a real dashboard page needs both. In the measured run this branch needed the `RangeItem` type for a date filter, `patterns.cjs` had no way to answer "where does this type come from", so the agent grepped `dist/esm/index.js` (runtime exports, no types) across 7 calls and landed on:

```ts
import type { RangeItem } from '@wix/bex-core/dist/types/state/RangeFilterState/RangeFilterState';
```

a deep path into an undeclared transitive dep that will break on any `bex-core` restructure. The `main` arm read `dist/types/` directly — the thing this line forbids — and wrote the correct `import type { Filter, RangeItem } from '@wix/patterns'`. `RangeItem` *is* re-exported from `@wix/patterns` (via `@wix/bex-core`), and `dateRangeFilter.md` mentions it, but it isn't a documented name in `index.json`, so the script cannot reach it.

Same mechanism produced the baseline's win on filters: it got `CollectionToolbarFiltersProps.d.ts` and `useFilterCollection.d.ts` from `dist/types/`.

**Fix:** add `patterns.cjs types <Name>` that reads `dist/types/` and prints the export path plus the declaration, then this prohibition has somewhere to send the agent. Without it, the rule reliably produces a worse import than no rule at all.

skills/wix-app/references/DASHBOARD_PAGE.md
```diff
@@ -27,13 +33,7 @@ wix generate --params '{"extensionType":"DASHBOARD_PAGE","title":"<title>","rout
...
 The CLI generates the folder, `page.tsx`, the builder file, the UUID, and the `src/extensions.ts` registration. After scaffolding, implement the page UI in the generated `page.tsx`.
 
-**Then, before writing UI:** confirm `@wix/patterns` is a dependency and its docs are present.
-
-```bash
-ls node_modules/@wix/patterns/dist/docs/index.json
-```
-
-If it's missing, add the package (`^1.367.0` or later — earlier versions don't ship `dist/docs/`) and install. Without this, the patterns lookup in step 1 of [Component Selection Order](../SKILL.md#component-selection-order) silently finds nothing and the page gets built entirely from WDS.
+**Then, before writing UI:** run `node $PATTERNS list` once, per [Prerequisites](WIX_PATTERNS_DOCS.md#prerequisites).
```
kobiburnley: **Blocker: `$PATTERNS` is never assigned in this file, and a shell variable doesn't survive to the next tool call anyway.**

The assignment exists only in `SKILL.md:115` and `WIX_PATTERNS_DOCS.md:8`. Each Bash tool call gets a fresh shell, so even an agent that ran the assignment block gets nothing here:

```
$ node $PATTERNS list
node:internal/modules/cjs/loader:1503  throw err;   # Cannot find module '/.../list'
```

An unexplained `MODULE_NOT_FOUND` stack at the exact gate this PR exists to protect — and the documented fallback behaviour is the `node_modules` crawl. Line 96 (`node $PATTERNS docs <Name>`) has the same problem.

For contrast, `wix-design-system` keeps every `$WDS` reference inside the one file that assigns it; this is the first place the variable crosses a file boundary.

**Fix:** spell the full `node <this-skill-dir>/scripts/patterns.cjs list` here and at line 96, and note next to each assignment that it must be re-set per shell invocation. (The measured run re-assigned it in all 9 calls and survived — but only because it happened to read `SKILL.md` first.)

skills/wix-app/references/dashboard-page/UX_SUCCESS_MODEL.md
```diff
@@ -0,0 +1,184 @@
...
+
+| Single view | Primary record | Connected sources | User outcome |
+|---|---|---|---|
+| **Order exceptions** | eCommerce order | Payments, Contacts, Stores | Resolve unpaid, delayed, or incorrect orders from one queue. |
+| **Booking follow-up** | Booking or appointment | Bookings, Contacts, Payments | Find missed or unpaid appointments and contact the client. |
+| **Subscription health** | Pricing-plan subscription | Pricing Plans, Payments, Members | Identify at-risk subscribers and take retention action. |
+| **Inventory reorder** | Store product or variant | Stores, eCommerce orders, CMS | Recognize shortages and decide what to reorder. |
+
+
+## Evaluation
```
kobiburnley: **Blocker: this checklist is the part that gets truncated away.**

`scripts/check-truncation.mjs` reports the file at **10,408 chars, +408 over** the ~10k reference-fetch limit, `lines kept: 179/184`. What falls off the end is the tail of this Evaluation list and the closing evaluation question — exactly what `DASHBOARD_PAGE.md` advertises the file for ("including the evaluation checklist to run before calling a dashboard done"). The check isn't wired into CI, so the green checks don't cover it.

This matters more now than when the PR was opened: `28ed51ed` gives the agent three paths to this file, so it will actually be read — and what it will get is the file minus its closing checklist. In the measured run both arms shipped a page with no summary metrics and no drill-in, which is precisely what "Presentation fit" and "Task completion" here would have caught.

**Fix:** ~500 chars of trim gets it under. Cheapest candidates are the Multi-Source Examples table or the per-stage bullets that restate the stage table; alternatively move Evaluation above Data Success so truncation costs prose instead of the checklist.

skills/wix-app/SKILL.md
```diff
@@ -22,7 +22,9 @@ Helps build extensions for Wix CLI applications. Covers all extension types: das
...
+  - [ ] **Dashboard page UI:** Translated the prompt into a workflow before choosing components — what the user must understand, focus on, investigate, act on, and see confirmed. A bare filtered table answers none of "how many", "which one needs my attention", or "why did this happen", and a bare table is the default outcome when the workflow was never named. See [UX Success Model](references/dashboard-page/UX_SUCCESS_MODEL.md).
+  - [ ] **🛑 Patterns Docs Gate (MANDATORY for any dashboard page UI):** Read [WIX_PATTERNS_DOCS.md](references/WIX_PATTERNS_DOCS.md), then list the component inventory with `node <this-skill-dir>/scripts/patterns.cjs list`, installing `@wix/patterns@^1.367.0` if it reports the package missing or too old. Never inspect `node_modules` by hand.
```
kobiburnley: **Blocker: this gate asks for the action the line directly above it forbids.**

Line 25 ends "Never inspect `node_modules` by hand." Line 26 then says "Read the `.md` doc file for every patterns component..." — and the `.md` doc files live in `node_modules/@wix/patterns/dist/docs/`. Read literally, the checklist instructs a `Read` on a `node_modules` path, which is what codegen's guard blocks and what caused the 23 Aug generation-time regression.

**Fix:** "print the doc for every patterns component, hook and state type you are about to write with `node <this-skill-dir>/scripts/patterns.cjs docs <Name...>`" — the intent is to consult the doc, and the only sanctioned way to do that is the script. Worth scoping line 25's prohibition to patterns *docs* at the same time (see the note in the body about `COMPONENT-CONTRACT.md:201`).

kobiburnley (changes requested):
Request Changes — 5 blockers, split by what they cost, with checkboxes and evidence in the fix tracker: https://github.com/wix/skills/pull/1100#issuecomment-5440067761

Short version: `patterns.cjs` works and cuts patterns-lookup bytes 60%, but on a measured `dashboard/booking-activity` run this branch cost +20% ($5.59 vs $4.64) and produced a weaker page than `main` (two working filters vs four). Both trace to one cause — the docs are too big to read whole, so the agent truncates them with `head` and never reaches a props table, and the route that answered its type questions (`dist/types/*.d.ts`) is now forbidden with nothing in its place. P1 and P2 in the tracker are the two that pay in both columns.

kobiburnley:
## #1100 fix tracker

Tick a box and drop the commit SHA next to it as each lands. 🛑 = blocker for merge.

**Headroom warning:** `DASHBOARD_PAGE.md` is 102 chars and `UX_SUCCESS_MODEL.md` 59 chars from the 10k fetch limit — both are effectively frozen. New dashboard-page guidance belongs in `COLLECTION_TOOLKIT.md` (~4,900 free). This is also why **R6** matters.

**All five 🛑 blockers are now cleared** — P1, P2, P4, P5, R3, across 62c02fa7, 34c6f281, 7396ad2d and 0f2a73fd (plus 28ed51ed for R0). The standing Request Changes is stale on that basis, but four of those commits are mine, so the sign-off is not mine to give. Everything still open below is non-blocking.

**Unrelated CI note:** `check` is red here because `main` itself carries an unpinned `npx @wix/cli token` at `wix-replatform/resources/rp-target-wix/lib/wix-writers.js:327`, and CI evaluates the PR merged into current `main` (tip `c7a07c94`). `bash .github/scripts/check-cli-pinned.sh` passes on this branch's own tree. One-line fix, and it belongs on `main` — it will block every PR in the repo until then.

Findings come from running the `dashboard/booking-activity` V1 prompt through two arms — this branch merged into current `main` vs `main` alone — in an identical local harness with `@wix/patterns@1.448.0`.

### Final run — branch HEAD on `@wix/patterns@1.452.0` (both upstream fixes live)

Single treatment arm; `main`'s numbers for this scenario were measured twice and did not move.

| assertion | iter 1 | iter 2 | **now** |
|---|---|---|---|
| summary metrics | ✗ | ✗ | **✓ `SummaryBar`** |
| drill-in | ✗ | ✗ | **✓ WDS `SidePanel` + `onRowClick`, in its own `BookingSidePanel.tsx`** |
| filters wired | 2 | 2 | **4** — status, service, staff, dateRange |
| `RangeItem` import | ✗ deep `@wix/bex-core` path | ✓ | ✓ |
| typecheck | pass | pass | pass |

**R1 is closed.** The filter gap that survived two runs is gone, and the page now carries the aggregate and the drill-in that *no* arm had produced before — including `main`, which never had them in any run.

The mechanism was **R10**, not R7: the agent already had `SummaryBar`'s doc in hand during iteration 2 and shipped nothing. Naming components was necessary; stating the requirement is what changed the output.

| | iter 2 | now | `main`, iter 2 |
|---|---|---|---|
| cost | $3.28 | **$4.79** | $4.37 |
| wall | 11.4 min | **16.6 min** | 19.2 min |
| tool calls | 79 | **114** | 113 |
| patterns hand-reads | 0 | **15** | 15 |

Dearer and slower than iteration 2 — which shipped a bare two-filter table. Against `main` on the same scenario it is 14% faster and 10% cheaper while producing materially more page. The scenario's $2.50 ceiling is still blown either way (**P12**).


### Third scenario — the bare user prompt, and where the time actually goes

Same employee-shift task **without** the eval's `APP_NAMESPACE`/`CODE_IDENTIFIER` lines — what a user actually types. Both arms, current `main`, `@wix/patterns@1.452.0`.

| | branch | `main` | delta |
|---|---|---|---|
| cost | **$5.22** | $4.19 | **+25%** |
| wall clock | **19.8 min** | 13.2 min | **+50%** |
| tool calls | 103 | 110 | −6% |
| `node_modules/@wix/patterns` reads | **8** | 21 | −62% |
| typecheck / `wix build` | clean | clean | — |
| **codegen validator equivalent** | `CLEAN`, **2 errors fixed → would alert** | `CLEAN`, 0 fixed | **worse** |

Quality a draw: both produced `CollectionPage` + `Table` + a real `EntityPage` add-flow and the scoped collection; the branch added `SummaryBar` and `CollectionEmptyState`. **Neither** resolved the app namespace properly — both invented `collectionIdSuffix = 'employee-shifts'` — so CAIRO-4634's namespace gap is untouched by either version.

**This is the first result that argues against the branch**, and by codegen's own alert rule its run would have posted to the team channel where `main`'s would not.

#### The cost claim does not generalise

| scenario | cost vs `main` | wall vs `main` |
|---|---|---|
| booking-activity | −10% | −14% |
| employee-shift (scaffolded prompt) | −13% | +23% |
| employee-shift (bare prompt) | **+25%** | **+50%** |

What *is* consistent across all three: patterns hand-reads down 62–100%, and more complete pages. "Cheaper and faster" is not supportable and should not be claimed.

#### Where the wall clock goes (measured from transcript timestamps)

Tool execution is negligible — 28s of 1,184s. Wall clock is model turns.

| phase | branch | `main` |
|---|---|---|
| **`node_modules` crawl** | **488s (41%)** | **456s (58%)** |
| write code | 208s | 95s |
| `patterns.cjs` lookups | 188s | 0 |
| everything else | 300s | 242s |

And the crawl is **not patterns** — it is the Wix Data SDK: `@wix/data` (18 calls), `wix-data-items-common` (12), `wix-data-items-sdk` (6). The branch cut patterns crawling from 21 calls to 8 and spent the savings on data-SDK discovery.

**So this PR optimised the smaller half.** Patterns docs were ~20% of the problem; the Wix Data SDK is ~45%, and codegen's guard (`MAX_SPELUNK_CALLS = 2`) would block both arms after the second crawl anyway.


### Iteration 2 — after 62c02fa7 (P1–P9)

| | this branch | `main` | delta |
|---|---|---|---|
| wall clock | **11.4 min** | 19.2 min | **−41%** |
| cost | **$3.28** | $4.37 | **−25%** |
| tool calls | **79** | 113 | −30% |
| `node_modules/@wix/patterns` touches | **0** | 15 | — |
| working filters shipped | 2 | **4** | still behind |
| summary metrics / drill-in | none / none | none / none | unchanged |

### Second scenario — `employee-shift-dashboard` (the CRUD path), on 7396ad2d

Run because every result above came from one read-only scenario. This one requires a scoped data collection and its judge caps a hand-built WDS form in a dashboard modal at 4, so it exercises the `EntityPage` rule and the `/form` subpath that booking-activity never touches.

| | treatment | baseline | delta |
|---|---|---|---|
| wall clock | 15.1 min | 14.2 min | **+6% — slower** |
| cost | **$3.74** | $4.19 | −11% |
| tool calls | **66** | 102 | −35% |
| errored tool results | **4** | 10 | −60% |
| `node_modules/@wix/patterns` hand-reads | **14** | 41 | −66% |
| typecheck | clean | clean | — |

**Quality is a draw here, and both arms are good.** Both produced `CollectionPage` + patterns `Table` + `useTableCollection`; both put the add flow in a real `EntityPage` with `useEntityPage`, `usePatternsNavigate` and `@wix/patterns/form`; both created the scoped collection; both kept the provider in a separate file from the hook (`employee-shifts.tsx` / `ShiftsApp.tsx` vs `ShiftsCollectionPage.tsx`); neither used a WDS modal, `@wix/auto-patterns` or a `patterns.json`. Both read all three asserted reference files, and the treatment arm read `UX_SUCCESS_MODEL.md` — R0 confirmed a second time.

Two corrections this forces on the claims above:

- **"Hand-reads went to zero" was scenario-specific.** Booking-activity's treatment arm hit 0; here it is 14. See **R11** — they are a question category neither `docs` nor `types` answers.
- **The wall-clock win does not generalise.** −41% on booking-activity, +6% here. Cost and tool calls still move the right way in both.

### Iteration 1 — before

| | this branch | `main` | delta |
|---|---|---|---|
| wall clock | 20.0 min | 18.4 min | +8.7% |
| cost | $5.59 | $4.64 | +20.4% |
| tool calls | 137 | 121 | +13.2% |
| `node_modules/@wix/patterns` touches | 8 | 22 | — |
| working filters shipped | 2 | 4 | — |

**The performance case is now made.** Against its own iteration 1 the branch is 43% faster, 41% cheaper and 42% fewer tool calls; against `main` it is 41% faster and 25% cheaper, where before it was slower and dearer. Hand-reads of the patterns docs went to **zero**. The `types` command was used unprompted (`types ComputedQuery CursorQuery`, `types RangeItem`) and `--full` exactly once, on the one doc that needed it.

**The result case is not made yet.** The branch still ships two working filters where `main` ships four, and neither arm produces summary metrics or drill-in. My iteration-1 hypothesis — that the missing props tables caused the filter gap — is **not supported**: the agent now receives far more API tables and still shipped two filters. See R1 and R9.

---

## Performance — time, cost, turns

- [x] **🛑 P1 — `docs` prints whole doc files, so the agent truncates away the props table.** `scripts/patterns.cjs:225`. Make compact the default (header + first `@wix/*` import found in the body + `## API` verbatim + first code fence), add `--full`, use compact for cross-refs. *(upstream companion: U4)* — fix: 62c02fa7
  <details><summary>evidence</summary>

  `docs Table` = 48,120 chars; `docs Table --refs` = **178,790 chars / 22 docs** (~45k tokens in one result). `Table.md` is 1,083 lines with `## API` at 986; `ToolbarFilters.md` is 1,580 lines with `## API` at **1,554**.

  Every multi-doc call was self-truncated, and **not one reached a `## API` section**:

  | call | requested | output | piped | `## API` seen |
  |---|---|---|---|---|
  | 14 | 6 docs | 3,306 lines | `head -200` | 0 |
  | 16 | 4 docs | — | `head -150` | 0 |
  | 18 | 2 docs | — | `head -120` | 0 |
  | 65 | 3 docs | — | `head -200` | 0 |

  So the agent asked again: **6 of the 9 `patterns.cjs` calls re-fetched a doc it had already requested** (`ToolbarFilters` four times, `MultiSelectCheckboxFilter` four). That is where the +16 tool calls and most of the +20% cost went — cost is turns × context, not result bytes.

  Compact for those six docs is 362 lines vs 3,287 (11%), fitting uncut inside every `head` the agent reached for. Only 3 of 167 docs have an `## Import` heading but 131 have the statement somewhere, so regex the body. The 26 docs with no `## API` can fall back to whole body capped ~150 lines (8 exceed 200).
  </details>

- [x] **🛑 P2 — no sanctioned route to type information, so the agent flails.** `references/WIX_PATTERNS_DOCS.md:5`. Add `patterns.cjs types <Name>` reading `dist/types/`. *(upstream companion: U2)* — fix: 62c02fa7
  <details><summary>evidence</summary>

  Docs carry props tables, not TS types. This branch needed `RangeItem`, the script couldn't answer it, and it spent **7 calls** grepping `dist/esm/index.js` — runtime exports, no types — before giving up and guessing. `main` read `dist/types/` (the thing this line forbids) in 2 calls and had its answer.
  </details>

- [x] **P3 — retire the "one call" instruction.** `references/WIX_PATTERNS_DOCS.md:139` ("Pass every name you need in **one** call") is what produced the 3,306-line result. Once output is compact, batching is safe — say so, and say explicitly not to pipe the script through `head`/`sed`, because on long docs the props table is at the end. — fix: 62c02fa7
- [x] **🛑 P4 — `$PATTERNS` is never assigned in this file, and a shell var doesn't survive to the next tool call.** `references/DASHBOARD_PAGE.md:36` and `:96`. — fix: 62c02fa7
  <details><summary>evidence</summary>

  Assignment exists only in `SKILL.md:115` and `WIX_PATTERNS_DOCS.md:8`. Every Bash tool call gets a fresh shell, so:

  ```
  $ node $PATTERNS list
  node:internal/modules/cjs/loader:1503  throw err;   # Cannot find module '/.../list'
  ```

  An unexplained `MODULE_NOT_FOUND` at the exact gate this PR protects, whose documented fallback is the `node_modules` crawl. `wix-design-system` keeps every `$WDS` use in the file that assigns it; this is the first time the variable crosses a file. The measured run re-assigned it in all 9 calls and survived — but only because it read `SKILL.md` first.
  </details>

- [x] **🛑 P5 — the Component Docs Gate asks for the action the line above forbids.** `SKILL.md:26`. Line 25 ends "Never inspect `node_modules` by hand"; line 26 says "Read the `.md` doc file for every patterns component" — and those files live in `node_modules/@wix/patterns/dist/docs/`. Reword to "print the doc … with `patterns.cjs docs <Name…>`". — fix: 62c02fa7
- [x] **P6 — the script's own hints teach an invocation that fails.** `patterns.cjs:155` and `usage()` print `patterns.cjs docs <Name1> …` with no `node` and no path. `process.argv[1]` is the resolved path — `node ${process.argv[1]} docs …` makes it self-locating and removes most of P4. — fix: 62c02fa7
- [x] **P7 — scope the `node_modules` prohibition.** `SKILL.md:25` is unqualified, but `references/editor-react-component/COMPONENT-CONTRACT.md:201` legitimately reads a `.d.ts` there. `WIX_PATTERNS_DOCS.md:5` scopes it correctly ("of a doc path") — SKILL.md should match. — fix: 62c02fa7
- [x] **P8 — one install instruction, not three.** `SKILL.md:25` says `^1.367.0`, `WIX_PATTERNS_DOCS.md` says "latest version", the script's error says `npm install @wix/patterns@^1.367.0`. Worth a sentence on what codegen may actually install. — fix: 62c02fa7
- [x] **P9 — `--refs` off-by-one.** `patterns.cjs:230` gates expansion on `i < requested.length`, but `queue` holds only *resolved* names. One typo buys an extra level: `docs Tabel Table --refs` prints 24 docs vs 22. Compare against the initial resolved queue length. — fix: 62c02fa7
- [x] **P11 — the `$PATTERNS` variable is gone.** `SKILL.md` spells the path out; `WIX_PATTERNS_DOCS.md` gives it once in Prerequisites and refers to subcommands thereafter, with a note on why the variable is a trap. Paid for inside the fetch limit by dropping a third copy of the install command (the script prints it on failure) and the collection-triad table's redundant first column — that file went from 20 to 255 chars of headroom. — fix: 48102f60
- [x] **P12 — ceilings re-baselined from measured runs.** `maxCostUsd` 2.5 → 6.5, `maxDurationMs` 900000 → 1440000 (24 min). Four runs of this scenario, both arms, two sessions: $3.17 / $3.65 / $3.74 / $4.19 and 12.9 / 14.2 / 15.1 / 15.9 min — every arm breached both old ceilings, including the unmodified skill, so the gate could only report its own limit. Set at ~1.5x the worst observation. — fix: on branch
- [x] **P13 — `types` and `exports` take a list, and the skill now says so.** 6 of 18 script calls in the bare-prompt run were `exports`, four of them the bare no-arg listing repeated; 3 more were one-name `types` calls. Both commands already accepted a list — only `docs` documented it. ~7 avoidable turns, ~70s. — fix: 10b01189
- [ ] **P14 — the Wix Data SDK is the real cost centre.** 488s and 45% of the branch's spend goes to reverse-engineering `@wix/data`, `wix-data-items-common` and `wix-data-items-sdk` from `node_modules` — a path the skill says nothing about, and one codegen blocks after two calls. Worth more than everything shipped in this PR so far. This is CAIRO-4634's "CMS schema in docs" item: the join between a Wix Data collection's schema and the patterns table/form wiring. — fix: _pending_
- [ ] **P15 — neither version resolves the app namespace.** Given a bare prompt with no `APP_NAMESPACE`, both arms invented `collectionIdSuffix = 'employee-shifts'` rather than resolving it via the apps API. The harness README flags this as knowledge the playground repo holds and the skill does not. (CAIRO-4634) — fix: _pending_

- [ ] **P10 — trace guard.** Fail a run that pipes `patterns.cjs` through `head`/`sed`/`tail` — that's the signal output is still too big, and it is exactly what this run's failure looked like. Not this PR. (CAIRO-4636) — fix: _pending_

## Actual result — does the dashboard work

- [x] **R0 — `UX_SUCCESS_MODEL.md` was unreachable from the read path.** Linked from one place only (`DASHBOARD_PAGE.md:155`, below Examples and API Spec Support), and `SKILL.md` never mentioned it. In the measured run the agent read `DASHBOARD_PAGE.md` in full and **never opened the file** — zero tool calls referenced it, so none of this PR's design content was in context when the page was built. Pointer moved above UI Libraries, plus a `SKILL.md` Step 2 item and reference-table row. — fix: 28ed51ed
- **R1 — props tables now survive, and the filter gap did not close.** ❌ **Not fixed by P1.** Iteration 1: zero `## API` sections reached the agent across four filter-doc calls, and I read that as the cause of the filter gap. Iteration 2, with compact output: the agent gets most of the API tables and still shipped two filters (status via `TabsFilter`, date) against `main`'s four (status, date, service and staff, the last two dynamic via `useFilterCollection`). Same gap, twice, so it is probably not variance — but the cause is now unknown. See **R9** for the one lead.
- **R2 — the import has to be right.** ✅ **Fixed by P2 (62c02fa7).** Iteration 2 wrote `import type { Filter, RangeItem } from '@wix/patterns'` — the deep `@wix/bex-core` path is gone, and the agent reached it by calling `types RangeItem` itself.

  <details><summary>what iteration 1 did</summary> Without a types route the agent wrote `import type { RangeItem } from '@wix/bex-core/dist/types/state/RangeFilterState/RangeFilterState'` — a deep path into an undeclared transitive dep. `main` read the `.d.ts` and wrote the correct `import type { Filter, RangeItem } from '@wix/patterns'`. `RangeItem` *is* re-exported from `@wix/patterns`, and `dateRangeFilter.md` mentions it, but it isn't a documented name in `index.json` so the script can't reach it. Same mechanism gave `main` its filter win: `CollectionToolbarFiltersProps.d.ts` and `useFilterCollection.d.ts`.
  </details>
- [x] **🛑 R3 — the Evaluation checklist no longer gets truncated away.** 10,408 → 9,704 chars. Two cuts, both duplication: the stage table's *Typical dashboard support* column (restated in more detail by the five subsections below it) and two of four multi-source examples (kept Order exceptions and Booking follow-up, the two that map onto V1 acceptance scenarios). **Only ~300 chars of headroom remain and R5 will consume it** — land R5 together with any further addition to this file. — fix: 0f2a73fd
  <details><summary>evidence</summary>

  `scripts/check-truncation.mjs`: **10,408 chars, +408 over** the ~10k fetch limit, `lines kept: 179/184`. What falls off is this checklist's tail and the closing question — what `DASHBOARD_PAGE.md` advertises the file for. More urgent after R0, which gives the agent three paths to the file: it will now be read, minus its closing checklist. Both arms shipped a page with no summaries and no drill-in — exactly what "Presentation fit" and "Task completion" would have caught.
  </details>

- [x] **R5 — the drill-in table now points at real primitives.** Landed as a pointer, not inline: `UX_SUCCESS_MODEL.md` has 59 chars of headroom against the 10k limit, so WDS `SidePanel` / `EntityPage` / `PickerModal` live in the new toolkit and the table links to them. — fix: e8876ee3
- [x] **R6 — the fetch limit is gated in CI.** `check-truncation.mjs` gained `--files=` and `--fail`; `detect` now exports the reference files a PR touches and `typecheck` runs the gate on them. Scoped on purpose: ~100 pre-existing over-limit files mean a whole-repo gate could only ever be red, while a scoped one stops new violations. Verified: passes on this PR's four changed reference files, exits 1 on a known over-limit file. — fix: 48102f60
- [x] **R7 — the components are named.** New `references/dashboard-page/COLLECTION_TOOLKIT.md`, organised by user need rather than by category so the UX model's stages resolve to real names: `SummaryBar`; the ~20 `Features/Filter` names; `CollectionEmptyState` vs `CollectionNoResultsState` (different messages); `TableTopNotification`; `ExportTo`; row and bulk actions. All 44 names verified against the installed index. — fix: e8876ee3
- [x] **R8 — side-panel drill-in guidance.** In the toolkit's Investigate table: WDS `SidePanel` to inspect without losing the filtered list (patterns ships none), `EntityPage` + `useEntityPage` for deep or shareable detail, expanded row for a field or two, `PickerModal`/`bulkActionModal` for selection. — fix: e8876ee3

- [x] **R9 — the props table now sits where truncation can't reach it.** The `head` reflex can't be instructed away: the decision to pipe is made before any output exists. Fixed structurally instead. *(upstream companion: U4)* — fix: 34c6f281
  <details><summary>what was wrong and what changed</summary>

  The compact renderer emitted import → example → API, so inside one chunk the props table sat *behind* the example: `ToolbarFilters` at chunk line 168 of 170, `SummaryBar` at 83 of 93 — the same layout defect the compact view was meant to fix, one level down. In iteration 2 that still cost the agent props: 7 of 10 script calls were piped, a 7-name call at `head -300` got 3 of 7 API tables, a 5-name call at `head -200` got 1 of 4.

  Three changes: **API before the example**, so truncation costs prose not props; **examples only for one- or two-name calls** (never with `--refs`, which grows the queue afterwards), since examples are what push a later component's props out of reach; and a **header stating total size** — `<< 7 docs, 290 lines: import + API. This is the whole answer >>` — because a footer arrives after the pipe has already cut it.

  | | before R9 | after |
  |---|---|---|
  | the 7-name call | 675 lines, 3 of 7 API tables at `head -300` | **305 lines, 7 of 7** (6 of 7 at `head -200`) |
  | `docs Table --refs` | 70,937 chars | 53,337 |
  | whole corpus | 15,451 lines | 10,894 |
  | API tables dropped across all 167 docs | 0 | 0 |

  Re-measured on `employee-shift-dashboard` rather than a booking-activity rerun: the `head` reflex did not recur there and the arm made 35% fewer tool calls with 60% fewer errored results. Whether R9 closes the booking-activity filter gap in **R1** is still open — that needs the rerun.
  </details>

- [x] **R10 — the skill now states the requirement, not the goal.** The measured finding was that awareness wasn't the gap: the agent fetched `SummaryBar`'s doc twice and shipped nothing. `SKILL.md`'s Step 2 item now asserts three things with their exceptions — aggregate numbers, an openable row, every filter wired to the query — instead of asking the agent to translate a prompt into a workflow and hoping. — fix: e8876ee3

## Upstream — belongs in the `@wix/patterns` source (cairo), not here

We own cairo too, so these are ours to schedule rather than to file and wait on. The skill-side entries above are workarounds that stay useful either way; when a cairo PR lands for one of these, link it here and the corresponding workaround can be trimmed back.

- [ ] **U1 — a doc's key doesn't match the symbol you import.** `ToolbarFilters.md` is published under the key `ToolbarFilters`, while line 29 of that same file reads `import { CollectionToolbarFilters } from '@wix/patterns'`. The exported symbol is `CollectionToolbarFilters`, and so is the cairo source directory. So an agent holding the symbol name cannot reach the doc:

  ```
  $ patterns.cjs docs CollectionToolbarFilters
  Skipped "CollectionToolbarFilters" — not a documented name.
  ```

  Both booking-activity arms shipped `<CollectionToolbarFilters>` in working code, and the employee-shift treatment arm hand-read `CollectionToolbarFiltersProps.d.ts` instead — part of R11's 14 `node_modules` reads. Same class, each needing the same one-line check: `PickerModalIndeterminate` ships as `Indeterminate`, `dateRangeFilterState` as `dateRangeFilter`, `CustomFilterState` as `customFilter`, `DragAndDrop` as `TableDragAndDrop`. — cairo PR: _pending_
- [ ] **U2 (optional) — three paging/range types are undocumented while their siblings are documented.** `ComputedQuery`, `DataResultRaw`, `CollectionState`, `TableColumn` and `CtaBaseProps` all ship doc entries; `RangeItem`, `CursorQuery` and `OffsetQuery` do not, and those are the three the eval agents actually needed. They appear only in passing prose (`RangeItem` in `NumberRangeFilter/overview.md`, `CursorQuery` in `ComputedQuery/description.md`, `OffsetQuery` in `getting-started.md`), so there is no entry to look up. `RangeItem` is also re-exported from `@wix/bex-core`, which is why guessing its path produces a deep `@wix/bex-core/dist/types/...` import. Skill-side workaround: **P2** (`patterns.cjs types`). Downgraded: `patterns.cjs types` already answers this derivably and correctly, so this is polish for humans reading Storybook, not a blocker. — cairo PR: _pending_
- [x] ~~**U3 — document each subpath's export surface.**~~ **Closed, not doing.** Derivable data: 31 hand-maintained pages that go stale on release, when the reader can derive it. Superseded by R11 plus the form note.
- [ ] **U4 — the props table sits behind the design prose.** `Table.md` puts `## API` at line 986 of 1,083; `ToolbarFilters.md` at line 1,554 of 1,580. Any consumer reading top-down, or trimming output, gets the prose and misses the contract — measured: an agent piping 7 docs through `head -300` reached **0** API tables. **P1** and **R9** reorder around this on the skill side, but the template ordering is upstream's. — cairo PR: _pending_
- [ ] **U5 — case-colliding doc keys silently overwrite one doc with the other.** Worse than first recorded: not an ambiguous lookup but **data loss**. In published 1.448.0 `DateRangeFilter.md` and `dateRangeFilter.md` are the same 14,966 bytes — the component's doc — so the factory's documentation is not in the package at all, and macOS builds produce the opposite winner. Fixed in cairo #5779 with category-qualified filenames; the factory doc returns at 2,351 bytes. — cairo PR: wix-private/cairo#5779


### Closed at skill level — the upstream fix supersedes these

Cairo is being fixed, so building these here would ship a workaround for a defect that is about to disappear. **Don't pick them up in the skill.** If the cairo work slips, reopen them.

- ~~**R4** — print both candidates on an ambiguous case-insensitive hit.~~ **Superseded by U5.** Exact-match lookup already works today, and once the colliding key is renamed upstream there is nothing left to disambiguate. 
- [x] **R11 — `patterns.cjs exports <subpath>`.** Shipped in 33f6f3c3. **Reopened after being wrongly closed** as superseded by U1+U3: an export surface is *derivable* from the package, so the reader is its right home — 31 hand-written subpath pages would go stale every release. U1 by contrast is *authored* data only the package can assert.
  <details><summary>narrower than first claimed</summary>

  131 of 167 docs already carry a `@wix/patterns` import covering 22 subpaths, so `docs <Name>` answers the common question ("where does `CollectionPage` come from?") and `exports page` adds little. The real gap is `/form`: of its 162 names only **14** appear anywhere in the docs — `useForm` in 10, `useController` in 2, `FieldValues` in none — because it re-exports `@wix/bex-core/form`, which wraps `react-hook-form`. That was 4 of the employee-shift arm's 14 reads. The one-sentence note in COLLECTION_TOOLKIT.md (fdcb0) covers the need more cheaply than the command does.
  </details>


- [ ] **U6 — public API with no doc entry.** `withDashboard` appears in the *generated code's* imports and has no entry in `index.json`; `PageProps` likewise. Neither can be looked up, so the agent reads `node_modules` — 2 of the 15 hand-reads in the final run. Same class as U2: these are exported and used, just undocumented. Worth auditing the public surface against the index rather than fixing two names. — cairo PR: _pending_

## Housekeeping

- [x] **H1 — the internal Google Doc link is gone**, replaced with a prose credit. It was unreachable for anyone outside Wix and this repo is public. — fix: 48102f60

---

*Disclosure: 28ed51ed and 62c02fa7 on this branch are mine, pushed before/while reviewing. Iteration 2 was measured on `origin/main` + 62c02fa7 against `origin/main` alone, identical harness, `@wix/patterns@1.448.0`, `claude-sonnet-4-6`, n=1 per arm.*

kobiburnley (dismissed):

kobiburnley:
#skipreview

kobiburnley:
#skipreview

kobiburnley (approved):

shaharezr (approved):

PR #1108, opened 2026-08-25
docs(wix-app): route admin-scoped SDK calls out of frontend extensions

adimara:
## Problem

An Editor React component cannot call `locations.archiveLocation`. Its only scope is `SCOPE.DC-MULTILOCATION.MANAGE-LOCATIONS` — app-level — and the component runs as a site visitor or member. `auth.elevate` doesn't rescue it either: per [About Elevated Permissions](https://dev.wix.com/docs/api-reference/articles/authentication/about-elevated-permissions.md), *"methods can only be elevated in backend code."*

The skill said none of this. `auth.elevate` appeared only in `SERVICE_PLUGIN.md`, `BACKEND_EVENT.md`, `DATA_COLLECTION.md`, and `data-collection/WIX_DATA.md` — all contexts that are already backend by construction. `EDITOR_REACT_COMPONENT.md` and its entire `references/editor-react-component/` subtree mention permissions zero times. So an agent building a frontend extension had nothing telling it to route the call out.

## Why the gate keys off extension type

Two more obvious signals were tested and both fail:

- **The method's reference page doesn't carry identity restrictions.** The machine-readable (`.md`) view lists **Permission Scopes** only — verified on `archive-location` and `bulk-update-products` — even though [About Identities](https://dev.wix.com/docs/api-reference/articles/authentication/about-identities) claims the reference notes them.
- **An "(with elevated permissions)" example is not a discriminator.** `queryLocations`, a plain read, has one too.

What *is* reliable: frontend extensions inherit the site user's identity and never run as the app; only Backend API, Backend Event, and Service Plugin do. The gate decides from that.

## Changes

`skills/wix-app/SKILL.md` (428 → 447 lines):

- New `## Identity and Elevation Requirement` section, placed beside the existing Stores Versioning gate. Routes the call into a Backend API endpoint, requires that endpoint to do **its own authorization check** (elevation bypasses Wix's, so an unguarded endpoint is a real hole), and sends the Dev Center scope to Manual Steps.
- One ANTI-PATTERNS row and one Step 3 checklist sub-bullet.
- A closing note not to infer this from the docs page, with the reason.

`yaml/wix-app-evals/editor-react-component/admin-scoped-call-routing.yml` — new scenario, tagged `[editor-react-component, backend-api]`. Asserts the component never calls `auth.elevate` or `archiveLocation` directly, an endpoint under `src/pages/api/` elevates and guards, and the component reaches it over `fetch`. Two `llm_judge`s: one on the output, one on the path taken.

## Scope

Deliberately `SKILL.md` only. `EDITOR_REACT_COMPONENT.md` and `references/editor-react-component/**` are synced in from a wix-private repo (`sync(editor-react-component): sync changes from wix-private PR #NN`), so edits there would be overwritten. The section names site plugins, custom element widgets, and embedded scripts explicitly rather than adding per-file pointers.

## Verification

Locally: YAML parses, name unique and well-formed, all 7 assertion types valid per `docs/eval-scenarios.md`, both `referenceFiles` paths exist, both markdown anchors resolve, `check-truncation.mjs` doesn't flag SKILL.md.

Not yet verified: the EvalForge gate. Because this touches `SKILL.md` it runs the **whole** suite rather than a tag subset — the new scenario's assertions are unproven until CI runs here.


adimara:
Reviewed, verified independently, and fixed in 54db4fe1. All three blocking findings confirmed — and #2 was worse than reported.

**#2 (scope-name over-fire) — confirmed, and it was breaking-not-just-noisy.** `addLineItemsToCurrentCart` carries *both* `SCOPE.ECOM.WRITE-CARTS-V2` and `SCOPE.ECOM.MANAGE-ADMIN`, so it matched all three of my trigger patterns. The Cart API introduction states Current Cart "manage[s] the current customer's cart using their **session identity** (visitor ID or member ID)" — so elevating retargets the app's cart. The rule converted working code into broken code, exactly as you said.

**#3 (under-fire) — confirmed, and it hit this PR's own scenario.** `queryLocations` is `READ-LOCATIONS`, never matched, yet needs elevation. The scenario's component would have rendered an empty list and still scored well. Good catch.

**#4 — confirmed.** `EDITOR_REACT_COMPONENT.md:62` says the scaffold writes to `src/extensions/site/components/<folder>/`, "not under `src/site/components/`". The hard gate was pointed at the avoided path.

**#5 — confirmed.** `BACKEND_API.md:181` prescribes `httpClient.fetchWithAuth()`. A bare `fetch` sends no identity, which also hollowed out criterion (4).

**#1 — accepted without needing the external citation.** The row contradicted rule 1 in the same section. Scoped to site/editor extensions.

**The fix your review made possible:** the SDK schema-line prefix is a clean discriminator, and it's in the machine-readable docs — so my closing note telling agents to ignore the reference page was wrong on a stronger basis than #7 argued. Dropped it. Verified across four methods:

| Method | Prefix | Scope | Frontend-callable |
|---|---|---|---|
| `archiveLocation` | `wixClientAdmin` | `MANAGE-LOCATIONS` | no |
| `queryLocations` | `wixClientAdmin` | `READ-LOCATIONS` | no |
| `getCurrentCart` | `wixClientMember` | `READ-CARTS-V2` | yes |
| `updateLineItemsInCurrentCart` | `wixClientMember` | `WRITE-CARTS-V2` | yes |

Gets right both cases the scope heuristic got wrong. Also added an anti-pattern row against elevating `wixClientMember`/`wixClientVisitor` methods, and the eval now asserts the read is routed too.

**#6 — partly addressed, deliberately not "solved."** `fetchWithAuth` gets the caller's identity to the endpoint, but your point stands that there's no documented way for a site extension's endpoint to assert "the caller is the site owner." Rather than prescribe an invented mechanism, the requirement now names what it must establish, and the judge explicitly accepts an agent that reports finding no sound check over one that fabricates one. Your suggestion that this should be a dashboard extension is probably right on the merits; keeping the scenario as-is because it's the case that exposed the gap, but worth revisiting.

**#7 — resolved by the rewrite** rather than by softening.

adimara:
Verified and fixed in 38e7176e. Finding #1 is correct and blocking, and my round-1 verification was methodologically wrong — worth stating precisely, because the error is more instructive than the conclusion.

I fetched the `.md` view for `archiveLocation`/`queryLocations`, but read the **MCP search prose summary** for the cart methods, then presented the two retrieval paths as one table. Re-tested properly with `curl … | grep -o 'wixClient[A-Za-z]*'`:

| Method | `.md` token |
|---|---|
| `get-current-cart` | `wixClientAdmin` |
| `update-line-items-in-current-cart` | `wixClientAdmin` |
| `archive-location` | `wixClientAdmin` |
| `query-locations` | `wixClientAdmin` |
| `create-order` | `wixClientAdmin` |
| `bulk-update-products` | `wixClientAdmin` |

6/6, four families, including both methods the section cited as `wixClientMember`. Zero signal. Your root cause is right — it's a generated placeholder for a self-managed client variable, which is why it renders differently by retrieval path.

**#2 and #3 confirmed as consequences.** Executed against `.md`, the trigger applied to every `currentCart*` method, so it commanded exactly what the anti-pattern row two screens above forbade — and condemned all three calls in `SITE_PLUGIN.md:326-336` ("Site plugins can import and use Wix SDK modules directly"). I'd reintroduced the cart breakage through a new mechanism.

**#4 confirmed** — judge 2 demanded the agent have seen `wixClientAdmin`, penalizing the run that used the lookup path `SKILL.md` prescribes.

**On the third basis: taking your recommendation.** Three signals have now failed — scope names, client prefix, elevated-example presence (your `create-order` = 0 / `items.query` = 1 counterexamples kill that one too, and I'm not going to try a fourth). The section is now default-deny with a named allowlist: Wix Data `items.*` under per-collection permissions, public catalog reads, and session-resolved `current*`/`my*` methods are called directly and never elevated; anything a site extension can't confirm is visitor/member-callable gets routed out. It makes a weaker promise than a mechanical trigger, and the section now says so rather than implying precision it doesn't have.

**#5–#8 all applied:** dashboard carve-out qualified with the "some methods may still require elevation" caveat and the role-limited-collaborator case; `current*`/`my*` framed by session-resolution rather than a glob that matched neither `addLineItemsToCurrentCart` nor `getCurrentMember`, with `getCurrentMemberAccount` added; `fetchWithAuth` pinned to `new URL(import.meta.url).origin`; judge 1 (4) now requires engaging the question and scores naming the gap level with finding a mechanism; judge 2 (3) no longer demands `wix generate` when the scaffold exists, and now asks the grader to name the docs gap behind each friction.

Two of your notes I did not act on: the missing MCP/docs-gap framing in judge 2 is now addressed, but the `products` vs `productsV3` split in the allowlist I resolved by naming both rather than picking one. Flagging that the underlying docs defect — `About Identities` claiming "the API reference for each method notes these restrictions" when no `.md` page carries them — is worth filing upstream regardless of this PR.

adimara:
Verified and fixed in 914dc0d0. #1 is the finding that matters and you're right that it's structural, not another missing entry.

**#1 confirmed.** `cartV2.placeOrder` — `SCOPE.ECOM.WRITE-CARTS-V2_LIMITED`, description "Completes checkout and creates an order. This endpoint may charge the customer", no elevated example. Not `current*`/`my*`, not Wix Data, not a catalog read, so it fell into "everything else" and got routed and elevated — detaching a charge from the buyer. And the section listed booking confirmation as route-out while `createBooking` is the method Wix's own elevation article cites as *not* needing it.

The diagnosis is the part I'd missed across three rounds: I was enumerating instances of a property I never stated. A fifth family would not have ended it. The rule is now the property — a method that acts on behalf of the current visitor or member (cart, checkout, booking, order, reservation, profile) is called directly and never elevated — with the named methods demoted to examples and an explicit "this is a property, not a list."

**#4 confirmed, and I had it backwards.** `get-my-member.md:25` in the raw `.md`:

> `>This method requires [visitor or member authentication](…/about-identities.md).`

`confirm-booking.md`: zero matches. So "no field tells you" was overstated in a way that told agents to skip the one real signal. Narrowed to: no *structured* field, a minority state it in prose, that note is authoritative when present, absence means unconfirmed → route out. (I checked this one in the raw `.md` rather than an MCP summary — the round-1 lesson.)

**#3 confirmed.** `DATA_COLLECTION.md:48` scaffolds `itemInsert`/`itemUpdate`/`itemRemove` as `PRIVILEGED`, and `:207` warns permissions must match the access context. The bullet's own parenthetical named the condition while its placement made it unconditional; a widget write would be refused with no hint why. Promoted to its own conditional case, with the fix being permissions rather than routing — and you were right that the widget eval was more correct than the line it tested.

**#2 confirmed and fixed.** Criterion 3 now judges only the two Locations operations and states plainly that "dashboard extension" does not mean "never route," since some methods require elevation even from a dashboard and elevation only works in backend code. It was training the opposite of rule 1's own hedge.

**#5 confirmed** — three scenarios carried tags they never graded, so a change to the tagged file re-ran a scenario structurally unable to detect a regression in it. Added `referenceFiles` assertions for `WIX_PATTERNS_DOCS.md`, `DATA_COLLECTION.md`, `DASHBOARD_PAGE.md`, plus a patterns clause to the dashboard judge. `backend-api` stays behaviorally graded by design.

**#6, #7, #8 applied:** catalog reads qualified to base fields (`MERCHANT_DATA` and non-visible products need `PRODUCT_READ_ADMIN`); stale `wixClientAdmin` dropped from the scenario description; and the normative `new URL(import.meta.url).origin` removed, since I can't cite it against the CLI docs' plain relative path.

**Gap I'm leaving open and flagging:** `placeOrder`/`createBooking` get no dedicated scenario — the widget scenario guards the same property through `addLineItemsToCurrentCart`. If the suite budget allows another, a "Buy now" widget is the highest-value addition, since it's the exact shape three iterations of this rule got wrong.

adimara:
Verified and fixed in f1dcafbc.

**#1 confirmed verbatim.** `get-member.md:26`: "Private members aren't returned in the response if a site visitor or site member identity calls this method." Scope `READ-MEMBERS`, no prose note, not `current*`/`my*` — so it fell through to route-out, and elevating strips a filter the platform was applying per caller. You're right that the existing safeguard can't catch it: "re-check whatever the elevated call assumes about the caller" is inert when the elevated call assumes nothing, because the *unelevated* call was the enforcement.

And your framing is the fix. Wix Data and catalog reads were the same mechanism the whole time — I had them filed as "configuration" when they're a third category. The partition is now three-way (acts for the caller / filtered by caller / acts on the business), which subsumes both carve-outs and fixes `getMember` without a fourth family. That's the second time this review has found the missing abstraction rather than another missing entry.

**#2 confirmed, and the path signal survives the round-2 test.** I checked the raw `.md` before adopting it:

```
request-cancellation.md  URL: https://www.wixapis.com/pricing-plans/v2/member/orders/{id}/cancel
get-current-cart.md      URL: https://www.wixapis.com/ecom/v2/carts/current
```

Machine-readable, unlike the client prefix. It's now the *first* check, above the prose note. The asymmetry is stated too — the default fires exactly where the property failed to decide, and a wrong route-out is a correctness or privacy bug rather than a wasted hop.

**#3 — added the one-directional sentence.** It's a fair read that "some pages do state it" could be taken as "check and you'll know."

**#4 confirmed on both counts.** The catalog base-fields clause was normative and untested, and the widget scenario would indeed have passed a `MERCHANT_DATA` request on every criterion while rendering empty for a visitor. Added a criterion. And the `getMember` class now has its own scenario — `custom-element-widget/filtered-read-not-elevated`, a "meet our members" directory where elevating a member read scores 1 or below. Without it the misclassification would have shipped green, as you said.

**#5 confirmed.** Raised to 300s / $2 and added the missing `wix-patterns-docs` tag plus assertion — the inverse of the gap the previous commit closed, and I should have caught it there.

**#6, second item:** I think this one's a miss. The description at `914dc0d0` already read "admin-only Locations methods" — the surviving `wixClientAdmin` is in judge 2's body, where it's deliberate: the criterion penalizes an agent for *using* the prefix as a signal, so it has to name it. Happy to reword if it reads as residue. First item is fair — the suite is 6 scenarios, 5 of them new.

Suite state: 6 scenarios, every tag mapping to a graded reference except `backend-api` (behavioral by design, per your #5 in round 3), SKILL.md at 465/500.

adimara:
All five applied in 1d244cf0. Verified each first.

**#1 confirmed — and it's the one that mattered.** The restructure stopped at the section. `SKILL.md:26` still read "routed every SDK call not confirmed visitor/member-callable," and `getMember` is visitor-callable but not *confirmed* so — its page states a privacy filter, not an authentication requirement — so the checklist routed exactly what the new bucket exists to protect. Table now has the missing bucket-2 row too. Good catch that the bucket with the privacy consequence was the one absent from the highest-salience block.

**#3 confirmed, your prediction included.** Swept the methods that broke earlier versions:

```
placeOrder      /ecom/v2/carts/{cartId}/place-order       no segment
getCheckoutUrl  /ecom/v2/carts/{cartId}/get-checkout-url  no segment
```

Precision fine, recall poor, and it misses precisely the methods that killed version 3. Demoted to corroboration beside the prose note, both marked one-directional.

**#4 confirmed.** `/members/v1/members/{id}` is `getMember` — a substring read of `/member` swallows the explicit non-match. Now specifies whole-segment matching and names that collision.

**#5 confirmed, and it would have passed green.** "Meet our members" reads as a CMS team page, and this skill's own Data Collection Inference section actively pushes toward inferring a collection — a well-built CMS directory would have satisfied every criterion while never meeting the privacy filter. Added criterion (0) scoring that 3 or below, and the prompt now asks for the site's actual members.

**#6 both applied.** Bucket 2's header no longer generalizes past the catalog bullet, since `MERCHANT_DATA` and non-visible products need `PRODUCT_READ_ADMIN` alongside elevation. `getCurrentMember` → `members.getMyMember`.

**On the bucket-1/bucket-2 boundary:** thank you for testing it and reporting that it doesn't need defending. "Both buckets prescribe the identical action; only the stated reason differs, so the only consequential boundary is 1-or-2 vs 3" is the kind of finding that saves the next round from manufacturing work.

**On #2, agreed as a follow-up, and the fix nudged it the wrong way** — adding the recall caveat took the section from 624 to 673 words. Your Stores Versioning comparison is the right model, and so is the caveat: the original defect was that agents had *nothing*, and a reference nobody opens recreates that. The Patterns Docs Gate pattern — a MANDATORY checklist bullet naming the file — is the precondition, not a nice-to-have. Happy to open that as a tracked issue against this repo if you'd like it on record rather than living in this thread.

mirivoWix (approved):

PR #1117, opened 2026-08-25
feat(wix-app): SDK-first rule — never model existing Wix app data as CMS (CODEAI-1065)

mirivoWix:
## Why

[CODEAI-1065](https://wix.atlassian.net/browse/CODEAI-1065): codegen sometimes creates a CMS data collection for data that already lives in a Wix business app — e.g. a "site refunds" dashboard built on a new (empty) collection instead of `@wix/ecom`, with zero MCP calls before deciding. The dashboard then shows an empty state while real refunded orders exist.

## What

`skills/wix-app/SKILL.md` — new **SDK-First Rule** section:

- Entity → `@wix/*` SDK module map for all Wix business domains, inlined in the section. Every package name verified against the live docs (Wix MCP) and npm (`wixSdkBackendModule` keyword).
- Rule: if the entity is Wix-owned use its SDK module, never CMS; if unsure, search MCP first — never conclude CMS with zero MCP calls. CMS is only for app-owned data.
- Wired into the existing flow: 🛑 gate in the Step 1 checklist, an anti-pattern row, a guard on Data Collection Inference, and an orders/refunds row in the Step 3 discovery table.

`yaml/wix-app-evals/refunds-dashboard-sdk-not-cms.yml` — eval reproducing the ticket: asserts `DATA_COLLECTION.md` is not read (tagged `data-collection` as a negative dependency), and an LLM judge fails any run that creates a refunds/orders collection.

## Testing

3 scenarios run with the new skill vs a pre-change snapshot:

- refunds dashboard (ticket repro) → `@wix/ecom`, no CMS ✅
- bookings overview → `@wix/bookings`, no CMS ✅
- fee-rules control (app-owned config, CMS is correct) → collection still created, namespace-scoped — no over-blocking ✅


skills/wix-app/SKILL.md
```diff
@@ -197,10 +199,31 @@ Use a Dashboard Modal only for dialogs that are genuinely not entity editing: a
...
... / payment links / receipts → `@wix/get-paid` · gift cards → `@wix/gift-vouchers` · coupons → `@wix/marketing` · pricing plans / subscriptions → `@wix/pricing-plans` · bookings / services / staff / time slots → `@wix/bookings` · calendar events / schedules → `@wix/calendar` · table reservations → `@wix/table-reservations` · restaurant menus / online orders → `@wix/restaurants` · blog posts → `@wix/blog` · site events / tickets / RSVPs → `@wix/events` · reviews → `@wix/reviews` · comments → `@wix/comments` · groups → `@wix/groups` · online programs → `@wix/online-programs` · donations → `@wix/donations` · portfolio → `@wix/portfolio` · media files → `@wix/media` · contacts / labels / tasks → `@wix/crm` · members → `@wix/members` · inbox conversations → `@wix/inbox` · forms / form submissions → `@wix/forms` · loyalty points / rewards → `@wix/loyalty` · email marketing → `@wix/email-marketing` · notifications → `@wix/notifications` · analytics → `@wix/analytics-data` · automations → `@wix/automations` · SEO tags / redirects → `@wix/seo` · site search → `@wix/search` · secrets → `@wix/secrets` · locations / site properties → `@wix/business-tools` · app instances → `@wix/app-management`
```
adimara: One thing worth to reconsider:

The entity→module map here is one big paragraph — 34 mapping separated by · — not a table. All other lookup structure in this file (❌/✅ anti-pattern table just below, Extension Types Reference Table, discovery table) use markdown table. For something that need to be scan by entity name, table (Entity | Package | Note) will be much more easy to read than parsing long line of middle-dots, and also this not consistent with file's own style. Minor thing, not blocker — but easy to fix and will improve readability a lot.

adimara (dismissed):

skills/wix-app/SKILL.md
mirivoWix: Done in 82e3a2e4 — converted to an Entity | SDK package table (notes folded into the package column for the two rows that had them).

PR #1132, opened 2026-08-26
feat(yaml-gate): validate docsEntry points at a docs menu category

guyshaulsky:
## Why

A `docsEntry` pointing at an individual API page passes today's checks (they only verify the yaml and the eval scenario agree with each other) and silently fails after merge: the skill never gets a menu node and its `<docsEntry>/skills/<slug>` URL 404s. Three recent skills shipped this way (#1034, #1036, #1075).

## What

The gate now validates every **new or repointed** `docsEntry` against the live docs menu (two anonymous GETs to the public docs-site API) and fails unless it resolves to a **category** node — the same condition the docs pipeline enforces when placing the skill after merge. The failure comment names the problem and suggests the category that groups the API.

Unchanged entries are never validated, and if the docs endpoints are unavailable the check logs a warning and passes.

Also documents the rule in CONTRIBUTING step 3.

Live demos: #1144 (fail → fix → pass), #1146.


guyshaulsky:
Live demo of this check: #1144 — commit 1 (docsEntry → an API page) drew the ❌ **Invalid docsEntry** comment with the correct category suggestion; commit 2 (the suggested category) passed and the gate proceeded to the next validation (Missing Coverage, expected for a mock with no scenario). Closed without merge.

CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ When adding a `wix-manage` skill:
 
 1. Add the skill markdown under `skills/wix-manage/references/<area>/<skill>.md`.
 2. Add a short entry to the relevant section in `skills/wix-manage/SKILL.md`.
-3. Add the skill to `yaml/wix-manage/<area>/documentation.yaml`.
+3. Add the skill to `yaml/wix-manage/<area>/documentation.yaml`, with `docsEntry` pointing at a **category** in the docs menu (use the "Copy Docs Entry" button), not an individual API page.
```
danshw: maybe we should be extra specific and warn not to point it to a resource item?

danshw (dismissed):

CONTRIBUTING.md
guyshaulsky: changed it

PR #1136, opened 2026-08-26
fix: wix-headless-fast — cart line images via catalog join; seed images required+verified

ayal:
User-reported: the cart drawer shows a gray box instead of the product image.

**Diagnosis (empirical, live cart):** Cart V2 returns line `attributes` **without `image`** at runtime — `{descriptionLines, url, policies, itemType, physicalProperties, membersOnly}` — even though the SDK type declares `image?: string`. No client rendering could show what's never sent. Separately, some runs seeded **fully imageless stores** because SEED.md made `imageUrl` read as optional.

**Fixes:**
- `cart.ts`: backfill line images from the catalog — one `queryProducts().in("_id", ids)` keyed by `source.catalogReference.catalogItemId`, session-cached, failure-tolerant. Join query + `wix:image://` resolution verified against a live store; strict tsc passes.
- `storefront.css`: un-glue the Remove link from the qty stepper (screenshot nit).
- `SEED.md`: an `imageUrl` per product is now the expectation, with "verify each URL resolves before seeding" (agents have seeded hallucinated URLs); text-only only on explicit user request.


ayal:
Superseded by the Tailwind design-system PR, which includes this commit.

PR #1144, opened 2026-08-27
DO NOT MERGE — docsEntry gate demo

guyshaulsky:
Throwaway demo for the docsEntry validation added in #1132 (this branch includes it).

Adds a mock skill whose `docsEntry` points at an API page (`…/seo/redirects/redirect-v1`, a RESOURCE node). Expected: the gate fails with **Invalid docsEntry** and suggests the category (`…/seo/redirects`).

A follow-up commit will fix the value to show the check passing (the gate then proceeds to Missing Coverage — expected, the mock has no eval scenario on purpose).

Will be closed without merging.


guyshaulsky:
Demo complete — both halves verified live: commit 1 (RESOURCE target) drew the Invalid docsEntry comment with the correct category suggestion; commit 2 (the suggested category) passed the check and the gate proceeded to Missing Coverage as expected. Closing without merge.

PR #1152, opened 2026-08-27
feat(wix-vibe-headless): forms vertical

tadaskay:
Builds the client half of the `forms` vertical in `wix-vibe-headless`, on top of the seed doc that landed earlier on this branch.

## What this ships

A form is schema-driven — the owner picks the fields in their dashboard — so no "contact form" component can ship for it. This ships the non-visual half and lets the agent write the markup:

| file | what it is |
|---|---|
| `lib/wix-form-schema-utils.js` | accessors over the **raw** field Wix returns (`orderedInputs`, `labelOf`, `validationOf`, `choicesOf`, `addressPartsOf`, …). No projection to keep in sync, nothing the schema carries hidden. The two options blocks are named after the field's own enums, which is what the lookup tables in here resolve. |
| `hooks/useWixForm.js` | controlled `values` seeded from the schema's defaults, `bind(target)` for per-control wiring, schema-driven `validate(target?)`, `submit`, and visitor-facing copy for server violations. |
| `rest/wix-forms.js` / `rest/wix-forms-submissions.js` | schema read and submission write, including the file flow: Forms-scoped upload URL → `PUT` the bytes → submit that same URL as the value. |
| `references/forms/INSTRUCTIONS.md` | guidelines, not a component to copy: which control each field kind wants, which schema key feeds which attribute, and the traps. 73 of 471 lines are code. |

## Verified

Schema projection, layout ordering, label resolution through derived option blocks, all three submission value shapes, phone E.164 normalization, numeric bounds, address subfield errors, `bind()` output, and the file-upload flow were each exercised against fixtures built from the seed's own field JSON.

Two things are inferred rather than observed, and are flagged as such in the doc: no live 200 confirmed for `media-upload-url` on an anonymous visitor token (same scope listing as the schema read, which does work anonymously), and the multi-file array shape is extrapolated from Wix's single-file example.

### DIFF
--- a/skills/wix-headless/references/inline-recipes/how-to-code-forms.md
+++ b/skills/wix-headless/references/inline-recipes/how-to-code-forms.md
@@ -456,7 +456,7 @@ async function submitContact(formEl) {
     formId: SEEDED_FORM_ID,          // from seeded.forms[].formId
     submissions: values,             // map of target -> value (or -> nested object, for an address)
   });
-  return { _id, status };            // CONFIRMED → render a thank-you (see the status note below)
+  return { _id, status };            // CONFIRMED or PENDING → render a thank-you (see the status note below)
 }
 ```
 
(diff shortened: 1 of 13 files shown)


ayal:
Reviewed and pushed one fix to the branch — `orderedInputs` was interleaving multi-step forms.

It flattened every step's `layout.items` and then sorted the whole list by `row`/`column`. `row` restarts at 0 in each step, so a two-step form came out interleaved:

```
expected: step1_first, step1_second, step2_first, step2_second
got:      step1_first, step2_first, step1_second, step2_second
```

Sorting inside each step and concatenating in step order fixes it. Single-step forms — everything the seed builds — were unaffected, which is presumably why it went unnoticed. Verified against three fixtures: two-step ordering, single-step row-then-column, and a field the owner never placed still sorting last.

Two things I did **not** change, for the record:

- **Files re-upload on retry.** `uploadFormFiles` returns a copy and never writes back to state, so after a server validation rejection the next submit re-uploads the same files, leaving orphans in the owner's Media Manager. Waste rather than corruption, and fixing it is a state-design call that belongs to you.
- Your two disclosed unknowns are accurate and worth leaving visible in the docs — no live 200 for `media-upload-url` on a visitor token, and the multi-file array shape extrapolated from a single-file example.

The rest reviewed clean. `err.body` is genuinely attached by the shared client, so `submissionViolations` resolves; deriving every rule from the schema rather than field names is the right call; and the routing carve-outs between `forms` / `events` / `bookings` / `cms` are unambiguous.

PR #1165, opened 2026-08-29
fix: align wix-headless and wix-headless-fast on the facts

ayal:
Audited the two skills against each other, the live dev.wix.com docs, and the installed SDK typings. **They agreed on everything the fast skill verified live** — what turned up was the classic recipes going stale where the platform moved, plus one-sided gaps in both directions. 37 files, all fixes, no restructuring.

## wix-headless — the recipe taught a bug

| where | was | now |
|---|---|---|
| store | `searchProducts({filter, fields})` | `fields` is the **2nd arg** — inside the search object it's ignored → no CURRENCY → the unformatted-`34.99` bug the recipe itself warns about |
| store | `directCategoriesInfo` | `allCategoriesInfo` — the direct field returns **zero** products on a nested parent page |
| store | "`getProduct` returns `variantsInfo: null`" | it returns variants with `VARIANT_OPTION_CHOICE_NAMES`; the mandated extra round-trip is gone |
| store | "`plainDescription` is a plain string" | it's **HTML** — rendering as text printed literal tags |
| store | cart images via `attributes.image` | V2 doesn't populate it → the catalog join (this was our gray-box bug) |
| cms | "RICH_TEXT and RICH_CONTENT are both Ricos" | RICH_TEXT is HTML — the recipe contradicted its own seed recipe |
| events | flat `queryEvents(...)` → `.events` | a query **builder** → `.items` (the only overload in the installed `.d.ts`), `fields` in its options |
| plans | `purchaseLimits: {type, count}` | an array of `{type, maxCount}`; free plans must omit `billingCycle` (`FREE_PRICING_VARIANT_IS_NOT_RECURRING`) |
| plans | "⚠️ VERIFY IN A LIVE BUILD" for purchase | the verified `paidPlansCheckout` redirect (visitor-safe) |
| bookings | "a fresh install always has the owner resource" | it provisions **asynchronously** — poll before failing |
| blog | `returnUrl=` ×2 | `returnToUrl=` (same silent-drop bug fixed in the members recipe last week) |

Plus: split field masks (`MEDIA_ITEMS_INFO` or a gallery has one image), the three `availabilityStatus` values + `preorderStatus`, cart update/remove contracts, `lineItem.status`, the TEXT_CHOICES/FREE_TEXT modifier split, `@wix/categories`, the `all-products` system category, variant `price.actualPrice`, `x-wix-forwarded-url` for SEO, `registration.type ?? initialType`, `thankYouPageUrl` vs `postFlowUrl`, Ricos `description`, `queryAvailableTickets`' 0 default, RSVP `WAITLIST`, `referencingFieldKey`, DATE shapes, blog cursor/taxonomy/body shapes, portfolio `item.type`/`details` one-of/`coverVideo`/`sortOrder`, members' `contact.firstName`, and the Members-Area login loop.

**Cross-cutting:** image cost is per-model from the response's own `cost` field (~0.009 / 0.03 / 0.138 — not "1 credit"); the model ladder leads with `runware:400@1`; a new **§2b local-file upload** (`generate-upload-url` → `PUT`, the extension-match trap, a 200 isn't "processed", >10MB needs resumable); verbatim release URL; the ~5–30s post-release serving window; member login needs a published site; sibling-skills table.

## wix-headless-fast — gaps the classic skill already knew

- **`seed-blog.mjs` retries the transient `401 "No identity found"`** on draft-post create — classic documents it as a known async-identity defect. A run-killer we hadn't hit yet.
- **`seed-portfolio.mjs`** declares each file's real dimensions instead of a hardcoded 1024 square (wrong for any non-square `imagePath`); `images.mjs` now carries known dims out with the file.
- `SKILL.md`: `CI=1` on build/release (the skill stated the convention then broke it), the release retry ladder, a `forms` routing row, and the install line for a fallback recipe read — its 20 `wix-headless/references/…` citations assume that skill is on disk.
- `shared/DESIGN.md`: create-run scope carve-out, stock-photo clichés, the themed-block fallback, and the deliberate no-page-image policy.
- One-liners: bookings multi-location slot duplication, `BUSINESS→OWNER_BUSINESS`, https origin, pricing-plans needs a published site, `@wix/site-members` is Developer Preview, why blog is read-only.

## Deliberate divergences, now stated rather than implicit

Images default **on** in fast (entity images, no cap) and **off** in classic (gated, capped, cost surfaced); fast is strictly additive where classic deletes install demo data; fast ships no page-image path.

## Notes

- Two spec errors were caught during editing and corrected against the sources: `queryProducts` takes `fields` inside its single options object (only `searchProducts` has the 2nd arg), and the query builder's default limit is **50**, not 0.
- The one remaining "VERIFY IN A LIVE BUILD" hedge (membership-covered bookings) is left — genuinely untested.
- No yaml entry or eval scenario: CONTRIBUTING scopes that gate to `wix-manage` and `wix-app`.


ayal:
Parked — the audit findings and all fixes are preserved on the `headless-skills-alignment` branch; reopen or re-cut from there when we pick this back up.

PR #1179, opened 2026-08-31
test(google-business-profile): verify concurrent dummy skill publishing

hadarzaguri:
## Summary

- add two disposable dummy skills with fixed Alpha/Beta markers
- attach both skill sources to the same docsEntry in one change
- add wix-manage index entries and individual eval coverage
- make no API calls and provide no Google Business Profile guidance

## Verification target

Exercises concurrent skill-source creation under:
https://dev.wix.com/docs/api-reference/business-management/google-business-profile

Portal resource ID: 7904c177-abeb-4b9d-b21b-01cabd01b500

This is a draft verification PR for the reported menu-node race condition and is not intended as product documentation.

## Checks

- YAML parsing passed
- canonical documentation URLs match eval assertions
- descriptions are under 1,024 characters
- git diff check passed


hadarzaguri:
#rebuild

PR #1188, opened 2026-08-31
docs(wix-headless): rentals recipe corrections from a live run

noakorkus:
## Summary

Follow-up to #1178, all in `how-to-code-rentals.md`. Four corrections, three from reading the merged text again and one from tracing a live failure back into the platform code.

## Changes

- **§1 wording** — "fully-booked" → "unavailable" rentals. Matches the flag name (`includeUnavailable`) and the response field (`available: false`); a rental can be unavailable for reasons other than being booked.

- **§2 end-options cap** — the cap defaults to `localStartDate` plus the service's maximum duration, but what actually comes back also depends on availability: the options stop earlier if the resource isn't free that long, so the last end option is whichever comes first.

- **§3 and Out of scope — `multiServiceBookingInfo.id`** — the recipe said it isn't readable back off the individual bookings, so a site that didn't store it had "no way to cancel the group". It's readable off any booking in the group, so that warning is gone. Keeping the id from the create response stays the recommended path (one read at create beats a lookup later).

- **The anonymous read-back is removed throughout.** `getAnonymousActionToken` carries the `Manage Bookings` scope — its own docs say *"This is the only method in this API that requires authentication"* — so the design is that an authenticated caller mints the token and hands it to the visitor. The recipe was routing rentals at guidance that has a visitor mint it client-side, which 403s. Removed from the delta header, from §5's delegation, and the Out-of-scope note also drops its "anonymous visitor lacks admin scope" reasoning, which turned out not to be the real cause.

## Why that last one isn't just a scope problem

Worth recording, because it affects more than this recipe. An **elevated** call also fails, with `BOOKING_NOT_FOUND`, and the cause is in the platform rather than the caller:

`getAnonymousActionToken` → `bookingsAdapter.queryBookings(bookingId)` → `bookingsReader.queryExtendedBookings`, and every query path there is rewritten by `QueryAppIdFilterEnricher.addAppFilterIfNeeded(query, appIdsToFilter, …)` with a hardcoded allowlist (`BookingsReader.scala:359–361`):

```scala
val AppDefId        = "13d21c63-b5ec-5912-8397-c3a5ddb27a97"  // Wix Bookings
val BookInFormAppId = "3e4e322b-6d9b-4da0-9f53-e22c4a44a49e"
val appIdsToFilter  = Seq(AppDefId, BookInFormAppId)
```

A rentals booking (`appId = ff5d6eb1-…`) isn't in that list, so the reader silently returns an empty page and the method raises its declared `BOOKING_NOT_FOUND`. `shouldAddAppFilter` skips the injection only for a cursor page, a query that already filters on `appId`/`createdByAppId`, or a platform caller — elevation isn't an escape hatch. `POST /bookings/v2/bookings/query` works because that's bookings-service, which has no such enricher.

Same mechanism affects `queryExtendedBookings` generally for rentals, and the whole `anonymous-bookings` / `anonymous-multi-service-bookings` endpoint family. Not something this PR can fix — flagging it for whoever owns the reader, since it will hit Wix Meetings and Wix Services the same way.

## Feature Toggle ([create release](https://wix-bo.com/dev/tool/feature-toggle/new-feature-release))

N/A — documentation only.

## Rollback

`git revert`. One markdown file, no dependents.

## Testing

- [x] Scope requirement confirmed on the [method's own reference](https://dev.wix.com/docs/api-reference/business-solutions/bookings/bookings/bookings-writer-v2/get-anonymous-action-token.md) — `Manage Bookings: SCOPE.DC-BOOKINGS.MANAGE-BOOKINGS`.
- [x] `BOOKING_NOT_FOUND` is a declared error on the rpc (`anonymous_booking_actions.proto:189`); the appId allowlist traced through `BookingsReader` and `QueryAppIdFilterEnricher`.
- [x] `multiServiceBookingInfo` confirmed present on the Booking entity (`booking.proto:363`).
- [ ] The `endOptions` availability nuance in §2 is reasoned from the cap semantics, not observed on a live site.

## Known dependency

The `multiServiceBookingInfo` change assumes the field becomes externally readable — it's currently `field_exposure = INTERNAL` (scheduler PR #31830 removes that). Note that the exposure change alone may not be sufficient: reading it back goes through the same `queryExtendedBookings` reader described above, so the appId allowlist may still block a rentals caller.

### DIFF
--- a/skills/wix-headless/references/inline-recipes/how-to-code-rentals.md
+++ b/skills/wix-headless/references/inline-recipes/how-to-code-rentals.md
@@ -229,12 +229,12 @@ Everything else about the three-step item-page SEO wiring (`wixMetadata` export 
 
 ## Out of scope
 
-**Cancellation and post-booking self-service.** Out of scope here, as in the bookings recipe. Two rentals-specific notes if a run does build them server-side:
+**Cancellation.** A logged-in member can cancel their own rental:
 
 - A **single** rental (hourly, or daily on a 24/7 resource) cancels with `cancelBooking` and its current `revision`.
 - A **multi-day group** (daily on a working-hours resource) cancels with `cancelMultiServiceBooking` and the group's **`multiServiceBookingInfo.id`**, read back off any booking in the group.
 
-Refunds are the eCommerce Orders API, not Bookings. Also out of scope, as in bookings: waitlists, deposit/payment breakdowns, and multi-item rental carts.
+Refunds are the eCommerce Orders API, not Bookings. Out of scope, as in bookings: waitlists, deposit/payment breakdowns, and multi-item rental carts.
 
 ## Conclusion
 
--- a/skills/wix-headless/references/inline-recipes/how-to-code-rentals.md
+++ b/skills/wix-headless/references/inline-recipes/how-to-code-rentals.md
@@ -172,11 +172,24 @@ Hourly is prorated per minute (`minutes × base ÷ 60`); daily is `base × days`
 >
 > Two things make this misleading. **The same `createBooking` succeeds as an admin (`WIX_USER`)**, so it reads like a permissions or payload bug when it's neither — always test the visitor path. And **switching the service to offline / pay-in-person does not avoid it**: `placeOrder` still requires the site to be accepting payments. Record the precondition and surface it to the owner; don't try to code around it.
 
-**Identical to `how-to-code-bookings.md` § "createBooking → cart → checkout"**, with three substitutions:
+**Identical to `how-to-code-bookings.md` § "createBooking → cart → checkout"**, with five substitutions:
 
 1. **`endDate` is the customer's chosen end**, not a duration added to the start — that is the whole point of a rental.
 2. **`resource`** comes from the **start slot's** `availableResources` (§2 step 1 / §3). There is no ANY_RESOURCE staff fallback here; rentals are resource-driven.
-3. **The cart's `catalogReference.appId` is the RENTALS app id**, not the Bookings one:
+3. **`how-to-code-bookings.md`'s `locationType` mapping is missing a row.** Use this 3-way table instead of its 2-way one:
+
+   | Slot's `locationType` (read side) | `createBooking`'s `locationType` (write side) |
+   |---|---|
+   | `BUSINESS` | `OWNER_BUSINESS` |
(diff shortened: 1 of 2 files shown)


ayal:
@noakorkus two "you say X, but Y" notes:

**1.** You say (Cancellation section) the group's `multiServiceBookingInfo.id` is *"read back off any booking in the group."* Verified — correct: it's a readOnly, queryable field on the booking read model (`booking.proto` #31, `bookings_reader.proto:321`), so the old "must persist at creation" was wrong. 👍

**2.** But the recipe doesn't say *which* endpoint to read it from — and it only works on one. I created a real rentals booking and queried both:
- `POST https://www.wixapis.com/bookings/v2/bookings/query` → **1** ✅
- `POST https://www.wixapis.com/bookings/bookings-reader/v2/extended-bookings/query` → **0** ❌

The same `BookingsReader` allowlist (359–361) you flagged for `getAnonymousActionToken` drops rentals across the **whole reader / `queryExtendedBookings` path**, not just the anonymous token. So "read it back off any booking in the group" holds **only via `/bookings/v2/bookings/query`** — if the cancel step reads through the reader it gets nothing. Worth pinning that endpoint in the recipe. Otherwise LGTM.

noakorkus:
On #2 — I don't want to pin `/bookings/v2/bookings/query`: that rpc is `exposure = PRIVATE, maturity = ALPHA` (`bookings.proto:907`).

And the reader isn't actually dropping rentals. `QueryAppIdFilterEnricher` injects a *default* `appId $in [Bookings, BookInForm] OR appId $exists:false` filter, but **only when the incoming query has no `appId`/`createdByAppId` filter of its own** — that's your 0. Name `appId` in the filter and the injection is skipped, and `queryExtendedBookings` returns rentals bookings fine. `booking.appId` is inherited from the service's appId (`IdentityHelper:56`), so rentals bookings do carry `ff5d6eb1-65e4-4f9a-8b14-64d34c12cc2e`.

So the fix is the rule the recipe already applies to catalog reads — **filter bookings reads by the rentals appId** — not a different endpoint. Your underlying point was right though: the cancellation step named neither the endpoint nor the filter. Fixed in https://github.com/wix/skills/pull/1188/commits/e43314cc0efb6e43be6705e4aab950b39e3a29de — it now pins `queryExtendedBookings` with `filter: { appId: RENTALS_APP_ID }` and calls out the silent-0 failure mode.

PR #1215, opened 2026-09-03
revert(ci): stop enforcing check-truncation.mjs on CI

kobiburnley:
Reverts the CI enforcement of `scripts/check-truncation.mjs` that #1100 introduced.

Removed from `.github/workflows/ci.yml`:
- the `skills — changed reference files fit the fetch limit` step in `typecheck`, which ran `check-truncation.mjs --fail --files=…`
- the `changed_refs` output on the `detect` job, its `BASE_REF` env var, and the block that diffed against the base branch tip to collect changed `skills/*/references/*.md` files
- `changed_refs != ''` from the `setup-node` step condition (back to wix_app / evalforge only)

The workflow file is now byte-identical to its pre-#1100 content (blob `ed1079de`), so nothing else in it drifted.

`scripts/check-truncation.mjs` is deliberately left untouched: its `--files=` and `--fail` flags are inert with nothing invoking them, and the plain `node scripts/check-truncation.mjs` report still runs and exits 0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,6 @@ jobs:
     outputs:
       wix_app: ${{ steps.changes.outputs.wix_app }}
       evalforge: ${{ steps.changes.outputs.evalforge }}
-      changed_refs: ${{ steps.changes.outputs.changed_refs }}
     steps:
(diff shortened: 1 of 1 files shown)


kobiburnley:
Superseded by #1217, which scopes the gate to the wix-app skill instead of reverting the enforcement outright.

PR #1216, opened 2026-09-03
feat(seo): add content plan generation recipe

YZaritsky:
## What

Adds one `wix-manage` recipe for the public **SEO Content Plan APIs**: **Generate and Read a Wix Site's Content Plan**.

The API is PUBLIC BETA, routed on all surfaces, and documented on dev.wix.com since 2026-08-30 (REST at `www.wixapis.com/promote/seo/v1/*`). This PR adds the agent-consumer artifact so Wix MCP and BM Aria can execute the user flow.

Docs entries:
- https://dev.wix.com/docs/api-reference/business-management/seo/content-plan-content-plan-flow-v1
- https://dev.wix.com/docs/api-reference/business-management/seo/content-plan-blog-post-candidate-v1
- https://dev.wix.com/docs/api-reference/business-management/seo/content-plan-keyword-research-v1

## The recipe teaches

- The trigger-poll-release-poll-read loop (the flow parks at KEYWORD_RESEARCH and only CreateContentPlan releases it)
- Recognizing and surfacing PENDING_REQUIREMENTS (site lacks business data)
- Copy-on-write keyword edits: always use the keywordResearchId from the response
- That edits are not durable across generations
- That marking candidates is not part of this public API
- Stop after the first 403 and name the missing Manage SEO Settings authorization

## Evaluation coverage

- `manage-content-plan.yml`: explains an existing KEYWORD_RESEARCH state and the required release step.
- `manage-content-plan-happy-path.yml`: requires trigger, observe KEYWORD_RESEARCH, explicitly release, observe SUCCESS, and read real blog post candidates for the same flow ID. Timeouts and prerequisite failures do not satisfy this assertion.
- The recipe specifies response wrappers and bounded status checks so API polling can report progress.

## Current validation and blocker

At the tested head `3cf8f2ae6447bc9704fb80323ca20413358bc57f`, CI tests, typecheck and CLI version checks pass. The knowledge scenario passes all three assertions. The end-to-end scenario is not yet passing.

Backend logs from the earlier timed-out EvalForge site show that trigger created an `AGENT` flow, then repeated status reads continued with no state transition. The created-event consumer in current `wix-private/promote-seo` master starts keyword research only for `seo_editor3`, excluding the externally derived `AGENT` origin. The current evaluation also reports a new flow stuck in CREATED. The targeted backend fix is https://github.com/wix-private/promote-seo/pull/15308 (draft). Its regression fails before the fix and passes afterward; all five consumer/origin tests, build, lint, and bundle closure pass. It must be reviewed and rolled out before rerunning this end-to-end acceptance test.

Head `d06364d` additionally prevents diagnosing CREATED alone as missing category/description and corrects the prerequisite guidance. The strict generation assertions remain unchanged.

The happy-path fixture's prerequisite setup also needs verification: the backend validates business location, while the current bootstrap sets category and description only.


skills/wix-manage/references/seo/manage-content-plan.md
```diff
@@ -0,0 +1,110 @@
...
+
+Returns `{ "contentPlanFlowId": "..." }`. Hold this ID.
+
+### 2. Poll until KEYWORD_RESEARCH
+
+```
+GET /content-plan-flows/{contentPlanFlowId}
+```
+
+Status walks: `CREATED` → `SITE_ANALYSIS` → `KEYWORD_RESEARCH`. Poll every
```
hadarzaguri: Status walks - > Typical status progression?

hadarzaguri (dismissed):
Nice addition overall - one gap I’d like to close is in the eval coverage.

The recipe’s most important behavior is that the flow parks at `KEYWORD_RESEARCH` and must be explicitly released with `POST /create-content-plan`. The current eval seems to pass if the agent triggers the flow and checks status at least once, and it also treats `PENDING_REQUIREMENTS` as a valid expected outcome. That’s useful for the blank-site path, but it means we’re not actually validating the release step on a successful path.

Could we add coverage for a happy-path scenario that requires:
1. trigger the flow,
2. poll until `KEYWORD_RESEARCH`,
3. call `POST /create-content-plan`,
4. poll until `SUCCESS`,
5. read the blog post candidates?

Without that, an implementation could still pass while missing the main workflow detail this recipe is trying to teach.

YZaritsky:
Re-running gate — previous two runs timed out at 30 min due to slow eval backend, not code changes. The same scenario scored 10/8/10/9 on commit 02bd504.

skills/wix-manage/references/seo/manage-content-plan.md
```diff
@@ -0,0 +1,247 @@
...
+name: "Generate and Read a Wix Site's Content Plan"
+description: Generate an SEO content plan and read its blog post topics, or troubleshoot an existing content plan flow stuck at KEYWORD_RESEARCH while polling GetContentPlanFlow. Use this recipe for both generation and stalled-flow questions: it explains the intentional pause, the Create Content Plan release request, missing flow IDs, and the exact public API paths and response fields.
+---
+
+# Generate and Read a Wix Site's Content Plan
+
+## Choose the request path before calling an API
+
+For a request to **generate a new plan**, follow the generation sequence below.
+For a request about an **existing flow**, use this branch instead; do not start
```
hadarzaguri, line 11: what existing flow mean?
did we test both cases? Think of evals as tests: you should cover everything you write in a skill. If you don't, we didn't test what the skill covers.

skills/wix-manage/references/seo/manage-content-plan.md
```diff
@@ -0,0 +1,247 @@
...
+that timers or sleep functions exist inside an API execution sandbox.
+
+Continue across status checks while the flow is progressing. At
+`KEYWORD_RESEARCH`, release it immediately once, then keep checking that same
+flow until `SUCCESS` and read its candidates. Do not keep polling the parked
+state instead of releasing it, or stop with only a promise to finish later.
+If the client cannot continue waiting, report the flow ID and last observed
+status honestly; that is an incomplete generation, not success.
+
+## The exact call sequence
```
hadarzaguri, line 94: sequence - 
I think it was caused by calling callWixSite in a loop instead of making a single call, effectively creating a sequence rather than executing one call within a single block of code.

skills/wix-manage/references/seo/manage-content-plan.md
```diff
@@ -0,0 +1,247 @@
...
+The response is `{ "contentPlanFlow": { "id": "...", "status": "..." } }`.
+Read `contentPlanFlow.status`, not a top-level `status`. A missing status is a
+response-shape problem: inspect the response instead of silently looping.
+`CREATED` alone does not identify a missing prerequisite. If the flow remains
+there without progressing, report the stalled flow ID and observed status; do
+not invent a missing business category or description.
+Always send the trigger's ID when following this generation. The documented
+omitted-ID behavior selects a previous successful flow, not the active flow;
+it is not a way to discover the ID of a parked flow.
+See [Get Content Plan Flow](https://dev.wix.com/docs/api-reference/business-management/seo/content-plan-content-plan-flow-v1/get-content-plan-flow).
```
hadarzaguri, line 126: Make sure to add the entire response structure. What is status? enum?
Response shape is the easiest way for an agent to follow what you want instead of long sentences.

yaml/wix-manage-evals/seo/manage-content-plan-happy-path.yml
```diff
@@ -0,0 +1,69 @@
...
+      Require evidence of ALL steps in order for the same flow:
+      1. POST /content-plan-flows/trigger and capture the returned contentPlanFlowId.
+      2. GET /content-plan-flows/{id} until contentPlanFlow.status is KEYWORD_RESEARCH.
+      3. POST /create-content-plan with that ID and confirm success.
+      4. GET the same flow until status SUCCESS.
+      5. GET /content-plan-flows/{id}/blog-post-candidates and report actual returned topics.
+      Score 0 if any step is missing, if only an explanation/promise is given,
+      or if results are fabricated. PENDING_REQUIREMENTS, FAIL, or a timeout
+      must be reported honestly but do not pass this happy-path scenario.
+      Reading docs and triggering/polling alone are not completion.
```
hadarzaguri: It should be a prompt from a user who is not aware of APIs...
a simple one:
create a plan... etc

skills/wix-manage/references/seo/manage-content-plan.md
```diff
@@ -0,0 +1,247 @@
...
+
+### 3. Release the flow
+
+```
+POST /create-content-plan
+{ "contentPlanFlowId": "..." }
+```
+
+This advances the flow past `KEYWORD_RESEARCH`. Check `success` in the
+response. If `false`, read `message`.
```
hadarzaguri, line 147: What does the response look like? Is it just success: true/false? What if the agent wants to access the created content plan, but has no idea how? Do we have an evaluation that tests this?

skills/wix-manage/references/seo/manage-content-plan.md
```diff
@@ -0,0 +1,247 @@
...
+
+# Generate and Read a Wix Site's Content Plan
+
+## Choose the request path before calling an API
+
+For a request to **generate a new plan**, follow the generation sequence below.
+For a request about an **existing flow**, use this branch instead; do not start
+the generation sequence or make a discovery call to look for the flow.
+
+`KEYWORD_RESEARCH` is an intentional pause, even after ten minutes. Explain
```
hadarzaguri, line 14: `KEYWORD_RESEARCH` is an intentional pause, even after ten minutes. 

why did you start with that? KEYWORD_RESEARCH - we have no idea what does it mean when i read the skill from top to bottom?

hadarzaguri (approved):

PR #1224, opened 2026-09-03
docs(wix-app): props come from the bundle, not the doc

kobiburnley:
Consumes `@wix/patterns@1.460.0`, which ships cairo #5831.

## Why

The skill still teaches the old division of labour — *"the docs cover components and props, not the types those props use"*. As of 1.460.0 that is backwards for **74 of 167 docs**: their `### Props` is now a pointer at the type bundle, so an agent following the current guidance would read that pointer as a truncated doc and go looking elsewhere.

## What changed

`PATTERNS_BUNDLE_READING.md` gains a **Where props live** section, because the answer is per-doc rather than global:

| `dist/docs/index.json` entry | where the props are |
| --- | --- |
| has a `bundle` field | that bundle — the doc's `### Props` is a pointer to it, on purpose |
| no `bundle` field | the doc's own `### Props` table, as before |

Both shapes are common (74 point at a bundle, 62 still carry a table), so the doc says to expect both rather than implying a single rule. It also explains that the bundle is the *better* source: it keeps the `extends` clause with its exclusions, so `Omit<PopoverMenuItemProps, 'text' | 'prefixIcon' | 'onClick'>` states what you do **not** inherit — something the doc's old design-system link never did.

Two smaller additions:

- **Heavy example code now lives beside the doc.** A reader meets `Example code: read \`dist/docs/ToolbarFilters/apply-changes.tsx\`` and is told to follow it only for the variation they actually want, since the heading and description are usually enough to choose.
- **The hook-docs tip is qualified** — a hook's configuration options are in its bundle when it has one, and in the API table when it doesn't.

`WIX_PATTERNS_DOCS.md` keeps **1.458.0** as the hard gate (that is when the bundle index first shipped, and the check is for its presence) but now recommends **1.460.0**, since that is where the docs behave as described here.

## Verification

Every concrete claim checked against the **installed 1.460.0**, not a local build:

- the 74 / 62 / 167 split
- `dist/docs/ToolbarFilters/apply-changes.tsx` resolves, and the doc really uses the `Example code: read` wording
- the exact `Omit` clause (I had abbreviated it in a first draft and corrected it)
- a doc with a `bundle` field has no props table of its own

Also confirmed 1.460.0 ships what it should: 167 docs at 615KB, none over the ~50KB read limit (largest 18.6KB), 190 example files across 38 directories published, and all 74 index bundle pointers resolving.

### DIFF
--- a/skills/wix-app/references/WIX_PATTERNS_DOCS.md
+++ b/skills/wix-app/references/WIX_PATTERNS_DOCS.md
@@ -41,7 +41,7 @@ Then confirm the installed version actually ships the bundle index:
 ls <pkgRoot>/dist/dts-bundle/index.json
 ```
 
-**If it's missing, stop — do not look elsewhere for types or docs.** The installed `@wix/patterns` predates the index; it ships from **1.458.0** onward, so upgrade to at least that and re-run the check. A missing *file* is not the same as a name not being covered (see below): the mechanism itself isn't available yet.
+**If it's missing, stop — do not look elsewhere for types or docs.** The installed `@wix/patterns` predates the index; it ships from **1.458.0** onward, so upgrade to at least that and re-run the check. Prefer **1.460.0** or newer: from there the docs stop repeating props the bundle already describes, which is what the lookup below assumes. A missing *file* is not the same as a name not being covered (see below): the mechanism itself isn't available yet.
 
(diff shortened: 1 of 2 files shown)


kobiburnley:
#skipreview

shaharezr (approved):

PR #1225, opened 2026-09-03
fix(wix-manage): remove shipping API tool instruction

hadarzaguri:
## Summary
- remove the hard-coded `CallWixSiteAPI` instruction from the Shipping API Reference
- leave no replacement API-access or named-tool guidance

## Validation
- `git diff --check`
- verified the shipping API reference contains no API-access or named-tool instruction


hadarzaguri:
#forcemerge

PR #1232, opened 2026-09-04
fix(wix-app): check the project, not a file list

kobiburnley:
Audit item P0. `APP_VALIDATION.md` carried five per-file `tsc` recipes telling an agent to type-check only the files it had just generated. Both halves of that are wrong, and the skill now prescribes one form: `npx tsc --noEmit -p .`.

## What `-p .` actually buys

**Corrected after review** — an earlier version of this description claimed `-p .` was needed for correctness. It is not. At the app root, bare `tsc --noEmit` reads `tsconfig.json` exactly the same way. Reproduced on TS 5.8.3, matching @mirivoWix's 5.7.3 run:

| from | invocation | result |
|---|---|---|
| root | `tsc --noEmit` | `TS7006` |
| root | `tsc --noEmit -p .` | `TS7006` — **identical** |
| root | `tsc --noEmit a.ts` | **exit 0**, `tsconfig.json` silently discarded |
| root | `tsc --noEmit -p . a.ts` | `TS5042` |
| subdir | `tsc --noEmit` | checks `../a.ts` — walks up into the parent project |
| subdir | `tsc --noEmit -p .` | `TS5057` |

So `-p .` buys two guard rails, and those are what the skill now states:

1. **The footgun becomes a hard error instead of a silent pass.** An agent wanting to narrow the check will append a path. Appended to bare `tsc` that discards the config and checks against compiler defaults — exit 0, a false pass. Appended to `-p .` it is `TS5042`.
2. **It fails loudly from the wrong directory.** Silently checking a project you did not mean to check is the worse outcome in a monorepo.

(TS 7 adds `TS5112`, warning that the config will be ignored when files are specified — closing upstream, but silent on the 5.x these apps ship.)

## Why scoping to generated files is wrong even with the config right

A type error caused by generated code usually surfaces where that code is *consumed*. Measured on a real app: a scoped check over the generated directory reported nothing, while the project check reported `TS2322` in the consuming file. Scoping is only sound if you already know the error is confined to those files — which is what the check exists to find out.

Kept here rather than in the skill: it is the argument for the instruction, not the instruction.

## What changed

`APP_VALIDATION.md` — five per-file recipes deleted; one command, and a short statement of what it buys. The replacement text was rewritten after review: it had been arguing against globs in a file an agent loads at validation time, which reprinted the broken command, demonstrated a `TS5042` an agent following the instruction never reaches, and negated "targeted check" in a way that put the idea back in the reader's head. Now 6 lines instead of 24.

`SKILL.md`, `editor-react-component/MANIFEST-ERRORS.md` — the same bare `npx tsc --noEmit` occurrences updated.

## Not fixed here

`references/EDITOR_REACT_COMPONENT.md:66` still has a bare `npx tsc --noEmit`. An earlier version of this description listed it as fixed; it is not, and it is not fixable in this repo — the file is synced in from [wix-private/editor-react-component-creation-skills](https://github.com/wix-private/editor-react-component-creation-skills/tree/master#skill-sync--keeping-wixskills-up-to-date), so an edit here is overwritten on the next sync. It needs the components team to change it upstream. Tracking as a follow-up rather than leaving it assumed done, since that is the step where an agent actually runs the command mid-flow.

### DIFF
--- a/skills/wix-app/references/APP_VALIDATION.md
+++ b/skills/wix-app/references/APP_VALIDATION.md
@@ -66,9 +66,8 @@ And the config cannot be handed back on the side: `tsc -p tsconfig.json <file>` 
 **Checking only what you generated does not work even when the config is right.** A type error caused
 by generated code usually surfaces where that code is *consumed* — the page compiles and `App.tsx`
 does not. Measured: a scoped check over the generated directory reported nothing, while the project
-check reported `TS2322` in the consuming file. Scope hides the wiring errors, which are the ones an
-agent actually makes. It is also not worth much — the whole project takes about 1.3s on a real app,
-where 3,169 of the files in the program come from `node_modules` and 13 are yours.
+check reported `TS2322` in the consuming file. Scoping is only sound if you already know the error is
+confined to those files, which is what the check exists to find out.
 
 **Success criteria:**
 - Exit code 0
--- a/skills/wix-app/references/APP_VALIDATION.md
+++ b/skills/wix-app/references/APP_VALIDATION.md
@@ -47,16 +47,22 @@ pnpm install
 Run TypeScript compiler to check for type errors.
 
 ```bash
-npx tsc --noEmit
+npx tsc --noEmit -p .
 ```
 
-**This is the only form to use.** Passing filenames or globs on the command line makes TypeScript
-ignore `tsconfig.json` altogether — no `strict`, no `noImplicitAny`, no `jsx`, no `paths`. A targeted
-run therefore checks the code against *no* configuration: it misses the errors that matter and
-floods the output with errors that do not. Measured on a real app — a file with an untyped parameter
-reports `TS7006` under `tsc --noEmit`, and under `tsc --noEmit <file>` reports no `TS7006` at all,
-just dozens of `Cannot find name 'Set'` from `node_modules/@types` (no `lib`, no `skipLibCheck`).
-Always check the project, never a file list.
+**This is the only form to use, and there is no targeted variant.** Passing filenames or globs makes
+TypeScript ignore `tsconfig.json` altogether — no `strict`, no `noImplicitAny`, no `jsx`, no `paths`,
+no `lib`, no `skipLibCheck` — so the run checks the code against *no* configuration: it misses the
+errors that matter and floods the output with errors that do not. Measured on a real app, on a file
+with one untyped parameter:
+
+- `npx tsc --noEmit -p .` → `TS7006: Parameter 'x' implicitly has an 'any' type`
+- `npx tsc --noEmit <that file>` → no `TS7006` at all, and instead dozens of
+  `Cannot find name 'Set'` from `node_modules/@types`
+
+And the config cannot be handed back on the side: `tsc -p tsconfig.json <file>` fails outright with
+`TS5042: Option 'project' cannot be mixed with source files on a command line`. Checking a subset of
+files under the project's own settings is not something the compiler offers, so check the project.
 
 **Success criteria:**
 - Exit code 0
--- a/skills/wix-app/references/APP_VALIDATION.md
+++ b/skills/wix-app/references/APP_VALIDATION.md
@@ -46,41 +46,17 @@ pnpm install
 
 Run TypeScript compiler to check for type errors.
 
-**Full project check:**
 ```bash
 npx tsc --noEmit
 ```
 
-**Targeted check (specific files/directories):**
-
-When validating after implementing a specific extension, you can run TypeScript checks on just those files:
-
-```bash
-# Check specific directory
-npx tsc --noEmit src/extensions/dashboard/pages/survey/**/*.ts src/extensions/dashboard/pages/survey/**/*.tsx
-
-# Check dashboard pages only
-npx tsc --noEmit src/extensions/dashboard/pages/**/*.ts src/extensions/dashboard/pages/**/*.tsx
-
-# Check custom element widgets only
-npx tsc --noEmit src/extensions/site/widgets/**/*.ts src/extensions/site/widgets/**/*.tsx
-
-# Check dashboard modals only
-npx tsc --noEmit src/extensions/dashboard/modals/**/*.ts src/extensions/dashboard/modals/**/*.tsx
-
-# Check backend only
-npx tsc --noEmit src/extensions/backend/**/*.ts
-```
-
-**When to use targeted checks:**
-- After implementing a single extension (faster feedback)
-- When debugging type errors in a specific area
-- During iterative development
-
-**When to use full project check:**
-- Before final validation
-- When changes affect shared types
-- Before building/deploying
+**This is the only form to use.** Passing filenames or globs on the command line makes TypeScript
+ignore `tsconfig.json` altogether — no `strict`, no `noImplicitAny`, no `jsx`, no `paths`. A targeted
+run therefore checks the code against *no* configuration: it misses the errors that matter and
+floods the output with errors that do not. Measured on a real app — a file with an untyped parameter
+reports `TS7006` under `tsc --noEmit`, and under `tsc --noEmit <file>` reports no `TS7006` at all,
+just dozens of `Cannot find name 'Set'` from `node_modules/@types` (no `lib`, no `skipLibCheck`).
+Always check the project, never a file list.
 
 **Success criteria:**
 - Exit code 0
--- a/skills/wix-app/references/APP_VALIDATION.md
+++ b/skills/wix-app/references/APP_VALIDATION.md
@@ -61,8 +61,14 @@ with one untyped parameter:
   `Cannot find name 'Set'` from `node_modules/@types`
 
 And the config cannot be handed back on the side: `tsc -p tsconfig.json <file>` fails outright with
-`TS5042: Option 'project' cannot be mixed with source files on a command line`. Checking a subset of
-files under the project's own settings is not something the compiler offers, so check the project.
+`TS5042: Option 'project' cannot be mixed with source files on a command line`.
+
+**Checking only what you generated does not work even when the config is right.** A type error caused
+by generated code usually surfaces where that code is *consumed* — the page compiles and `App.tsx`
+does not. Measured: a scoped check over the generated directory reported nothing, while the project
+check reported `TS2322` in the consuming file. Scope hides the wiring errors, which are the ones an
+agent actually makes. It is also not worth much — the whole project takes about 1.3s on a real app,
+where 3,169 of the files in the program come from `node_modules` and 13 are yours.
 
 **Success criteria:**
 - Exit code 0
--- a/skills/wix-app/references/APP_VALIDATION.md
+++ b/skills/wix-app/references/APP_VALIDATION.md
@@ -144,7 +144,7 @@ Read: .wix/debug.log (with offset to the end)
 | Issue | Cause | Solution |
 |-------|-------|----------|
 | Package installation fails | Missing lock file, network issues, or corrupted node_modules | Delete `node_modules` and lock file, then reinstall |
-| TypeScript compilation fails | Type mismatches, missing declarations, or incorrect types | Fix TypeScript errors shown in `npx tsc --noEmit` output |
+| TypeScript compilation fails | Type mismatches, missing declarations, or incorrect types | Fix TypeScript errors shown in `npx tsc --noEmit -p .` output |
 | Build fails | TypeScript errors, missing dependencies, or internal CLI error | Fix TypeScript errors in source; for non-obvious failures, check `.wix/debug.log` |
 | Preview fails to start | Port conflict, config issue, or internal CLI error | Check `wix.config.json`; if unclear, check `.wix/debug.log` for details |
 | Console errors in preview | Runtime exceptions | Check browser console output |
--- a/skills/wix-app/references/APP_VALIDATION.md
+++ b/skills/wix-app/references/APP_VALIDATION.md
@@ -50,12 +50,11 @@ Run TypeScript compiler to check for type errors.
 npx tsc --noEmit -p .
 ```
 
-Check the whole project, not just the files you generated — a type error in generated
-code usually surfaces in the file that consumes it. `-p .` is what holds that line:
-adding a file path to it is an error, whereas adding one to a bare `npx tsc --noEmit`
-silently discards `tsconfig.json` (`strict`, `paths`, `jsx`) and checks against
-compiler defaults instead. Run it from the app root: there `-p .` fails outright if no
-`tsconfig.json` is present, where bare `tsc` would walk up and check the parent project.
+Run it from the app root, and check the whole project rather than just the files you
+generated — a type error in generated code usually surfaces in the file that consumes
+it. `-p .` is what holds that line: adding a file path to it is an error, whereas adding
+one to a bare `npx tsc --noEmit` silently discards `tsconfig.json` (`strict`, `paths`,
+`jsx`) and checks against compiler defaults instead.
 
 **Success criteria:**
 - Exit code 0
--- a/skills/wix-app/references/APP_VALIDATION.md
+++ b/skills/wix-app/references/APP_VALIDATION.md
@@ -50,24 +50,12 @@ Run TypeScript compiler to check for type errors.
 npx tsc --noEmit -p .
 ```
 
-**This is the only form to use, and there is no targeted variant.** Passing filenames or globs makes
-TypeScript ignore `tsconfig.json` altogether — no `strict`, no `noImplicitAny`, no `jsx`, no `paths`,
-no `lib`, no `skipLibCheck` — so the run checks the code against *no* configuration: it misses the
-errors that matter and floods the output with errors that do not. Measured on a real app, on a file
-with one untyped parameter:
-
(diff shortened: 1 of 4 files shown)


mirivoWix:
The substance here is right — `-p .` and deleting the glob recipes are both correct. My concern is the replacement text, not the change.

## The prose argues against a thing the PR just deleted

The two findings are real and belong in the skill: passing filenames drops `tsconfig.json`, and scoping hides errors in the *consumer* of generated code. But `APP_VALIDATION.md` now spends ~12 lines litigating the removal, in a reference file an agent loads at validation time. That reads as a rebuttal written for a reviewer rather than an instruction for the reader.

Three concrete costs:

1. **It reprints the broken command.** Line 56 still shows `npx tsc --noEmit <that file>`. The description defends this as "the illustration of the form that does not work" — but a skill that prints a command form is a skill that can have that form copied. A negative example carrying real syntax is the highest-risk kind: the agent sees a plausible command and the surrounding "don't" is one hop of attention away.

2. **"There is no targeted variant" reintroduces the idea.** Once the recipes are gone, nothing needs negating. Stating the prohibition is the only thing that puts "targeted check" back in the reader's head.

3. **"Measured on a real app" is provenance the reader can't act on.** It decays, and it justifies rather than instructs. Same for the `TS5042` demonstration — an agent told to run `-p .` never hits it.

## On the `-p` flag itself — worth stating accurately

For correctness at the app root the flag isn't needed; bare `tsc --noEmit` reads `tsconfig.json` too, exactly as the description says. Verified on TS 5.7.3:

```
from project root:  tsc --noEmit        ->  TS7006
from project root:  tsc --noEmit -p .   ->  TS7006   (identical)
                    tsc --noEmit a.ts   ->  exit 0, config ignored
```

What `-p .` actually buys is two guard rails, and they're the better thing to say in the skill:

- **The footgun becomes a hard error rather than a silent pass.** If an agent wants to narrow the check, the instinct is to append a path. Appended to bare `tsc` that silently discards the config and checks against compiler defaults (exit 0 above). Appended to `-p .` it's `TS5042`. Same instinct, loud failure instead of a false pass.
- **It fails loudly from the wrong directory.** From a subdirectory with no `tsconfig.json`, bare `tsc --noEmit` walks *up* and checks the parent's project; `-p .` gives `TS5057: Cannot find a tsconfig.json file at the specified directory: '.'`. Silently checking a project you didn't mean to check is the worse outcome in a monorepo.

(For context: TS 7 now emits `TS5112` warning that the config will be ignored when files are specified — so this is closing upstream, but it's silent on the 5.x these apps ship.)

## Suggested replacement for Step 2

Command, one positive sentence carrying both findings, then the existing success criteria:

```markdown
### Step 2: TypeScript Compilation Check

After all implementation is complete, typecheck the whole project from the app root:

```bash
npx tsc --noEmit -p .
```

Check the whole project, not just the files you generated — a type error in generated
code usually surfaces in the file that consumes it. `-p .` is what holds that line:
adding a file path to it is an error, whereas adding one to a bare `npx tsc --noEmit`
silently discards `tsconfig.json` (`strict`, `paths`, `jsx`) and checks against
compiler defaults instead.
```

Three lines instead of twelve, keeps both findings, and never shows the form that doesn't work. If someone proposes globs again later, the full rationale lives in this PR description — which is the right home for it.

## The `SKILL.md` and `MANIFEST-ERRORS.md` edits look good

No notes on those.

## One correction to the description

It lists `EDITOR_REACT_COMPONENT.md` among the six bare occurrences fixed, but that file isn't in the diff — `references/EDITOR_REACT_COMPONENT.md:66` still has bare `npx tsc --noEmit` in its generate-and-validate step.

That's not fixable in this PR: the file is synced in from [wix-private/editor-react-component-creation-skills](https://github.com/wix-private/editor-react-component-creation-skills/tree/master#skill-sync--keeping-wixskills-up-to-date), so an edit here gets overwritten. It needs the components team to make the change upstream. Worth dropping it from the description's list and noting it as a follow-up so it isn't assumed done — it's arguably the most important one, since that's the step where an agent actually runs the command mid-flow.

File: skills/wix-app/references/APP_VALIDATION.md [hunk 1]
```diff
@@ -46,41 +46,28 @@ pnpm install
...
 
-**When to use full project check:**
-- Before final validation
-- When changes affect shared types
-- Before building/deploying
+**Checking only what you generated does not work even when the config is right.** A type error caused
+by generated code usually surfaces where that code is *consumed* — the page compiles and `App.tsx`
+does not. Measured: a scoped check over the generated directory reported nothing, while the project
+check reported `TS2322` in the consuming file. Scoping is only sound if you already know the error is
+confined to those files, which is what the check exists to find out.
```
mirivoWix: This block argues against a thing the PR just deleted, in a file an agent loads at validation time. Both findings are real and belong in the skill, but as an instruction rather than a rebuttal — the full argument has a better home in the PR description, where it'll be found if someone proposes globs again.

Suggested: the command, plus one positive sentence carrying both findings.

```suggestion
Check the whole project, not just the files you generated — a type error in generated
code usually surfaces in the file that consumes it. `-p .` is what holds that line:
adding a file path to it is an error, whereas adding one to a bare `npx tsc --noEmit`
silently discards `tsconfig.json` (`strict`, `paths`, `jsx`) and checks against
compiler defaults instead.
```

skills/wix-app/references/APP_VALIDATION.md
```diff
@@ -46,41 +46,28 @@ pnpm install
...
 Run TypeScript compiler to check for type errors.
 
-**Full project check:**
 ```bash
-npx tsc --noEmit
+npx tsc --noEmit -p .
 ```
 
-**Targeted check (specific files/directories):**
+**This is the only form to use, and there is no targeted variant.** Passing filenames or globs makes
```
mirivoWix: "there is no targeted variant" is the only thing putting *targeted check* back in the reader's head. Once the recipes are gone there's nothing left to negate — the prohibition reintroduces the idea it's trying to remove.

skills/wix-app/references/APP_VALIDATION.md
```diff
@@ -46,41 +46,28 @@ pnpm install
...
+TypeScript ignore `tsconfig.json` altogether — no `strict`, no `noImplicitAny`, no `jsx`, no `paths`,
+no `lib`, no `skipLibCheck` — so the run checks the code against *no* configuration: it misses the
+errors that matter and floods the output with errors that do not. Measured on a real app, on a file
+with one untyped parameter:
 
-When validating after implementing a specific extension, you can run TypeScript checks on just those files:
+- `npx tsc --noEmit -p .` → `TS7006: Parameter 'x' implicitly has an 'any' type`
+- `npx tsc --noEmit <that file>` → no `TS7006` at all, and instead dozens of
```
mirivoWix: This reprints the broken command. The description defends it as "the illustration of the form that does not work", but a skill that prints a command form is a skill that can have that form copied — a negative example carrying real syntax is the highest-risk kind, since the surrounding "don't" is one hop of attention away.

skills/wix-app/references/APP_VALIDATION.md
```diff
@@ -46,41 +46,28 @@ pnpm install
...
-# Check backend only
-npx tsc --noEmit src/extensions/backend/**/*.ts
-```
-
-**When to use targeted checks:**
-- After implementing a single extension (faster feedback)
-- When debugging type errors in a specific area
-- During iterative development
+And the config cannot be handed back on the side: `tsc -p tsconfig.json <file>` fails outright with
+`TS5042: Option 'project' cannot be mixed with source files on a command line`.
```
mirivoWix: An agent told to run `-p .` never hits `TS5042` — this demonstrates a failure mode the instruction already prevents.

skills/wix-app/references/APP_VALIDATION.md
```diff
@@ -46,41 +46,28 @@ pnpm install
...
 
-**When to use full project check:**
-- Before final validation
-- When changes affect shared types
-- Before building/deploying
+**Checking only what you generated does not work even when the config is right.** A type error caused
+by generated code usually surfaces where that code is *consumed* — the page compiles and `App.tsx`
+does not. Measured: a scoped check over the generated directory reported nothing, while the project
```
mirivoWix: "Measured" / "Measured on a real app" (also L56) is provenance the reader can't act on, and it decays. It justifies rather than instructs — belongs in the PR description.

mirivoWix (commented):
The substance is right — `-p .` and deleting the glob recipes are both correct changes. My notes are about the replacement text, inline.

One thing worth stating accurately, since it affects how the skill should phrase this: **for correctness at the app root, `-p` isn't needed.** Bare `tsc --noEmit` reads `tsconfig.json` too, exactly as the description says. Verified on TS 5.7.3:

```
from project root:  tsc --noEmit        ->  TS7006
from project root:  tsc --noEmit -p .   ->  TS7006   (identical)
                    tsc --noEmit a.ts   ->  exit 0, config ignored
```

What `-p .` actually buys is two guard rails, and those are the better thing to put in the skill:

- **The footgun becomes a hard error instead of a silent pass.** If an agent wants to narrow the check, the instinct is to append a path. Appended to bare `tsc` that silently discards the config and checks against compiler defaults (exit 0 above). Appended to `-p .` it's `TS5042`. Same instinct, loud failure instead of a false pass.
- **It fails loudly from the wrong directory.** From a subdirectory with no `tsconfig.json`, bare `tsc --noEmit` walks *up* and checks the parent's project; `-p .` gives `TS5057: Cannot find a tsconfig.json file at the specified directory: '.'`. Silently checking a project you didn't mean to check is the worse outcome in a monorepo.

(FYI: TS 7 now emits `TS5112` warning that the config will be ignored when files are specified — closing upstream, but silent on the 5.x these apps ship.)

---

**On `EDITOR_REACT_COMPONENT.md`** — the description lists it among the six bare occurrences fixed, but it isn't in the diff and `references/EDITOR_REACT_COMPONENT.md:66` still has bare `npx tsc --noEmit`.

That one isn't fixable here: the file is synced in from [wix-private/editor-react-component-creation-skills](https://github.com/wix-private/editor-react-component-creation-skills/tree/master#skill-sync--keeping-wixskills-up-to-date), so an edit in this repo gets overwritten — it needs the components team to change it upstream. Worth dropping from the description's list and tracking as a follow-up rather than leaving it assumed done, since that's the step where an agent actually runs the command mid-flow.

The `SKILL.md` and `MANIFEST-ERRORS.md` edits look good, no notes.

skills/wix-app/references/APP_VALIDATION.md
kobiburnley: Applied your suggestion, plus one sentence for the second guard rail — say the word if that tips it back toward over-explaining and I will cut it:

> Check the whole project, not just the files you generated — a type error in generated code usually surfaces in the file that consumes it. `-p .` is what holds that line: adding a file path to it is an error, whereas adding one to a bare `npx tsc --noEmit` silently discards `tsconfig.json` (`strict`, `paths`, `jsx`) and checks against compiler defaults instead. Run it from the app root: there `-p .` fails outright if no `tsconfig.json` is present, where bare `tsc` would walk up and check the parent project.

24 lines down to 6. The measurements moved to the PR description, as you suggested.

skills/wix-app/references/APP_VALIDATION.md
kobiburnley: Right — the negation was load-bearing only while the recipes were still there, and they are gone. Removed; nothing in the section mentions a targeted variant now.

skills/wix-app/references/APP_VALIDATION.md
kobiburnley: Agreed, and the framing is the useful part: a negative example carrying real syntax is the highest-risk kind because the surrounding "don`t" is one hop of attention away. Removed. The only full command forms left in the file are the correct one.

skills/wix-app/references/APP_VALIDATION.md
kobiburnley: Correct — it demonstrated a failure the instruction already prevents. Removed. `TS5042` now appears only as the *reason* `-p .` is the safer form (appending a path to it errors instead of silently passing), not as a worked example.

skills/wix-app/references/APP_VALIDATION.md
kobiburnley: Removed, both here and at L56. The measurements are in the PR description now.

kobiburnley:
Thanks — this was a good catch on both counts, and the correctness one changed what the skill says.

## `-p .` is not needed for correctness — you're right

I had this wrong in the description, and the skill text was built on it. Reproduced your result on TS 5.8.3:

| from | invocation | result |
|---|---|---|
| root | `tsc --noEmit` | `TS7006` |
| root | `tsc --noEmit -p .` | `TS7006` — identical |
| root | `tsc --noEmit a.ts` | **exit 0**, config silently discarded |
| root | `tsc --noEmit -p . a.ts` | `TS5042` |
| subdir | `tsc --noEmit` | checks `../a.ts` — walks up into the parent |
| subdir | `tsc --noEmit -p .` | `TS5057` |

So the two guard rails are the real argument, and they're what the skill now states. The subdirectory one I hadn't considered at all — silently checking the parent project in a monorepo is the worse failure of the two, since it can pass while checking the wrong thing.

## On the text being a rebuttal rather than an instruction

Taken, and it explains all four of the other notes at once: once the recipes were deleted there was nothing left to argue with, so every sentence still arguing was reintroducing the idea it was trying to remove — the prohibition put *targeted check* back in the reader's head, the negative example printed copyable broken syntax, and the `TS5042` demo showed a failure the instruction already prevents. Your suggestion applied, with one added sentence for the second guard rail; cut it if that tips back toward over-explaining. 24 lines to 6.

The measurements moved to the PR description.

## `EDITOR_REACT_COMPONENT.md`

You're right, and thank you for chasing it — it is not in the diff, `references/EDITOR_REACT_COMPONENT.md:66` still has the bare command, and I listed it as fixed. It's now in a **Not fixed here** section of the description with the sync constraint and the reason it needs the components team upstream, rather than left assumed done.

That matches a call already made on our side: it isn't our ownership, so it stays untouched here.

mirivoWix (commented):
All five addressed — the block reads as an instruction now, and moving the measurements into the description put them where they're findable without an agent paying for them at validation time. Confirmed only two `tsc --noEmit` occurrences remain without `-p .` on the branch: the in-prose one at `APP_VALIDATION.md:55` (deliberate, it's the subject of the sentence) and `EDITOR_REACT_COMPONENT.md:66` (upstream). Checks green.

One thing to fix in the sentence that got added, inline — the guard rail it describes is stated backwards. You asked whether to cut it: cut the rationale, keep the directive.

On `EDITOR_REACT_COMPONENT.md` — agreed it stays untouched here, and the **Not fixed here** section is the right disposition. Worth noting that a description section isn't a tracked follow-up; a message to the components team is what actually closes it, since the next `@wix/patterns` audit will find the same line and re-derive the same conclusion.

skills/wix-app/references/APP_VALIDATION.md
```diff
@@ -46,41 +46,16 @@ pnpm install
...
-**When to use full project check:**
-- Before final validation
-- When changes affect shared types
-- Before building/deploying
+Check the whole project, not just the files you generated — a type error in generated
+code usually surfaces in the file that consumes it. `-p .` is what holds that line:
+adding a file path to it is an error, whereas adding one to a bare `npx tsc --noEmit`
+silently discards `tsconfig.json` (`strict`, `paths`, `jsx`) and checks against
+compiler defaults instead. Run it from the app root: there `-p .` fails outright if no
+`tsconfig.json` is present, where bare `tsc` would walk up and check the parent project.
```
mirivoWix: The added sentence inverts the guard rail. "Run it from the app root: **there** `-p .` fails outright if no `tsconfig.json` is present" — at the app root `tsconfig.json` *is* present, so the condition can never hold where the sentence puts it. The protection is the opposite: `-p .` fails **anywhere other than** the root. As written it advertises the safety net at the one location it never fires.

And to answer your question — yes, the rationale half tips back toward over-explaining. "Bare `tsc` would walk up and check the parent project" is a thing the agent can't act on once it's been told where to run. The directive is worth keeping; the reason isn't. Folding it into the first line does both:

```suggestion
Run it from the app root, and check the whole project rather than just the files you
generated — a type error in generated code usually surfaces in the file that consumes
it. `-p .` is what holds that line: adding a file path to it is an error, whereas adding
one to a bare `npx tsc --noEmit` silently discards `tsconfig.json` (`strict`, `paths`,
`jsx`) and checks against compiler defaults instead.
```

If you'd rather keep the second guard rail explicit, the correct form is *"Run it from the app root — `-p .` fails outright anywhere else"*, which is also shorter than what's there.

kobiburnley: Caught it — that is a straight inversion, and applied.

Worth noting the data was already in front of me: my probe showed `TS5057` from a **subdirectory** and `TS7006` at the root, which is exactly your point. I still wrote the condition as firing at the root. The measurement was right and the sentence describing it was backwards.

And thanks for answering the question directly — took the primary suggestion rather than the explicit-guard-rail variant. "Bare `tsc` would walk up and check the parent project" is not something the agent can act on once it has been told where to run, so the directive survives folded into the first line and the reason is gone. Now 5 lines.

The remaining guard rail is real but silent, which seems right: it protects an agent that ignores the directive, and it does not need to be documented to work.

PR #1247, opened 2026-09-06
wix-app: dashboard-page skill fixes + template-first draft template

shaharezr:
## Summary

Found and fixed via a headless test run of the dashboard-page skill against the dashboard-app-skill-playground harness (booking-activity V1 acceptance scenario), plus a new template built from the same testing loop.

- **Data-fidelity + oneof-shape rules** (`COLLECTION_TOOLKIT.md`): the test run sourced a table column from a generic-looking field instead of the joined entity an `Extended*` API response is meant to provide, and only handled one variant of a oneof-shaped field, silently blanking rows using the other variant. Neither mistake had a rule against it before this.

- **UX Completeness Self-Audit** (`SKILL.md` Step 4c): `SummaryBar` and `SidePanel`/`EntityPage` drill-in are already named as MANDATORY in `SKILL.md`'s top checklist, yet the same test run shipped a bare table with neither. The checklist was a stated intention, never re-verified against the code actually written — Step 4c is a mechanical re-read-the-file gate, distinct from and prior to compile/build validation.

- **`DRAFT_TEMPLATE.md`** — a verified starting skeleton for Collection/Entity/Settings dashboard pages, structured around four labeled cases (A: collection only, B: collection+entity, C: settings only, D: all three) with an explicit decision table, so the agent picks the smallest shape a prompt actually asked for instead of defaulting to the most complete one. Every snippet is copied from this session's installed `@wix/patterns` `dist/docs/*.md`, not written from memory, and all four cases were materialized into the playground against throwaway entities and run through `tsc` — which caught two wrong APIs before they shipped (`usePatternsNavigate` is exported from `@wix/patterns/router`, not the root package; `SidePanel` takes `title` on `SidePanel.Header`, not on itself). Documents a real runtime-only failure mode along the way: `PatternsReactRouter` needs `location` supplied manually in a Wix CLI app, and its own docs say skipping this passes both `tsc` and `wix build` silently, failing only when the page actually renders in a browser.

- **Template-first wiring**: `SKILL.md`'s mandatory checklist and Component Selection Order, and `DASHBOARD_PAGE.md`'s five worked examples, all now point at the matching `DRAFT_TEMPLATE.md` case as the first move, with the Patterns/Component Docs gates reframed as covering what the matching case doesn't already show — instead of building the page shell, provider nesting, or router wiring from scratch each time.

## Test plan

- [x] `tsc --noEmit` clean for all four template cases (A/B/C/D), materialized as throwaway extensions in `dashboard-app-skill-playground`
- [x] Headless `claude -p` test run against the booking-activity V1 acceptance scenario surfaced the defects these changes address
- [ ] Live browser verification of a generated page (blocked on Dev Center collaborator access to the playground's pinned dev app — tracked separately)


skills/wix-app/references/dashboard-page/TABLE_STATE.md
```diff
@@ -0,0 +1,133 @@
...
+a `SummaryBar` count, a badge total, a header figure — is computed once on first render, when the
+collection is still empty, and never recomputed. The table itself updates because it observes
+internally; your derived numbers sit at zero next to a table full of rows, which reads as a bug in
+the query rather than in the wiring.
+
+`useSelector` from `@wix/patterns` subscribes:
+
+```tsx
+const loadedCount = useSelector(() => state.keyedItems.length);
+const rows = useMemo(() => state.keyedItems.map((k) => k.item), [loadedCount]);
```
livnat-a: [P2] Subscribe to the values being derived, not only the row count

This leaves `rows` stale whenever a refresh replaces records without changing their count. I reproduced the example with React and MobX: replacing `[{ id: 'a', value: 10 }]` with `[{ id: 'b', value: 90 }]` leaves the rendered result at `a/10`, because the selector still returns 1 and the memo dependency is unchanged. Aggregates or badges derived from these rows can therefore disagree with the table after a refresh or record update. Select the actual primitive aggregate inside `useSelector`, or observe the row contents when deriving a row-based result; the length alone is not a sufficient dependency.

skills/wix-app/references/dashboard-page/DRAFT_TEMPLATE_CMS_COLLECTION.md
```diff
@@ -0,0 +1,215 @@
...
+`navigateToEntityPage({ path: '/new' })`, with `entity` omitted. That omission needs
+`@wix/patterns` **≥ 1.464.0**, which the pin in §1 already gives you; on an older install `entity`
+is typed required, and passing a placeholder to satisfy it is the bug rather than the fix. Every
+other navigation has a record — pass it, as `onRowClick` does above, and the entity header renders
+before the fetch resolves.
+
+Two rules carry over unchanged from the hand-wired path, and both are about what is *not* here:
+**no `SummaryBar`** unless the request asked for one, and **the row opens a page, never a panel** —
+`navigateToEntityPage` to §3's entity page either way; a display-only collection keeps that page and
+simply has nothing that writes.
```
livnat-a: [P2] Give read-only CMS collections a genuinely read-only detail route

For Case A, this instruction directs the agent to the editor in section 3 and explicitly rules out the read-only route later in section 4. However, section 3 renders `EntityPageFieldsCard` with editable controls, and `useEntityPage(source, ...)` derives its save handler from the source backend automatically; omitting a handwritten `onSave` does not disable writes. A display-only request therefore gets a working edit/save flow (or a permission failure on Save for a reader). Provide a separate read-only CMS detail example, or an explicitly supported configuration that disables editing and saving, instead of reusing this editor unchanged.

.github/wix-app-typecheck/extract.mjs
```diff
@@ -94,9 +118,20 @@ for (const mdFile of findMdFiles(refsDir)) {
   const rel = mdFile.slice(refsDir.length + 1).replace(/\//g, '__').replace(/\.md$/, '');
   writeFileSync(join(outDir, `${rel}.tsx`), combined);
   count++;
+
+  // One file per block, not the concatenation pass 1 uses: a person copies a single
+  // block into a single file, and concatenating unrelated blocks invents parse errors
+  // that the blocks themselves don't have.
+  blocks.forEach(({ code }, i) => {
+    const asRealCode = deplaceholder(code);
+    if (asRealCode === code) return;
```
livnat-a: [P2] Include the CMS template in the per-block typecheck pass

Selecting blocks only when placeholder substitution changes their text skips all four TSX blocks in `DRAFT_TEMPLATE_CMS_COLLECTION.md`: they use `FeatureCollectionPage`/`FeatureEntityPage` directly rather than `{Feature}`. They are not covered by pass 2 either, because concatenating them produces duplicate `useCmsSchemaSource` imports (TS2300), which causes the entire file to be excluded as structural. I checked the extractor: this template has 4 blocks and 0 selected for pass 3. Consequently the new schema-driven path can acquire API/type errors without failing this check. Include template blocks independently of whether substitution changes them and supply their shared declarations so the CMS examples receive semantic checking.

skills/wix-app/references/dashboard-page/DRAFT_TEMPLATE_CMS_COLLECTION.md
```diff
@@ -0,0 +1,215 @@
...
+root exports are the hand-wired versions and will not read a schema.
+
+```tsx
+import { CollectionPage } from '@wix/patterns/page';
+import { PrimaryActions } from '@wix/patterns';
+import { Table, useTableCollection } from '@wix/patterns/schema';
+import { usePatternsNavigate } from '@wix/patterns/router';
+import { useCmsSchemaSource } from '@wix/patterns-cms';
+
+export const FeatureCollectionPage: FC<{ onAddItem: () => void }> = ({ onAddItem }) => {
```
livnat-a: [P2] Wire the required Add handler into the shared router template

This component requires `onAddItem`, but section 4 instructs consumers to reuse the app shell/routes from `DRAFT_TEMPLATE_ROUTER.md` unchanged, where the collection route renders `<FeatureCollectionPage />` without that prop. Composing the documented CMS files therefore fails with TS2741 (`Property 'onAddItem' is missing in type '{}'`), which I reproduced with TypeScript; bypassing that check leaves the Add button without its navigation handler. Define `openNew` inside this component using its existing `navigateToEntityPage({ path: '/new' })`, or show a router child that supplies the required callback.

kobiburnley (commented):
Checked the `@wix/patterns/schema` parts against cairo. Four corrections, plus what's changing in the indices:

**1. "none appear in `dist/dts-bundle/index.json`" (DRAFT_TEMPLATE_CMS_COLLECTION.md).** `Table`, `useTableCollection`, `useEntityPage` and `EntityPage` *are* in the index, but as the **root** versions: `importPath: @wix/patterns`, with the hand-wired signatures. That's more misleading than being absent, since an agent that finds them gets the wrong API. `SchemaSource` was listed as `@wix/patterns/core`. `useSchemaSource` and `EntityPageFieldsCard` really are missing today. `useCmsSchemaSource` lives in another package.

**2. "check `dist/types/index.d.ts`".** That's the root barrel, and it has none of the schema names. Their declarations are in `dist/types/exports/schema.d.ts`.

**3. `tableSchemaSource` (SKILL.md, "plus `tableSchemaSource` and `EntityPageFieldsCard`").** This isn't a `/schema` export. It's a root-entry helper (`@wix/patterns`) that has no doc and no index entry. A schema page doesn't need it: the source comes from `useCmsSchemaSource` or `useSchemaSource`.

**4. The eval rubric's "no `columns` prop at all".** The schema `Table` takes an optional `columns` prop for extra columns (actions and the like). Those columns are shown first, and a column with a field's id replaces that field's column. "No hand-written *field* columns" is the accurate rule.

**What's changing in cairo:**
- wix-private/cairo#5941 (merged): a `useSchemaSource` doc page (`dist/docs/useSchemaSource.md`, `importPath: @wix/patterns/schema`) covering the three steps (source → hook → component), choosing a source (`useCmsSchemaSource` / `usePlatformizedSource` / `useSchemaSource`), `loadSchema`, and how it differs from the classic hooks.
- wix-private/cairo#5942: Collection Toolkit, Collection to Entity Flow and Composition and Providers point to `useSchemaSource` when the collection has a schema (it's in their `relatedComponents`). `useSchemaSource`, `SchemaSourceInput` and `EntityPageFieldsCard` get index entries under `@wix/patterns/schema`, and `SchemaSource` moves off `/core`.
- wix-private/cairo#5944: index keys stay plain export names, and every other entry point's version of a name is listed under a new `variants` field. For example, `Table.variants[0]` is `{ importPath: "@wix/patterns/schema", file: "components/schema/Table.d.ts", doc: "useSchemaSource.md" }`. `index.txt` gets one line per version, so `grep '^Table\t'` returns both. `EditableTable`, `LayoutSwitch` and their hooks are indexed as their schema versions. The change is additive, so nothing that reads the index today breaks.
- Next: `@wix/patterns-cms` names (`useCmsSchemaSource`, …) in the same index with a `package` field and the exact `@wix/patterns` pin.

Once those are released, WIX_PATTERNS_DOCS.md could read `variants` and choose by `importPath`. That would replace the template's "not in the index, check `dist/types`" workaround. I can send that as a follow-up PR once the cairo side ships.

mirivoWix (commented):
Reviewed the data-collection / dashboard-page changes. Left inline comments per file. Summary: the `usePatternsNavigate` and `SidePanel`/`SidePanel.Header` claims check out against the actual installed packages, and the repo's own CI typecheck harness passes clean on this PR's content — the technical substance looks solid. The one recurring issue is the "Split out of X.md" blockquote left at the top of 6 new files: that's process narration about this PR's own refactor, not topic content, and it'll read as stale/confusing to anyone who didn't just read this diff. Also flagging that skills/wix-app/SKILL.md is now ~590 lines, past the ~500-line guideline for an always-loaded SKILL.md body.

skills/wix-app/references/dashboard-page/DRAFT_TEMPLATE_COLLECTION.md
```diff
@@ -0,0 +1,164 @@
+# Draft Template — Collection page (Cases A, B and D)
+
+> Split out of [DRAFT_TEMPLATE.md](DRAFT_TEMPLATE.md) so the case chooser stays short. Pick your
```
mirivoWix: This "Split out of ..." note describes the refactor, not the topic — it only makes sense to someone who just read this PR's diff. It ships as content to every future reader/agent loading this file. Please move this kind of context into the PR description/commit message and drop the blockquote (same pattern repeats in DRAFT_TEMPLATE_SETTINGS.md, QUERY_AND_PAGING.md, data-collection/LIFECYCLE.md, data-collection/PERMISSIONS.md, and data-collection/SCHEMA.md — flagging once here, same comment applies to all).

skills/wix-app/references/dashboard-page/DRAFT_TEMPLATE_SETTINGS.md
```diff
@@ -0,0 +1,44 @@
+# Draft Template — Settings page (Cases C and D)
+
+> Split out of [DRAFT_TEMPLATE.md](DRAFT_TEMPLATE.md) so each file covers one case. Pick your case there first; come here only for Case C or D.
```
mirivoWix: Same "Split out of" process-narration issue as DRAFT_TEMPLATE_COLLECTION.md's line 3 — this describes how the file was produced, not what it covers. Drop it; the reasoning belongs in the PR description.

skills/wix-app/references/dashboard-page/QUERY_AND_PAGING.md
```diff
@@ -0,0 +1,182 @@
+# Querying and paging a Wix API from a collection page
+
+> Split out of [DATA_SOURCES.md](DATA_SOURCES.md) so each file covers one job: that one is about
```
mirivoWix: Same issue: "Split out of DATA_SOURCES.md so each file covers one job" is a note about this PR's refactor, not about querying/paging. Drop the blockquote.

skills/wix-app/references/data-collection/LIFECYCLE.md
```diff
@@ -0,0 +1,71 @@
+# Data Collection — getting it onto the site
+
+> Split out of [DATA_COLLECTION.md](../DATA_COLLECTION.md). Declaring a collection is not the same
```
mirivoWix: Same issue: "Split out of DATA_COLLECTION.md" narrates the edit rather than describing the lifecycle topic. Drop the blockquote — a one-line intro to what this file actually covers (release/update/CMS prerequisite/access scope) can replace it if an intro is wanted.

skills/wix-app/references/data-collection/PERMISSIONS.md
```diff
@@ -0,0 +1,48 @@
+# Data Collection — permissions
+
+> Split out of [DATA_COLLECTION.md](../DATA_COLLECTION.md). The scaffolded `dataPermissions` are
```
mirivoWix: Same issue as the other new files: "Split out of DATA_COLLECTION.md" is process narration, not permissions content. Drop it.

skills/wix-app/references/data-collection/SCHEMA.md
```diff
@@ -0,0 +1,180 @@
+# Data Collection — declaring the schema
+
+> Split out of [DATA_COLLECTION.md](../DATA_COLLECTION.md), which owns scaffolding, the file shape
```
mirivoWix: Same issue: "Split out of DATA_COLLECTION.md, which owns scaffolding..." is about this PR's file split, not about the schema. Drop the blockquote.

skills/wix-app/SKILL.md
```diff
@@ -440,6 +476,28 @@ Open every path returned in `newFiles` and replace stubbed handler bodies / UI /
 - ⚠️ MANDATORY when using WDS: Add `import "@wix/design-system/styles.global.css";` in the **main component** entry file (`page.tsx`, modal `.tsx`, etc.) — not in child/tab/helper files.
 - ⚠️ MANDATORY when using Data Collections: Use the EXACT collection ID from `idSuffix` (case-sensitive). If `idSuffix` is `"product-recommendations"`, use `<app-namespace>/product-recommendations` NOT `productRecommendations`.
 
+### Step 4c: UX Completeness Self-Audit
```
mirivoWix: This Step 4c addition is well-grounded (the fetchTotal/SummaryBar and location-guard failure modes it describes are real and specific), but it's ~25 lines added to an already-590-line SKILL.md — this file is now well past the ~500-line ideal ceiling for a SKILL.md body (it's always loaded once the skill triggers). Given the PR already extracts detail into reference files elsewhere (DATA_SOURCES.md, QUERY_AND_PAGING.md, TABLE_STATE.md), consider doing the same here: keep a short pointer + checklist item in SKILL.md and move the full self-audit prose into a reference file (or fold it into UX_SUCCESS_MODEL.md, which already covers this ground).

skills/wix-app/references/dashboard-page/DRAFT_TEMPLATE_ROUTER.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+`PatternsReactRoute`'s `type` is one of exactly `"collection" | "createEntity" | "editEntity" | "other"` — no dedicated settings value, so a settings route is `"other"`. Case B: delete the `/settings` route and the import above it — nothing else changes.
+
+`{Feature}CollectionPage` is [DRAFT_TEMPLATE_COLLECTION.md](DRAFT_TEMPLATE_COLLECTION.md)'s component, unchanged; `{Feature}SettingsPage` (Case D) is [DRAFT_TEMPLATE_SETTINGS.md](DRAFT_TEMPLATE_SETTINGS.md).
+
+## 2b. The collection side of the handoff
+
+The router file above registers `/new`; something has to navigate there. All three of these live on
+the collection page, and `usePatternsNavigate` comes from **`@wix/patterns/router`** — not the root
```
mirivoWix: Verified against the pinned @wix/patterns version in .github/wix-app-typecheck (dist/docs/index.json): `usePatternsNavigate`'s `importPath` is indeed `@wix/patterns/router`, not the package root. This claim checks out.

skills/wix-app/SKILL.md
```diff
@@ -440,6 +476,28 @@ Open every path returned in `newFiles` and replace stubbed handler bodies / UI /
...
+- [ ] **Every row opens a page**: a `navigateToEntityPage` call literally appears in `onRowClick` — unless the prompt is explicitly a report or export-only view. A `SidePanel` in a collection page file is the defect this replaces; a display-only page routes to a read-only detail page, it does not fall back to a panel. `grep -n "<SidePanel" <page files>` should return nothing — match the JSX tag, not the bare word, or the templates' own "never a SidePanel" comments fail the check and invite someone to "fix" correct code.
```
mirivoWix: Verified against the installed @wix/design-system types: `SidePanelProps` has no `title` prop, while `SidePanel.Header`'s props do include `title?`. The "never a SidePanel" / EntityPage-for-drill-in guidance elsewhere in this PR is consistent with that. Also ran the repo's own .github/wix-app-typecheck harness against this PR's content end to end — exits 0, no semantic type errors.

PR #1251, opened 2026-09-06
fix(wix-app): stop teaching a CLI that no longer exists

kobiburnley:
Audit **S6** from `DOCS-AUDIT-c646996e`. Split out of #1250, which carried an unrelated audit item.

## The problem

`COLLECTION_TOOLKIT.md` still taught `patterns.cjs`'s subcommands in two paragraphs — `docs`, `types` and `exports` as commands, how to batch arguments into one call, and `exports` with no argument listing the entry points.

That script was deleted; lookups are direct file reads now. So the skill was telling an agent to run commands that don't exist.

## How it survived

Worth naming, because the same gap could hide others: the checkpoint that removed `patterns.cjs` grepped for **`patterns.cjs`**, found nothing, and called the repo clean. It never looked for the *subcommands*. Prose describing a removed tool doesn't mention the tool's filename.

## The change

Both paragraphs keep their advice and lose the tool.

- *"Ask for several names at once"* → **"read the index once, then go straight to files."** The batching advice was about turn cost in a CLI; the equivalent now is that `dist/dts-bundle/index.json` answers every name in that table in a single read, and each entry names the exact file to open next.

- The second paragraph was still true and worth more than its wrapper: a factory or hook's doc shows an empty `## API` because the props table is generated for components, so its signature lives in the bundle. Now stated as **"for these, read the bundle the index names, not the doc"** — which is also what would have kept the audit's run from guessing a hook signature.

Hard counts go with the tool. The entry-point number was correct at 1.460.0, but it is exactly the kind of fact that rots silently between releases.

1 file. No dependency on an unreleased `@wix/patterns`.

### DIFF
--- a/skills/wix-app/references/dashboard-page/COLLECTION_TOOLKIT.md
+++ b/skills/wix-app/references/dashboard-page/COLLECTION_TOOLKIT.md
@@ -34,9 +34,9 @@ Everything below is a real name in the installed `@wix/patterns`. Confirm the pr
 | A fixed in-memory option list for a filter | `useStaticListFilterCollection` |
 | Sorting | `Sortable Columns`, `MultiLevelSorting` |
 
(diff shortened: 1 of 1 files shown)


kobiburnley:
#skipreview

shaharezr (approved):

PR #1279, opened 2026-09-07
docs(wix-app): read @wix/patterns' own guides instead of restating them

kobiburnley:
> [!WARNING]
> **Draft until [cairo#5869](https://github.com/wix-private/cairo/pull/5869) is released.** This PR deletes the skill's copy of "which `@wix/patterns` component serves this need" because the library now ships that answer itself. Until a release carries it, every install fails the new floor check — the skill would correctly refuse to build a dashboard page. Nothing needs editing at merge time: the check is a key probe, not a version comparison.

## Why

Component selection was the only step in this chain with nothing tying it to the package it describes. The component docs and the type bundles are generated from the real installed `@wix/patterns` and validated against it. "Which component serves this need" was hand-written prose in this repo — and it drifted: this skill recommended `PrimaryPageButton` while cairo had already deprecated it, and that one fact took two commits in two repos to fix (#1273 and cairo#5863).

cairo#5869 moves that answer into the library as four guides whose component names **fail cairo's build** if one stops resolving. This side stops carrying a second copy.

| Was here | Is now |
| --- | --- |
| `dashboard-page/COLLECTION_TOOLKIT.md` | `dist/docs/Collection Toolkit.md` |
| `WIX_PATTERNS_DOCS.md` § Library Architecture | `dist/docs/Composition and Providers.md` |
| `WIX_PATTERNS_DOCS.md` § Collection → Entity Flow | `dist/docs/Collection to Entity Flow.md` |
| `dashboard-page/PATTERNS_BUNDLE_READING.md` | `dist/docs/Reading the Doc Indices.md` |
| `dashboard-page/ENTITY_PAGE_TOOLKIT.md` | `dist/docs/useEntityPage.md` |

**431 lines deleted, 78 added.**

## The entry point flips, which is a behaviour change

The chain used to open `dist/dts-bundle/index.json` and call it "the inventory". The guides are entries in **`dist/docs/index.json`** — which is also the larger of the two, 172 entries against 101 — so an agent following the old order would never have seen a guide. `WIX_PATTERNS_DOCS.md` is now one five-step procedure: index → composition → need → doc → example/types.

## What stayed behind, deliberately

The test applied to every candidate was *would a developer reading the library's doc, with autocomplete and the type in front of them, need this sentence?* Instruction about how an agent should behave does not become library documentation by changing repository. So this skill keeps: deciding which **library** owns a UI element, the rules for the reads themselves (exact `file` values, whole files, no `node_modules` spelunking, stop-and-say-so), and the fallback when patterns genuinely has no equivalent.

## Commits

| | |
| --- | --- |
| `2792b758` | state the patterns-before-WDS order once — it was in two files, two wordings, one authoritative |
| `776f04ed` | the move, and the entry-point flip |
| `598e4868` | eight inbound links, five files |
| `b24a20f2` | **the floor check that makes the deletions safe** |
| `d0d1b282` | evals |
| `3ad9f499` | the plan doc, re-verified |

`b24a20f2` is in this PR rather than a follow-up on purpose: between deleting the skill's copy and adding the probe, an agent on an older install has neither source.

## Evals

`employee-shift-dashboard.yml` asserted `ENTITY_PAGE_TOOLKIT.md` was read, so it had to change in this PR, not after it. Two tags naming deleted files went too.

That leaves a structural gap worth flagging in review: **`skill_was_called` only observes files under `skills/wix-app/`**, so now that the component tables ship inside the package, no coverage assertion can prove an agent read them. Two checks moved into the patterns `llm_judge` instead — that the names came from the guides rather than recall, and that a pre-guide install is reported rather than worked around. Correct patterns code written without opening a guide now caps at 6, since it means the chain was skipped and happened to work.

**Not included:** a scenario pinned to a pre-guide `@wix/patterns`, which is the only way to test the upgrade path for real rather than by rubric. It needs an EvalForge template with the version held below the floor. Shipping a scenario that claimed to test it while running against a current install would be worse than leaving the gap visible.

## Checklist

- Content is in an existing skill; no new top-level skill.
- `SKILL.md` index updated — two rows removed, none dangling; every `referenceFiles` path across all scenarios resolves.
- `wix-app` content change is covered by `employee-shift-dashboard.yml` (`skill_was_called` + three `llm_judge`s).
- Scenarios test behaviour, not skill text.


kobiburnley:
Re-synced onto `main` at `2d9d319e`. Three conflicts, all expected — two of them are commits that landed on files this PR deletes, so they needed deciding rather than resolving:

- **skills#1280** corrected `useOptimisticActions` in `COLLECTION_TOOLKIT.md` — it takes the collection state, not the state object the hook returns. **Carried into the cairo guide**, phrased in words rather than as a call, since that mention had gone stale in `DragAndDropOverview.md` precisely because it was written as one.
- **skills#1281** added read-batching guidance to `PATTERNS_BUNDLE_READING.md`. **Kept on this side**, in `WIX_PATTERNS_DOCS.md`'s new *Batch the reads* section — it is instruction about how an agent should read, not about what the format means, so the same test that decided the rest of the split puts it here.
- `WIX_PATTERNS_DOCS.md` conflicted on the section this PR replaces; resolved toward the Discovery Chain.

**The advisory floor moved to 1.468.0.** 1.466.0 and 1.467.0 were both cut while cairo#5869 was still open, so the number was already stale — which is the argument for the key probe being authoritative, now demonstrated rather than hypothesised. Nothing needs editing at merge time either way.

Still a draft, still blocked on cairo#5869 being released. That PR picked up three commits from review since — two real defects, including a wrong provider import mapping whose underlying claim came from `WIX_PATTERNS_DOCS.md` in this repo: the default provider has never auto-detected its environment. Worth knowing independently of this PR, since that line has been there a long time.

kobiburnley:
#skipreview

PR #1284, opened 2026-09-07
fix(wix-app): generate HTTP endpoints for the project runtime

omrimn:
Update the skill to generate `HTTP_ENDPOINT`, preserve each project runtime's directory and handler type, and assume the default `apiDir: "endpoints"` without editing `app()`.


mirivoWix:
### Two follow-ups before this lands

**1. Bump the CLI floor in `skills/wix-app/SKILL.md`**

The frontmatter still says:

```yaml
compatibility: requires `@wix/cli` >= 1.1.192.
```

but the happy path this PR introduces needs 1.1.243. I confirmed the cutoff by diffing the published tarballs — `HTTP_ENDPOINT` appears in `@wix/cli` 1.1.243 and in no earlier version:

| version | files in `build/` mentioning `HTTP_ENDPOINT` |
| --- | --- |
| 1.1.241 | 0 |
| 1.1.242 | 0 |
| 1.1.243 | 6 |

And on 1.1.239 the generator fails with `Unknown extension type: 'HTTP_ENDPOINT'`. So a project anywhere in 1.1.192–1.1.242 satisfies the declared compatibility and still silently drops into the hand-authoring branch. Please raise the line to `>= 1.1.243`.

One knock-on to decide either way: once the floor is 1.1.243, the "CLI older than 1.1.243" fallback in `SKILL.md` (Step 4a) and `BACKEND_API.md` (§ Generate for the Project Type) describes a configuration the skill no longer claims to support. Keeping it as a safety net is fine — just worth being deliberate rather than leaving the two statements in tension.

**2. This needs a major release, not a patch**

Raising the required `@wix/cli` version is a breaking change for consumers of `@wix/agent-skills` (currently `1.18.2`), so after merge someone should run [Release - Bump Version and Open PR](https://github.com/wix/skills/actions/workflows/release-bump.yml) with `version_strategy: major` → `2.0.0`. Flagging it here so it doesn't get picked up by the default patch flow.

mirivoWix:
### The new "verify the route responds" step has no reachable path

`SKILL.md:433` — *"verify the route responds; a passing build alone does not prove endpoint discovery"* — and `BACKEND_API.md:236` — *"request its expected URL on the running dev/preview server"*.

The instinct is right, and it's the most valuable thing this PR adds to the validation section. But as written there's no way to carry it out, so in practice it converts into a reported gap rather than a check. I ran the skill end-to-end with three independent agents (standalone × 2, Astro × 1); all three reached this step, all three could not complete it, and all three correctly declined to fake it.

Two separate problems:

**`wix preview` is not a server.** It has no options at all —

```
$ npx wix preview --help
Usage: wix preview [options]
Create a preview version of your application
Options:
  -h, --help  display help for command
```

— it uploads a version, prints `https://www.wix.com/app-installer?appId=…&versionOverrideId=…` links for Site/Editor/Dashboard, and exits. There is no listening process and no fetchable app-server origin in the output, so "dev/preview server" is only half true: `wix dev` is a server, `wix preview` never is.

**`wix dev` needs a dev site.** Without one it fails with `DevelopmentSiteNotConfigured` ("Please select a development site first using the 'wix dev-site' command"). That's recoverable via `wix dev-site select|create`, but it requires an authenticated account — so in CI, sandboxes, and most agent runs it isn't available, and the skill never mentions the prerequisite.

**Suggested fix — give step 3 a fallback instead of only an escape hatch.** Step 3 already says *"Report any runtime check that could not run"*, which is good discipline, but there's a real check available that closes most of the gap: assert the route in the build output. That's how I confirmed discovery in both runtimes:

- standalone — `grep 'pattern:' dist/server/index.mjs` → `pattern: "/hello"`
- Astro — `grep -r '/api/hello' dist/_worker.js/` → hits in `manifest_*.mjs`

This is strictly stronger than "the build passed": the route table is generated by the filesystem scan, so its presence is direct evidence of discovery. For standalone you can go one step further and import `dist/server/index.mjs` and call `worker.fetch(new Request('https://x/hello'))` — unknown paths return 404 while real ones proceed into the handler, which exercises the matcher itself.

One gotcha worth documenting if you adopt the grep: dynamic segments are emitted in path-to-regexp v8 quoted form, so `src/endpoints/users/[id].ts` appears as `pattern: "/users/:\"id\""`. That's correct output, not corruption — I confirmed `/users/123` matches and `/users/` and `/nope` 404.

Also minor, same section: step 3's *"Check URLs with a dev prefix and without one for release"* was the one line all three agents flagged as unclear about what they were supposed to actually run.

omrimn:
> ### Two follow-ups before this lands
> **1. Bump the CLI floor in `skills/wix-app/SKILL.md`**
> 
> The frontmatter still says:
> 
> ```yaml
> compatibility: requires `@wix/cli` >= 1.1.192.
> ```
> 
> but the happy path this PR introduces needs 1.1.243. I confirmed the cutoff by diffing the published tarballs — `HTTP_ENDPOINT` appears in `@wix/cli` 1.1.243 and in no earlier version:
> 
> version	files in `build/` mentioning `HTTP_ENDPOINT`
> 1.1.241	0
> 1.1.242	0
> 1.1.243	6
> And on 1.1.239 the generator fails with `Unknown extension type: 'HTTP_ENDPOINT'`. So a project anywhere in 1.1.192–1.1.242 satisfies the declared compatibility and still silently drops into the hand-authoring branch. Please raise the line to `>= 1.1.243`.
> 
> One knock-on to decide either way: once the floor is 1.1.243, the "CLI older than 1.1.243" fallback in `SKILL.md` (Step 4a) and `BACKEND_API.md` (§ Generate for the Project Type) describes a configuration the skill no longer claims to support. Keeping it as a safety net is fine — just worth being deliberate rather than leaving the two statements in tension.
> 
> **2. This needs a major release, not a patch**
> 
> Raising the required `@wix/cli` version is a breaking change for consumers of `@wix/agent-skills` (currently `1.18.2`), so after merge someone should run [Release - Bump Version and Open PR](https://github.com/wix/skills/actions/workflows/release-bump.yml) with `version_strategy: major` → `2.0.0`. Flagging it here so it doesn't get picked up by the default patch flow.

I keept the floor at 1.1.192 deliberately. The HTTP endpoint extension itself is supported on every CLI from 1.1.192; what 1.1.243 adds is the `wix generate` shortcut. The skill says so explicitly: Step 4a and the Generate section in BACKEND_API.md tell the agent that an older CLI predates the generator and to either update or create the endpoint file by hand in the runtime's directory.

mirivoWix:
Agreed — you're right and I'll withdraw the floor request. `compatibility` is the minimum for the skill to be usable, not for its fastest path, and the skill degrades with a documented fallback, so raising it to 1.1.243 would exclude setups that genuinely work. "Silently drops into the hand-authoring branch" was wrong of me: it's stated in both Step 4a and the Generate section, which is exactly what you pointed at. I should have exercised the fallback before arguing from the frontmatter.

Having now exercised it, there is one gap in it worth closing — in the standalone branch specifically.

**Hand-authoring a standalone endpoint also needs `@wix/custom-extensions >= 0.2.14`, and nothing on that path says so.** On 0.2.13 with a plain `app()`, a hand-created `src/endpoints/hello.ts` is never discovered, and the build still passes:

```
$ npx wix build
build-exit=0
$ ls dist/server/
(no server output)
```

Cause, from 0.2.13's `build/integration/node.mjs`:

```js
endpointRoutes: apiDir !== void 0 ? (await discoverEndpointRoutes(...)) : []
```

`apiDir` has no default until 0.2.14 — `constructor({ apiDir } = {})` in 0.2.13 vs `constructor({ apiDir = DEFAULT_API_DIR } = {})` in 0.2.14 — so nothing under `src/endpoints` gets scanned. Second failure on the same project: the import the fallback tells you to write doesn't resolve either, since `./types` also first ships in 0.2.14:

```
src/endpoints/hello.ts(1,31): error TS2307: Cannot find module '@wix/custom-extensions/types'
```

The doc already knows this on the other branch — *"The generator writes there and upgrades `@wix/custom-extensions` to `^0.2.14` if needed (the first release with that default and the `./types` export)"*. The hand-authoring paragraph just doesn't carry the same clause, and it's the branch where nothing bumps it for you. Since an un-updated CLI and an un-updated `@wix/custom-extensions` tend to travel together, this is likely to be the common case for exactly the projects the fallback targets.

Suggested: one clause on the fallback — in a standalone project, also ensure `@wix/custom-extensions >= 0.2.14` (below that there is no default `apiDir` and no `./types` export), then install before validating.

Astro is unaffected: `src/pages/api/*` is Astro's own routing, and I confirmed it works on 1.1.239.

---

**Correction to point 2 of my earlier comment.** I justified the major bump with "raising the required `@wix/cli` version is a breaking change." Since the floor is staying at 1.1.192, that reasoning no longer applies and I'm withdrawing it. Whether this still warrants more than a patch is a separate call I'll leave to you — I'm not claiming a compatibility break.

omrimn:
Re: https://github.com/wix/skills/pull/1284#issuecomment-5617263289

You were right: the old step told the agent to call the endpoint on a running server, and there usually isn't one. `wix preview` uploads a version and exits. `wix dev` only starts if a dev site was selected, which agents and CI rarely have.

I also asked myself why we check the endpoint at all. The only problem the check can find is a file saved in the wrong folder, so the runtime never picks it up. The build doesn't complain about that, the route just 404s. When the CLI generated the file, that almost never happens.

So in 22dadc20 I made the check optional instead of a required step:

- The validate steps are now: run `wix build`; if a dev site is configured, run `wix dev` and call the URL; then deploy. The text says plainly that `wix preview` is not a server.
- Below that, a short troubleshooting hint: if the route 404s or you created the file by hand, look for it in the build output. Standalone: `grep 'pattern:' dist/server/index.mjs` lists every route (a dynamic segment shows as `:"id"`, that's normal). Astro: search for the route string under `dist/`. If it's not there, the file is in a folder the runtime doesn't scan.
- The confusing "dev prefix" sentence is removed.

omrimn:
Re: https://github.com/wix/skills/pull/1284#issuecomment-5617516817

Thanks for exercising the fallback, that gap was real. Fixed in 066f6098.

Rather than add the clause only to the hand-authoring paragraph, the standalone paragraph now states the rule once, since it applies no matter who creates the file: standalone endpoints require `@wix/custom-extensions@^0.2.14`, because older releases neither scan `src/endpoints` by default nor export `./types`, and the build passes without serving anything. The generator upgrades the package; when creating the file by hand, you upgrade it yourself, then install before validating.

Agreed on the release: the CLI floor stays at 1.1.192 and nothing breaking changed, so the default patch flow is fine.

skills/wix-app/SKILL.md
```diff
@@ -8,7 +8,9 @@ compatibility: requires `@wix/cli` >= 1.1.192.
 
 Helps build extensions for Wix CLI applications. Covers all extension types: dashboard pages, modals, plugins, menu plugins, custom element widgets, Editor React components, site plugins, embedded scripts, backend APIs, events, service plugins, and data collections.
 
-**Scaffolding is owned by the Wix CLI.** For every extension type except Backend API, files, folders, builder boilerplate, UUIDs, and `src/extensions.ts` registration are generated by `wix generate --params`. This skill provides the decision logic, API guidance, configuration semantics, and business-logic patterns that fill in the generated stubs.
+**Scope:** These are app-extension workflows. Wix CLI headless sites retain their framework routing and authentication; use the [wix-headless skill](../wix-headless/SKILL.md). Using the Wix CLI alone does not make a project standalone or Studio 2.
```
mirivoWix: Please delete it, don't mention headless/studio2 in this skill :)

omrimn: removed

mirivoWix (approved):

PR #1291, opened 2026-09-08
feat(evalforge-yaml-gate): review skill and eval scenario content on PRs

omermesser:
# An AI reviewer for `wix-manage` content quality

A new `review` mode in `evalforge-yaml-gate` runs Claude Code over the changed skills and eval scenarios with the whole repo checked out, and posts findings as one PR comment.

## Triggers

| Event | Agent runs | Check |
|---|---|---|
| PR opened / reopened / ready for review | ✅ | ❌ on a blocking or malformed finding |
| push (`synchronize`) | ❌ | ❌ *"commit abc1234 has not been reviewed"* |
| `/review` | ✅ | ❌ on a blocking or malformed finding |
| nothing in scope, or non-Wix author | ❌ | ✅ |

Automatic on creation, manual after — a push doesn't spend, but it doesn't pass either, so nothing merges unreviewed.

Four things fail the check: an unreviewed head, a blocking finding, a malformed finding (the review is still published, with ⚠️ Review job partly completed), and a review that didn't complete — a timeout, a missing CLI, a bad envelope. A commit nobody could review is not a reviewed commit. `/re-eval` is the remedy for a flaky run, which is what it was built for in the eval gate.

The only green skips are the two where a review was never meant to happen.

## Rollout

`SKILL_REVIEW_BLOCK_MERGE` controls whether findings block; it's currently `false` to check the first couple of runs.

## Also in here

- `CONTRIBUTING.md` / `docs/eval-scenarios.md`: the rules the reviewer cites, written down first — the checklist's three groups now map to what's gated, what the review reads, and what it can't.
- `evalforge-re-eval.yml`: one entry in `GATES`.

PR #1305, opened 2026-09-09
test: exercise the skill review gate (do not merge)

omermesser:
**Test PR — do not merge.** Exercising the reviewer after the diff-range and sanitize changes on `skill-review-gate` (#1291).

What this round is actually testing:

- the reviewer now runs `git diff HEAD^1 HEAD` against GitHub's merge commit instead of `$BASE_SHA HEAD`, so the range should be this PR's own changes only — previously a re-run after `main` moved pulled in other people's merged files
- comment markers and the api key are neutralised once at the trust boundary, rather than at each render site

The planted issue:

- a named MCP tool (`CallWixSiteAPI`), which [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md#stay-agnostic-to-agent-and-client) forbids — expected `blocking`

Only `skills/wix-manage/references/domains/domains-dashboard-navigation.md` is in review scope. Everything else in the diff comes from being stacked on `skill-review-gate`. `SKILL_REVIEW_BLOCK_MERGE` is unset, so the check warns rather than fails.

Closing this once the review has run.

### DIFF
--- a/.github/actions/evalforge-skill-gate/tests/re-eval-script.test.ts
+++ b/.github/actions/evalforge-skill-gate/tests/re-eval-script.test.ts
@@ -60,6 +60,13 @@ const MANAGE_RUN: WorkflowRun = {
   html_url: 'https://github.com/wix/skills/actions/runs/901',
 };
 
+const REVIEW_RUN: WorkflowRun = {
+  id: 902,
+  status: 'completed',
+  conclusion: 'failure',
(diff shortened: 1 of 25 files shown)


omermesser:
Review ran as expected — 1 blocking finding on the one in-scope file, correctly narrowed from the stacked diff. Closing.

PR #1308, opened 2026-09-09
fix(google-ads): correct the recipe base URL that returns 404

gilisho:
## Summary

The Google Ads recipes tell callers to use `https://www.wixapis.com/google-ads/v1/...`. **That base returns 404 for every endpoint.** The real base is `https://www.wixapis.com/_serverless/pa-google/v1/...`, which is what each method's page in the [Google Ads API reference](https://dev.wix.com/docs/api-reference/business-management/marketing/ads/google-ads/introduction) shows.

This corrects the base in 5 of the 7 Google Ads recipes and in the `wix-manage/SKILL.md` index entries that mirror their frontmatter descriptions. The remaining 2 recipes are deferred to a follow-up (see below). Two eval-side changes ride along so the gate can actually catch this class of bug.

## Validation, endpoint by endpoint

Each path was probed against the live gateway before and after. A bogus control path on the same service (`/_serverless/pa-google/v1/bogus-xyz`) returns **404**, so a **400/403** response proves the route exists and only the request body or auth is missing.

| endpoint (after the fix) | before | after | reference |
|---|---|---|---|
| `GET /v1/accounts` | 404 | 403 | route exists; reference documents `POST` here ([Create Account](https://dev.wix.com/docs/api-reference/business-management/marketing/ads/google-ads/google-account-v1/create-account)) |
| `POST /v1/accounts` | 404 | 400 | [Create Account](https://dev.wix.com/docs/api-reference/business-management/marketing/ads/google-ads/google-account-v1/create-account) |
| `GET /v1/accounts/current-site` | 404 | 403 | [Get Account For Current Site](https://dev.wix.com/docs/api-reference/business-management/marketing/ads/google-ads/google-account-v1/get-account-for-current-site) |
| `PATCH /v1/accounts/{accountId}` | 404 | 400 | [Update Account](https://dev.wix.com/docs/api-reference/business-management/marketing/ads/google-ads/google-account-v1/update-account) |
| `DELETE /v1/accounts/{accountId}` | 404 | 403 | [Delete Account](https://dev.wix.com/docs/api-reference/business-management/marketing/ads/google-ads/google-account-v1/delete-account) |
| `GET /v1/accounts/current-site/conversion-actions` | 404 | 500 | **not in the reference** (route exists) |
| `GET /v1/incentives` | 404 | 400 | **not in the reference** (route exists) |
[...]

PR #1310, opened 2026-09-09
fix(evalforge): gate on org membership and the current committer

gilisho:
## Problem

The YAML gate decides who is a Wix author from the **first** commit's author email. Two things go wrong.

**It rejects Wix employees.** Anyone whose git identity is their GitHub noreply address fails, which is the default once GitHub email privacy is on. On #1308, opened by a Wix author from a branch on this repo, the gate failed with:

```
PR author gate failed: the PR's first-commit author email
(40733156+gilisho@users.noreply.github.com) is not a @wix.com address.
```

The skill review reads the same signal and *soft-skips* on it, so it logged `Skipping the skill review — the PR author is not a wix author` and still reported a **green check that had reviewed nothing**. That is the worse half of the bug: a gate that fails loudly is annoying, a review that passes without reviewing is misleading.

**The first commit is the wrong commit, and the weakest evidence.** A commit author email is whatever the author typed into `git config`, so it is self-asserted. And the opening commit can be cherry-picked from someone else while the actual work is pushed on top. What the eval spends on is the head commit.

## Change

Judge organization membership first, from `author_association` on the pull_request payload:

```ts
const ORG_ASSOCIATIONS = new Set(['OWNER', 'MEMBER', 'COLLABORATOR']);
```

GitHub computes that field server-side from the author's identity, so unlike a commit email it cannot be set by whoever wrote the commit. It also costs no API call and needs no extra token scope — reading membership directly with `orgs.checkMembershipForUser` would require `members: read`, which the default `GITHUB_TOKEN` does not carry, and would 404 for anyone whose org membership is private.

The email check remains as a fallback so the gate still clears authors on payloads that carry no `author_association` (a replayed or manually dispatched run), rather than turning a missing field into a refusal. That fallback now reads the **latest** commit instead of the first.

`getHeadCommitAuthorEmail` pages forward to the last page: the PR commits API returns commits oldest-first with no option to ask for the newest.

Both consumers in this action are updated — `gate.ts`, which throws, and `review.ts`, which soft-skips and now short-circuits on membership before making any commit lookup at all.

## Tests

`packages/evalforge-core/tests/author-gate.test.ts` goes from 7 cases to 15, covering the new predicate (`MEMBER`/`OWNER`/`COLLABORATOR` accepted; `CONTRIBUTOR`, `FIRST_TIME_CONTRIBUTOR`, `NONE`, empty and missing rejected), the org short-circuit making no API call, both fallback paths, and the refusal when neither signal clears the author. The existing spoofing cases for `isWixAuthorEmail` are untouched.

Verified locally: core suite 411 passing (403 on `main`, so +8 with no regressions), action typecheck clean, `dist/index.js` rebuilt with ncc and committed. Two core test files and several action test files fail identically on `main` in this checkout, so they are pre-existing local environment failures rather than anything this change introduces.

## Deliberately not included

`evalforge-skill-gate` makes the same first-commit check in `pr-lookups.ts` and `sync-run.ts`. I had it converted, then reverted it: its bundled `dist/` is what the action actually runs and it cannot be rebuilt from this machine (its `yarn install` fails against the registry), so shipping source that does not match the running artifact would be worse than leaving it internally consistent. It needs the same treatment from someone who can build it.

Also worth a decision separately: `skill-eval.yml` carries a literal `if: false`, so that workflow never runs on any PR.

### DIFF
--- a/.github/actions/evalforge-skill-gate/dist/index.js
+++ b/.github/actions/evalforge-skill-gate/dist/index.js
@@ -30090,30 +30090,105 @@ exports.TokenProvider = TokenProvider;
 
 Object.defineProperty(exports, "__esModule", ({ value: true }));
 exports.isWixAuthorEmail = isWixAuthorEmail;
+exports.isWixOrgAuthor = isWixOrgAuthor;
 exports.getFirstCommitAuthorEmail = getFirstCommitAuthorEmail;
+exports.getHeadCommitAuthorEmail = getHeadCommitAuthorEmail;
+exports.resolveWixAuthor = resolveWixAuthor;
 exports.assertWixAuthor = assertWixAuthor;
 const WIX_EMAIL_RE = /@wix\.com$/i;
+/**
+ * `author_association` values GitHub reports for someone who belongs to the repo's
+ * organization, or who has been granted access to the repo directly.
+ *
+ * This is the primary signal: GitHub computes it server-side from the PR author's
+ * identity, so unlike a commit author email it cannot be set by whoever wrote the
+ * commit. It also costs no API call and needs no extra token scope — reading org
+ * membership directly (`orgs.checkMembershipForUser`) would need `members: read`,
+ * which the default `GITHUB_TOKEN` does not carry, and would 404 for anyone whose
+ * membership is private.
+ */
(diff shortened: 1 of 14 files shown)


gilisho:
Superseded by #1311 — same change, re-authored under my GitHub noreply address. Closing to avoid a duplicate.

PR #1311, opened 2026-09-09
fix(evalforge): gate on org membership and the current committer

gilisho:
## TL;DR

The PR author gate asked "is the commit author email `@wix.com`?". That question is both too strict and not strict at all.

- **It locks out Wix employees.** If your git identity is your GitHub noreply address — the default once email privacy is on — you fail the gate. That is what happened on #1308.
- **It lets anyone else in.** A commit author email is free text that git copies from `user.email`. `git config user.email someone@wix.com` clears the gate, with no Wix account involved.
- **And one check failed silently.** The skill review skipped on the same signal but still reported a green check, so a PR could look reviewed when nothing had reviewed it.

This PR asks GitHub instead: is the author a member of the `wix` org? That comes from `author_association`, which GitHub computes from the account that opened the PR, so the author cannot set it. The commit email check is deleted, not demoted — a fallback would have kept the bypass.

It covers five call sites across three actions, including `skill-eval`, which turned out to have no author check at all rather than a weak one. And the gate never reddens a check over its own uncertainty: if the author cannot be resolved, the modes that skip say so on the PR instead of failing, which is the guarantee the soak period depends on.

**Reviewer's eye here:** `ORG_ASSOCIATIONS` in `packages/evalforge-core/src/author-gate.ts` is the entire trust decision. It accepts `OWNER` and `MEMBER`. `COLLABORATOR` is excluded on purpose, because repo push access is not org membership.

Nothing about who *can* merge changes — this only governs which PRs spend an eval run and get reviewed.

## Detail

The gate's failure on #1308, in full:

```
PR author gate failed: the PR's first-commit author email
(40733156+gilisho@users.noreply.github.com) is not a @wix.com address.
```

The skill review reads the same signal but *soft-skips* rather than throwing, so it logged `Skipping the skill review — the PR author is not a wix author` and reported success. A gate that fails loudly is annoying; a review that passes without reviewing is misleading.

Signing would not have helped, for two independent reasons. The gate never read `commit.verification` — only `commit.author.email` — so a signature could not have changed its answer. And signing is not required: `required_signatures` is disabled on `main`, so an unsigned commit is accepted regardless.

Reading the *first* commit compounds it. That commit can be cherry-picked from someone else while the real work is pushed on top, so the identity it carries need not belong to whoever wrote what is being evaluated.

## Change

Membership needs no extra token scope, which is why the association is read rather than queried directly: `orgs.checkMembershipForUser` would require `members: read`, which the default `GITHUB_TOKEN` does not carry, and would 404 for anyone whose membership is private.

`isWixAuthorEmail`, `getFirstCommitAuthorEmail` and `getHeadCommitAuthorEmail` are all deleted, along with the commit paging they needed.

The email path was unreachable as well as unsound. Every author-gated mode runs on a `pull_request` event, and `/re-eval` POSTs `actions/runs/{id}/rerun`, which replays the original payload — so the association is always there:

| Action | Site | Workflow | On refusal |
| --- | --- | --- | --- |
| `evalforge-yaml-gate` | `gate.ts` | `evalforge-yaml-gate.yml` | throws |
| `evalforge-yaml-gate` | `review.ts` | `evalforge-skill-review.yml` | soft-skips |
| `evalforge-skill-gate` | `pr-lookups.ts` | `evalforge-wix-app-gate.yml` | soft-skips |
| `evalforge-skill-gate` | `sync-run.ts` | `evalforge-wix-app-sync.yml` | soft-skips |
| `skill-eval` | `eval.ts` | `skill-eval.yml` | throws |

Because it is always there, the gate takes it as a **required** config field and needs no client at all. `assertWixAuthor` is synchronous and receives the association; deciding costs no API call, no token scope, and nothing that can fail in transit. An earlier revision of this branch made the field optional with an API lookup behind it, which described a payload no config can produce: every config populating the field already calls `getPrNumber`, which rejects a payload with no `pull_request`. That fallback is gone, and with it `resolveWixAuthor`, `getPullAuthorAssociation` and `PullAuthorClient`.

`requireAuthorAssociation` throws when the field is absent or not a string, the way `getPrNumber` throws on a missing number. A gate that cannot identify the author should fail loudly rather than quietly pick a branch. It takes the payload as an argument rather than reading `github.context` itself, so the shared package still needs no dependency on `@actions/github`, and it checks the type at runtime rather than asserting a shape onto `WebhookPayload`, whose `pull_request` is an open bag of `any`. It also replaces a `getAuthorAssociation` that had been copied into two actions' config modules.

Two things fell out of that, both cases that no longer exist rather than behaviour dropped on purpose. `checkPrAuthor` is synchronous and client-free, so `AuthorCheck` loses its `isUnexpected` flag and `runGate` no longer branches its log level on one — there is no lookup left to blip. And `review.ts` loses its try/catch, `sync-run.ts` its now-unused Octokit.

## Tests

`packages/evalforge-core/tests/author-gate.test.ts` covers the predicate (including `COLLABORATOR` being refused), that `requireAuthorAssociation` throws on every unusable payload, and that nothing the author controls can clear the gate — the spoof, as a regression test. Each action's tests drive the same paths through their own config.

All four suites pass, and all three bundles are rebuilt with ncc and committed:

| Package | Tests |
| --- | --- |
| `packages/evalforge-core` | 422 passing |
| `.github/actions/evalforge-yaml-gate` | 209 passing |
| `.github/actions/evalforge-skill-gate` | 222 passing |
| `.github/actions/skill-eval` | 99 passing |

**A note on running these locally:** the suites need Node 20+. On Node 18 most of them fail during module mocking with `(0 , U.tracingChannel) is not a function`, because a transitive `lru-cache` dependency uses `diagnostics_channel.tracingChannel`. That failure is environmental and looks alarming — it presents as ~100 failures with no obvious connection to the change. CI runs Node 24 and is unaffected.

## `skill-eval` had no gate at all

Worth calling out because it is not the same bug. `skill-eval` did not check the author email — it checked nothing. `runEval` went from config straight to spending, so it was the one entry point that would start an eval run for any PR author, including an outside fork.

Nothing is exposed today: `skill-eval.yml` carries a literal `if: false`, and `skill-eval-cleanup.yml` is off for the same reason, its comment recording that the action is superseded by the evalforge gates. The gate is added for whoever re-enables it, so that re-enabling does not also re-open the hole the other gates just closed. It uses the shared `assertWixAuthor` rather than a fourth copy of the check, which means the action now takes the same `portal:` dependency on the core package that the two evalforge actions already use.

Taking that dependency needed its CI wired up to match, since core's `main`/`types` point at a `dist/` that is not committed. `skill-eval-ci.yml` now builds core first, the way `ci.yml` already did for the two gate actions, and its path filter grew `packages/evalforge-core/**` — the action's bundle inlines core, so a core-only change makes it stale, and `ci.yml`'s staleness check covers only the two gates. Without that path the bundle would have drifted with nothing reporting it.

Deleting the action instead was considered and **rejected**: it stays, gated. So the wiring above is the long-term arrangement rather than a stopgap, and the path filter matters — it is what keeps this bundle honest whenever core changes.


.github/actions/evalforge-skill-gate/src/utils/pr-lookups.ts
```diff
@@ -20,10 +20,12 @@ const AUTHOR_ALLOWED: AuthorCheck = { allowed: true };
...
-  config: Pick<GateConfig, 'owner' | 'repo' | 'prNumber'>,
+  config: Pick<GateConfig, 'owner' | 'repo' | 'prNumber'> & { authorAssociation?: string },
 ): Promise<AuthorCheck> {
   try {
-    const email = await getFirstCommitAuthorEmail(octokit, config.owner, config.repo, config.prNumber);
+    // Org membership settles it with no API call; the current committer is the fallback.
+    if (isWixOrgAuthor(config.authorAssociation)) return AUTHOR_ALLOWED;
+    const email = await getHeadCommitAuthorEmail(octokit, config.owner, config.repo, config.prNumber);
     if (isWixAuthorEmail(email)) return AUTHOR_ALLOWED;
     return { allowed: false, reason: 'the PR author is not a wix author', isUnexpected: false };
```
gilisho: This is not a good check. User can have a Wix email without having it signed (`unverified commit`).

.github/actions/evalforge-skill-gate/src/utils/sync-run.ts
```diff
@@ -44,9 +44,11 @@ export async function runSync(): Promise<void> {
 
   const octokit = github.getOctokit(config.githubToken);
   const [owner, repoName] = config.repo.split('/', 2);
-  const authorEmail = await getFirstCommitAuthorEmail(octokit, owner, repoName, config.prNumber);
-  if (!isWixAuthorEmail(authorEmail)) {
-    core.info('Skipping wix-app sync — PR author is not a @wix.com address');
+  const authorized =
+    isWixOrgAuthor(getAuthorAssociation()) ||
+    isWixAuthorEmail(await getHeadCommitAuthorEmail(octokit, owner, repoName, config.prNumber));
```
gilisho: This is not a good check. User can have a Wix email without having it signed (`unverified commit`).

.github/actions/evalforge-yaml-gate/src/utils/config.ts
```diff
@@ -49,9 +51,16 @@ export function getSimpleConfig(): SimpleConfig {
...
+    authorAssociation: getAuthorAssociation(),
   };
 }
 
+/** `author_association` off the pull_request payload, absent on replayed/dispatched runs. */
+function getAuthorAssociation(): string | undefined {
+  const pr = github.context.payload.pull_request as { author_association?: string } | undefined;
+  return pr?.author_association;
+}
+
```
gilisho: I think `getAuthorAssociation` is duplication in an other check.

.github/actions/evalforge-skill-gate/src/utils/pr-lookups.ts
gilisho: Agreed, and it was worse than weak — it was unreachable *and* spoofable, so it narrowed nothing.

Every author-gated mode runs on a `pull_request` event (`evalforge-yaml-gate.yml`, `evalforge-skill-review.yml`, `evalforge-wix-app-gate.yml`, `evalforge-wix-app-sync.yml`), and `/re-eval` POSTs `actions/runs/{id}/rerun`, which replays the original payload. So `author_association` is always present and the email branch never ran — except for anyone who wanted it to, by setting `user.email` to any `@wix.com` address on an unsigned commit.

Removed in 18b718af. `isWixAuthorEmail`, `getFirstCommitAuthorEmail` and `getHeadCommitAuthorEmail` are all gone. Where a payload genuinely carries no association, the gate now asks the API for the same server-computed field rather than trusting an email.

.github/actions/evalforge-skill-gate/src/utils/sync-run.ts
gilisho: Same fix, same commit — `runSync` now calls `resolveWixAuthor` like the other three call sites, so there is one decision path rather than four hand-rolled ones.

.github/actions/evalforge-yaml-gate/src/utils/config.ts
gilisho: Right — it was defined in both actions' `config.ts`. It now lives in the shared package as `readAuthorAssociation(payload)`, and both actions call it.

It takes the payload as an argument rather than reading `github.context` itself, so the package still needs no dependency on `@actions/github` (same reason the octokit slice there is declared structurally). It also checks the field is a string at runtime instead of asserting a shape onto `WebhookPayload`, whose `pull_request` is typed as an open bag of `any`.

PR #1320, opened 2026-09-13
login: make awaiting_user a hard stop

gonengar:
## Why

An agent that receives `awaiting_user` and keeps building never surfaces the device code — the user can't see tool output, so login silently stalls until they ask what's going on. Observed in the wild with Grok, which kept scaffolding right past the login prompt.

## What

Makes `awaiting_user` an explicit **hard stop** everywhere the login flow is documented, and bakes the rule into the event itself:

- **`wix-headless/entry/skill.md`** and **`wix-headless-fast/entry/skill.md`** — the event-table row now says HARD STOP, with a new subsection spelling out the four rules: send a user-visible message in the same turn, no other tool calls, keep the bootstrap running, and on expiry re-run once and surface the new code.
- **`wix-headless/references/managed/AUTHENTICATION.md`** — hard-stop callout before the login steps; step 3 now forbids continuing the recipe or calling other tools in the same turn; adds the ~10-minute code-expiry rule.
- **`wix-auth/references/device-flow.md`** — Step 2 retitled as a hard stop, with the message template and an explicit "don't skip straight to polling" rule.
- **`wix-headless/entry/bootstrap.mjs`** — re-emits `awaiting_user` with an `instruction` field carrying the hard-stop rule, so an agent that skims the JSON output (and never read the skill doc) still sees it next to the code it must relay.

`node --check` passes on the modified script.

---
🤖 This PR was written and opened by Joseph, an automated AI agent (Claude Code), on behalf of Gonen Jerbi.

### DIFF
--- a/skills/wix-auth/references/device-flow.md
+++ b/skills/wix-auth/references/device-flow.md
@@ -34,7 +34,9 @@ Response:
 { "deviceCode": "…", "userCode": "XXXXXXXX", "verificationUri": "https://users.wix.com/login/device-login", "expiresIn": 600 }
 ```
 
-## Step 2 — Ask the user to authorize
+## Step 2 — Ask the user to authorize (hard stop)
+
+**Stop all other work.** The user cannot see tool output, so they never get the code unless you send a user-visible message in the **same turn** you receive `userCode`.
 
 Show the user this URL and code:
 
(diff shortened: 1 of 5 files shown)


ayal:
Closing this one — the concern is real (an agent that keeps building past awaiting_user never surfaces the code), but the fix sprays hard-stop directives across three skills and the bootstrap script. We'll rethink the essence and, if a change is warranted, make it narrowly in wix-headless only.

PR #1336, opened 2026-09-14
headless-fast: own the bootstrap instead of borrowing it

ayal:
`wix-headless-fast` had no bootstrap of its own. Both its quick-start and its Phase 1 curled **wix-headless's** copy:

```
https://www.wix.com/skills/headless/entry/bootstrap.mjs
```

One file, two skills — so a change to it silently changed behaviour for both.

That isn't hypothetical. #1321 made `awaiting_user` terminal (detach the login, exit), which instantly falsified the fast entry's description of the script it runs: it still told agents to *"let it keep running until the login completes"* and to wait for `logged_in`/`success` on a stdout that had already closed. #1325 fixed the prose, but the coupling that produced it stayed.

## This PR

- Adds `skills/wix-headless-fast/entry/bootstrap.mjs`, **byte-identical** to the `wix-headless` copy at this commit.
- Repoints every reference (quick-start, Phase 1 macOS/Linux, Phase 1 Windows) to `https://www.wix.com/skills/headless-fast/entry/bootstrap.mjs`.
- Drops the prose that called it borrowed — no longer "the same script the classic headless entry uses" or "the shared bootstrap script".

The two copies being identical today is the point: either can now change without moving the other.

## Worth deciding before merge

Duplication cuts both ways. It buys independence, and it costs a drift risk that nothing currently guards.

If these are meant to stay in lockstep, a CI check asserting the two files are identical would make drift loud instead of silent. If they're meant to diverge — which is the usual reason to fork a file — then no guard, and the fast copy is free to change.

Happy to add the check, or leave it. More changes are coming on this branch either way.


ayal:
Closing — wrong approach. The fast entry keeps using the shared bootstrap at wix-headless.

PR #1354, opened 2026-09-16
Default productType to PHYSICAL when unstated in create-product-catalog-v3

noakr:
## What

The Catalog V3 create-product recipe says product type is an "optional workflow input" — correctly meaning the agent shouldn't stop and ask the user for it before creating. In practice that phrasing was also read as "optional to omit from the request payload," which it isn't: the Catalog V3 create API has no server-side default for `productType`, so an omitted value fails validation at execute time.

Seen in a real conversation where the execute tool failed because `productType` was missing from the create request.

## Fix

Default to `productType: "PHYSICAL"` (with `physicalProperties: {}` on the product and every variant) whenever the user hasn't explicitly stated a type and it can't be inferred (e.g. from an attached digital file) — never omit `productType` from the request, and never ask the user to pick one first.

Stated in the same four places this doc already reinforces its other hard requirements, so the rule survives regardless of where in the recipe the agent is reading from:
- Preflight (case 3)
- The Type row in "Build each product"
- The "Hard requirement before every create call" warning box
- The Failure guide

## Testing

Docs-only change to a skill reference file; no code to run.


hadarzaguri (dismissed):
Let's add a scenario please

noakr:
Closing in favor of a same-repo PR (this one was opened from a fork, which the wix-manage eval bot can't run against — see CONTRIBUTING.md).

PR #1362, opened 2026-09-16
feat(ci): review skill unification, tighten eval-coverage guidance

omermesser:
Four changes to what the CI skill reviewer reads and reports.

**Skill unification.** The reviewer now checks whether a new skill should exist at all, or whether its content belongs in one we already have. `CONTRIBUTING.md` defines when two skills are genuinely separate — by intent and use case, not by how the request happens to be phrased — and lists the cases that should be unified instead. The prompt tells the agent to read the index section for the area under review and raise the overlaps it finds.

**Eval coverage.** A scenario has to give real confidence that the skill resolves the requests it exists for, and that a regression would be caught.

**`fix-before-merge` → `advisory`.** The old name said when to fix it, not what it was. Advisory means friction rather than a wrong answer, so the merge doesn't wait on it. `blocking` is unchanged.

**Scenario data.** Whatever a scenario's prompt refers to has to exist when the run starts — seeded through `siteSetup` and its `bootstrap` steps, or already prepared on the test account. The previous wording read as if `siteSetup` were the only legal answer, which left no room for a scenario that needs no site, or one pointed at a site we keep ready.

**Also:** `suggestion` is now required, so a finding always arrives with a fix rather than just a complaint.


.github/prompts/skill-review.md
```diff
@@ -1,20 +1,28 @@
...
+`CONTRIBUTING.md#when-a-skill-earns-its-place` is the standard for when content belongs in a skill that already exists rather than in a new one. Apply it on every PR: overlap is invisible in the diff and survives every automated gate, so you are the only reader positioned to catch it.
 
-There is deliberately no list of the guides' sections here: you have both guides in front of you. They are the baseline — check the change against them — but they are not the ceiling. Content that is wrong, contradicts itself, or would send an agent down the wrong path is a finding whether or not any section covers it: report it, and bring the wording you would use instead.
+Sweep first, and from the index rather than the diff. Open the whole section of `skills/wix-manage/SKILL.md` for the area this PR touches and read every entry in it, changed or not. Then open related skills and look for unifications to raise. Raise every unification you can see, however weak the signal: if two skills could be one they probably should be, and the contributor can disagree in a sentence.
```
hadarzaguri, line 17: Q - if two skills could be one they probably should be, and the contributor can disagree in a sentence.

Is it necessary to include this? Since we want real users to actually respond to the agent, adding extra instructions might cause them to skip reading it?

hadarzaguri (dismissed):

.github/prompts/skill-review.md
omermesser: im keeping it 😃

PR #1373, opened 2026-09-17
DO NOT MERGE — test fixture for the skill review comment flow

omermesser:
## DO NOT MERGE

A throwaway PR for watching the skill reviewer's comment behaviour end to end. Close it when the check on #1372 has been observed; nothing here is meant for `main`.

**Why it exists.** #1372 changes how the reviewer manages its PR comments — a new comment per review with the previous one collapsed, the triggering user named on the verdict, and the `/review` acknowledgement cleared once the review reports. For a `pull_request` event GitHub runs the workflow and action from the PR's own tree, so the only way to see the new behaviour before merge is a PR branched off that change. This is that PR.

**What to watch, in order.**

1. **Now** — one verdict comment, ending `triggered by @…` on the job-status line.
2. **Push any commit here** — the verdict stays put, and a reminder appears saying the new commit is unreviewed.
3. **Comment `/review`** — an acknowledgement appears next to both.
4. **When the review reports** — a new verdict is posted, the earlier one collapses to *"This comment was marked as outdated"*, and both the reminder and the acknowledgement are deleted. One visible verdict, one collapsed, nothing else.

**The example skill has ordinary mistakes on purpose**, so the verdict has something to report rather than coming back empty: it covers ground `query-sites.md` already covers using the same endpoint, it names a specific tool instead of the capability it needs, and it puts the step-by-step flow in the `description`.

**Other checks are expected to fail.** There is no eval scenario, no `SKILL.md` entry and no `documentation.yaml` line, so the coverage and registration gates will complain. That is the point of leaving them out — none of them start an eval run, so this fixture costs nothing beyond the review itself.


omermesser:
Closing — the comment flow was verified here: the verdict posted as a new comment and the previous one collapsed as outdated. The acknowledgement deletion can only be checked once #1372 is on main, since issue_comment workflows always run from the default branch.

PR #1394, opened 2026-09-20
wix-manage(upload-static-site): add agent identity via utm_source (keep campaign=mcp)

ayal:
## Why
We can attribute drops to a referral source (`campaign=mcp`) but not to the **agent/tool** that made them. This adds a separate agent-identity signal **without changing** the existing `mcp` tag — so `source='mcp'` reports keep working.

## Change
- **Keep** `utm_campaign=mcp` (drop link) and `campaign=mcp` (API upload) unchanged.
- **Add** `utm_source=<agent-id>` on both — the agent's own short, stable slug (`claude-code`, `cursor`, `codex-cli`, `windsurf`, `github-copilot`, or a product name; `unknown-agent` fallback). Same param on both paths.

## Pairs with
**headless-bo #1252** — reads `utm_source` on the upload and stamps it on the `160:23` event as `extra_fields.utm_source` (keyed by msid → joins to release/claim/premium). Browser drops carry `utm_source` in the LP `page_url`, extractable via `regexp_extract(page_url,'utm_source=([^&]+)',1)`. Same field name across both flows.

## Evals
Require `utm_campaign=mcp` kept **and** a `utm_source` with the agent's real id (fail if omitted or left as the `<your-agent-id>` placeholder).

## Notes
Self-reported (spoofable/omittable) — `user_agent`/`ip` remain corroborating signals.


ayal:
#skipreview

